PR #521, opened 2015-10-05
Swedish Translations Complete

Abbe98:
Needs testing


dg0yt:
This is merged.

Note that "unstable" builds for testing are available from the openSUSE build service for [Linux](https://software.opensuse.org/download.html?project=home%3Adg0yt&package=openorienteering-mapper-unstable) and for [Windows](http://download.opensuse.org/repositories/home:/dg0yt/Windows/) (selfextracting ZIP). This change might appear in app. one hour if the build doesn't break.

Issue #347, opened 2014-07-27
Documentation update


dg0yt:
**dg0yt reported on Sourceforge [[tickets:#347](https://sourceforge.net/p/oorienteering/tickets/347)]:**

The manual needs updating.

Collected by @Abbe98:

 - [X] Wiki icons missing (#311) (@dg0yt)
 - [X] FAQ needs update (updated from SF wiki -- @dg0yt)
 - [X] Update shortcut keys (#458)
 - [X] Map parts has no documentation (OpenOrienteering/mapper-manual#12)
   - [X] `map_menu.md` is missing map parts items
 - [X] `view_menu.md` is missing tag editor item (@wanacode)
 - [X] Tag editor has no documentation (needs more work -- @wanacode)
 - [ ] ~~Import has no documentation~~ (moved to #583)
 - [ ] ~~Missing docs for some tools?~~ (moved to #583)
 - [X] Update `georeferencing.md` (screenshot updated -- @dg0yt)
 - [ ] ~~Drawing tools "cheat sheet" (#160)~~ (moved to #583)
 - [X] Update screenshots (@dg0yt)

Abbe98:
Trying to cover whats needs to be done:

 - [ ] Wiki icons missing #311
 - [ ] FAQ needs update
 - [ ] Update shortcut keys #458
 - [ ] `map_menu.md` is missing map parts items
 - [ ] `view_menu.md` is missing tag editor item
 - [ ] Tag editor has no documentation
 - [ ] Import has no documentation
 - [ ] Missing docs for some tools?
 - [ ] Map parts has no documentation
 - [ ] Update `georeferencing.md`

Additional:

 - Update screenshots(some differs)

This is just what I found after a quick read-through.

dg0yt:
#347 is not directly related to the source, but to the manual's copy in the user wiki on sourceforge. I have not found yet a similar solution on Github which facilitates user contributions.
- GH wikis use a slightly different syntax for linking pages, and it is non-trivial to add/update pictures. These issues are show-stoppers.
- The manual might get a separate repository, with a different set of privileged accounts. On the other hand, contributions can be handled easily with pull requests.
- It might even be possible to have that separate repository published via jekyll. On the other hand I'm thinking about publishing updated manuals to the gh-pages branch of the mapper repository. This could be a simple optional build rule. I would publish the API documentation the same way.
[...]

PR #522, opened 2015-10-05
Meta text files to markdown

Abbe98:
Improves readability(mostly on Github) and maintainability.


dg0yt:
Okay, the readability on Github is improved indeed. But these files are also meant for offline viewing. From that point of view I have got the following issues:
- The headlines shall have a space before the first word, e.g. `## General`
- Inline URLs don't really need extra markup IMHO.
- Even single command lines should be formatted as code block, not as inline code. Otherwise the backticks end up in the command line when doing copy-and-paste from a text editor.

The README(.md) in the symbol sets directory is correct in its content, but it is sort-of obsolete. `test/symbol_set_t` performs all necessary steps on symbol sets and examples. Maybe another thing to update in #347.

Abbe98:
Updated

PR #532, opened 2015-10-28
Android: Prefer external storage

Abbe98:
Currently App Install Location is set to Auto this results in that Mapper first will be installed to the system storage, then the user can move the app to external storage if they wish to do so. If Mapper instead preferred external storage the app would be installed and downloaded to the external storage if such is available. The user would still be able to move the app to the system storage after the installation.

(Short story: for most users system storage is more valued then external storage.)

Example of an user benefiting from this approach: 

Currently the user has a lot of space available on his/her SD card but the system storage is almost full, this results in that the he/she can't download the app because it would first be downloaded to the system storage. If it would first be downloaded to the SD card this issue would never have happen.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionName="0.5.99" android:versionCode="1" package="org.openorienteering.mapper" android:installLocation="auto">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionName="0.5.99" android:versionCode="1" package="org.openorienteering.mapper" android:installLocation="preferExternal">
     <application android:name="org.qtproject.qt5.android.bindings.QtApplication" android:label="@string/app_name" android:icon="@drawable/icon">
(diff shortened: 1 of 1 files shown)


dg0yt:
There is a tradeoff: Some users benefit from being able to install despite lack of space in internal storage. But any user can be affected by the appliation being killed when he connects the device as mass storage to a computer. Mapper doesn't permanently save its data, so data loss can easily happen. This disadvantage outweights the benefit IMHO. And the number of devices which are low on memory but otherwise usable for Mapper is probably low.

Abbe98:
I'm unsure when data loss could happen and how easily this happens but if it is something that's a common issue for apps the disadvantage outweights.

I could not fin any more information on the topic by simply Google for it.

I guess we could test this by simply moving the existing app to a SD-card and try to cause data loss.

dg0yt:
Asking Google for "Android preferExternal" gives http://developer.android.com/guide/topics/data/install-location.html. Killing the app means any unsaved changes are lost. The app could be in the background at that time. 

The app might get killed when it is in the background anyway. Maybe we can come back on this proposal when we are better prepared for the app being killed.

Another idea would be to display a warning when the app runs from external storage.

Abbe98:
Saw now that apps is actually killed when the external storage gets connected, and immediately means we can't deal with it.

Closing this for now.

PR #540, opened 2015-11-18
Updated shortcut keys in documentation.

wanacode:
On Windows and Linux only number keys work with Ctrl + Shift, not with the Function keys.


dg0yt:
Merged, but commit message modified. Thanks.

PR #563, opened 2015-12-23
Update README

Abbe98:
Trying to make the README more intuitive and make sure it's useful for future contributors.


dg0yt:
Note that there is more information for contributors in the [wiki](https://github.com/OpenOrienteering/mapper/wiki).

PR #568, opened 2015-12-24
Training course symbol set

Abbe98:
A set with custom course symbols useful when making complex training courses.

For review before, edits to this pull request is likely. 

screenshot

### DIFF
--- a/symbol sets/CMakeLists.txt
+++ b/symbol sets/CMakeLists.txt
@@ -22,6 +22,9 @@ set(Mapper_SYMBOL_SETS
   10000/ISOM_10000.omap
   15000/ISOM_15000.omap
   
+  10000/Training_Course_Items_10000.omap
+  15000/Training_Course_Items_15000.omap
+  
   4000/ISSOM_cs_4000.omap
   5000/ISSOM_cs_5000.omap
   10000/ISOM_cs_10000.omap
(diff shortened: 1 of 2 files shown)


dg0yt:
- The symbol numbers are above 999.9. Do you know if this is compatible with OCD 8 or 11? Otherwise I would suggest using numbers between 900.0 and 999.9.
- The symbols need to get a description.
- Can the symbols be used in 1:10000 (1:5000, 1:4000) without change in dimensions?

Abbe98:
- something I didn't think of!
- work in progress
- no, I believe such a symbol set is most relevant for 15000 and 10000 but
maybe it should be available for sprint scales too.

Abbe98:
Tested the symbols in OCAD(11), and added basic descriptions.

Abbe98:
@dg0yt on the topic of dimensions, I'm targeting only 10000 & 15000 at this time. How is the ISOM symbolset currently dealing with generating the 10000 `.omap` file?

As far as I can see this would cause no issues with dimensions between the 15000 & 10000 `.omap`.

Symbian9:
This issue in some way related
https://github.com/OpenOrienteering/mapper/issues/664

dg0yt:
Any update on this one? There are now also control descriptions (#718) although not merge yet.

Abbe98:
I have not done any work on sprint scales,  for 10000/15000 where I think it most relevant. Also as mentioned above it's compatible with OCAD. It could be merged as it's?



I would personally opposite a merge of #718.

dg0yt:
If there are substantial arguments against #718, bring it up there.

dg0yt:
Do you want to update or elaborate this PR? Otherwise I'd rather close it.

PR #570, opened 2015-12-24
Russian translate

sembruk:
Added Russian translation file. It had been tested in 0.6.0 release.


dg0yt:
Болшое спасибо...

I will add your name to the list of contributors, okay?
The translation will be available in the unstable download, and maybe as an extra download for v0.6.0, after Christmas.

sembruk:
Пожалуйста!
> I will add your name to the list of contributors, okay?
Sure!
> ... maybe as an extra download for v0.6.0, after Christmas.
It will be great!

PR #617, opened 2016-03-08
Fr translation

ogeo-project:
I submit to a new french translation file.
This new file is correction and upgrade of existing file.
Best regards
Eric


dg0yt:
I did some edits (and wrote "minor" once more than is true, sorry). If there is some real name to be added to the list of contributors, let me know.

ogeo-project:
My real name is Eric Boulet, I live to Saint-Malo in France.
I'm in charge to "Dinan Course d'Orientation"

http://www.dinan-course-orientation.fr

Sportivement

Eric

dg0yt:
You will have to more corrections. 

- The following change was probably wrong.
    
   ```
          <message>
             <location filename="../src/gui/settings_dialog.cpp" line="218"/>
             <source>Preview the width of lines with helper cross</source>
    -        <translation>Aperçu de l&apos;épaisseur des lignes avec la croix d&apos;aide</translation>
    +        <translation>Aperçu de l&apos;épaisseur des lignes de la croix d&apos;aide</translation>
         </message>
    ```

- Before punctuation ( ! : etc.), you added an extra space character in some spots. I don't think it is correct. I admit I am not an expert in French grammar, but a) it should be consistently follow the language's standard, and b) space on screens is limited.

Since I didn't exactly merge your pull request, you may need to merge my change, or re-create your fork.

ogeo-project:
Yes, first point is mistake.

For second point, is typographic rule (space before ; : ! ...), but I 
agree with you : space on screen is limited.

I have a question about menu items : when character '&' (use to 
underline particular character into word) don't match any shortcut, it 
should produce bug, if I remove character '&' ?
Example : &Save ---> Enregistrer

New version without extra spaces in two or three days, it should be not 
last version ;-)

Best regards

Eric

dg0yt:
All right, please handle space before punctuation like you find it in mainstream software (e.g. office suits, operating systems) with professional translation.

For the ampersand '&', see http://doc.qt.io/qt-5/linguist-translators.html#changing-keyboard-accelerators. The '&' **defines** an "Alt Key Accelerator". As a translator, you may move it to another character, or even remove it. Be careful not to introduce duplicates. Again, consistency with other applications should be the major guideline.

There are also "Ctrl Key Accelerators", but I'm not sure if we provide them for translation (as we should).

Question: Do you find it hard to understand where each translation is used? It is not easy to improve information in this direction (the source code can be displayed in Qt Linguist). But it might be possible to have information the other way: Display a log of recently looked-up translations while the application is running. This could help to prioritize the translation effort.

ogeo-project:
Hi,

I start to second pass on french translation this morning.

It seems that *.ts file is not synchro. For example at 
"EditorSettingPage" item QT Linguist don't display source code. In this 
case, after loading my new french *.qm file in Mapper, this page is 
display in english.

Can you help me ?

Sportivement

Eric

dg0yt:
The .ts files are updated only occasionally, so they match the sources at the time of the update. "EditorSettingPage" refers to a class which I renamed recently, and I manually applied the change to all translations. But it is easy to spot in the program, it is the editor page in the settings dialog.

I guess I have to run the update process again soon. There is a lot of translation activity recently. If I knew in advance, update and translation could be coordinated...

PR #619, opened 2016-03-10
ISMTBOM 1:15000

999999333:
MTBO symbol set is done! made from official IOF specifications document. Smaller scales (10000, 7500, 5000) will be soon... Its very easy to convert foot-O maps to MTBO maps with a few clicks (some patch ridebility adjustments are needed).

image

If there is any bad created symbol let me know 

vystrizek
vystrizek
(Map: Sletiště, Czech republic, Mapped by J. Cabrnoch, Revision P. Vinš)


dg0yt:
Moc hezky, dekuji.

If you can describe what changes need to be done for different scales, I can add an automatic conversion, like I do for ISOM (15000->10000) and ISSOM (5000->4000).

dg0yt:
Scaling is applied to all symbol now, including course setting. This is different from what we do for ISOM/ISSOM at the moment.

PR #620, opened 2016-03-11
Added ISSkiOM 15000

999999333:
Added symbol set for ski orienteering, I think it need testing from anybody who is in Ski-O, I never made Ski-O map. Let me know with any problems :)
All symbols are from IOF specification (ISOM 2000 + ISSkiOM 2014)
image

Quick transformation of foot-O map 
image

image

(Map: Sletiště, Czech republic, Mapped by J. Cabrnoch, Revision P. Vinš)
Scaling is wrote below
image


dg0yt:
Did you originally create the missing symbols, or copy from another source?

999999333:
Created them, other (recommended ISOM) I copied from base symbol set in OOM

dg0yt:
It is okay to copy from (or better start with) the symbols sets from OOM.
I'm asking because I see a different (new) in color for 812 Eqipment deposit.

In the ISMTBOM symbol set, the color table is signifcantly different from our ISOM symbol sets. Where do these colors come from?

lorrieq:
Hey, 

Looking great, nice one!

I made the symbol sets as well, nice exercise to do.

I think that the colours I used are the same as 999999333, in that they are also different from your ISOM symbol set colours.

For the Ski-O colours, it mentions in section 5 of <a href="http://orienteering.org/wp-content/uploads/2010/12/International-Specification-for-Ski-Orienteering-Maps-2014-v3.pdf">this link</a> some PMS colour numbers for spot colours which I followed (such as PMS 136 for yellow).

dg0yt:
@lorrieq I'm talking about a particular color in the ISSkiOM file, and the full set of colors in the ISMTBOM file. The question is not related to the documents.

999999333:
@lorrieq ISMTBOM colours cmyk values come from ISOM_cs_15000 and order is as specified in iof document (there is missplaced black "roads" colour, because of connecting roads together).
image
image

And I told that ISSkiOM need testing... corrected.

dg0yt:
I finally pushed files which I created from Mapper's ISOM 15000 symbol set, thus actually reusing the color definitions which are essential and (somewhat) tested for overprinting simulation and CMYK output. ISOM symbol which are not used for Ski and MTBO are marked as hidden for now and only left out when creating a new symbol set (could be useful for adding automatic conversion, e.g. for runnability/vegetation). Now it is up to everyone else to test the result.

PR #640, opened 2016-03-29
fix obvious typo

matkoniecz:


dg0yt:
I will fix in an update to INSTALL.md.

PR #658, opened 2016-04-11
French translations

amalpeyre:
Translations are OK, symbol sets are still in progress.

Thx !


dg0yt:
I merged the translation updates and the tips. The symbol set can be merged when it is ready.
Merci beaucoup !

dg0yt:
I had to do two additional fixes. I also have to point out that colons (":") are now sometimes used with space before, sometimes without.

Issue #541, opened 2015-11-19
Manual will not open in Linux


wanacode:
I cannot open the manual in the software, either from the menu or with the shortcut key F1.

For now I am using the online version http://sourceforge.net/p/oorienteering/wiki/Manual%20Contents/

This is a problem on Linux Mint 17.1 with version 0.5.96 and also on Debian Jessie Xfce version 0.5.99

dg0yt:
Is there an error message, and what does it say?

wanacode:
No, just nothing happens. No error messages.

dg0yt:
This is strange. There should be an error message if either the manual or Qt Assistant would be missing. (The manual works in Windows and Ubuntu 14.04 packages.)

dg0yt:
This behavior is caused by qtchooser. Run "assistant" on the command line to see what happens. It even exits with an error. But Mapper doesn't wait for the exit code: it only verifies that the program could be started.

Solution: Set QT_SELECT environment variable to "qt5".

References:
- http://perezmeyer.blogspot.com.ar/2013/08/qt-in-debian-using-qt4-andor-qt5-in.html

dg0yt:
The common solution for Ubuntu, Debian, LFS, Arch is to set the QT_SELECT environment variable to "5", not "qt5".

PR #684, opened 2016-06-06
Set QT_SELECT=5 instead of QT_SELECT=qt5 env variable for Assistant

sikmir:
The problem is that I couldn't open the manual (just nothing happened, no error messages).
After looking at `strace ./Mapper` output, I discovered the follow error:
`assistant: could not find a Qt installation of 'qt5'`
Ther reason is that on ArchLinux/LFS:
```
> qtchooser -list-versions
4
5
default

```
But #541 works for Debian/Ubuntu, where both qt5 and 5 exist.


dg0yt:
Okay, the current Debian and Ubuntu packages do provide .../qtchooser/5.conf indeed.

PR #734, opened 2016-07-31
Ukrainian translation review

yevhenmazur:


dg0yt:
Thanks. Some items remain to be checked and finished, cf. https://hosted.weblate.org/projects/openorienteering/mapper/uk/.

yevhenmazur:
Yes, I know, but I cant approve these items until I don't see them in the interface. If somebody can help me with reproducing of the cases when these messages appears - it is more then welcome.

dg0yt:
Add a source comment in Weblate to ask for clarification.

yevhenmazur:
Thanks, will do so.

PR #753, opened 2016-09-07
8-bit encoding from general settings for new OCD importer

sembruk:


dg0yt:
Ah, I can push to pull requests, indeed...

Thanks for spotting the ignored setting. But I want to set the correct encoding already in the constructor. Okay?

sembruk:
> But I want to set the correct encoding already in the constructor. Okay? 

Sure!

PR #755, opened 2016-09-14
OGR file import with non-latin symbols in filename

sikmir:


dg0yt:
This might be tricky, at least cross-platform: This is going to work for Latin-1 (as it did before) and for systems where GDAL/OGR happens to accept Utf-8 (probably because the system locale is Utf-8). But I really don't know if there are systems which use a different locale (Japanese Windows 10? Korean macOS 10.11?). There is also [QString::toLocal8Bit()](http://doc.qt.io/qt-5/qstring.html#toLocal8Bit), but this function may return a byte array which is undefined if the string contains any characters that cannot be encoded in the locale.

sikmir:
[GDAL RFC 30: Unicode Filenames](https://trac.osgeo.org/gdal/wiki/rfc30_utf8_filenames):
"In brief it will be assumed that filenames passed into and returned by GDAL/OGR interfaces are UTF-8."
Implemented in [GDAL/OGR 1.8.0](http://www.osgeo.org/node/1135).

dg0yt:
Thanks for looking this up!

Issue #756, opened 2016-09-15
Linux: File types (*.omap, *.xmap, *.ocd) not associated with mapper


Symbian9:
### Steps to reproduce
1. Install OOMapper on Linux
2. Try open files *.omap, *.xmap and *.ocd from file manager
3. Files not recognized as OOMapper file formats

### Actual behaviour
MIME types of *.omap, *.xmap and *.ocd not automatically associated with OOMapper

### Expected behaviour
After installing OOMapper *.omap, *.xmap and *.ocd must be automatically associated with it using MIME types settings (install it in <code>/usr/share/mime/..</code> folder) and icons of this file types (<code>/usr/share/icons/hicolor/..</code>) must be same as OOMapper icon. Also files *.omap, *.xmap and *.ocd must recognized as "orienteering map" file type.

### Configuration
Mapper Version: 0.6.5
Operating System: LinuxMint 17.3 XFCE (64bit)

### REFERENCES
* https://help.gnome.org/admin/system-admin-guide/stable/mime-types.html.en
* https://developer.gnome.org/integration-guide/stable/mime.html.en
* https://freedesktop.org/wiki/Software/shared-mime-info/

### Subtasks (by @dg0yt)

- [x] MIME types for .omap and .xmap (@sembruk, #758)
- [x] MIME type for .ocd (@dg0yt)
- [x] Packaging update for openorienteering-mapper-unstable (@dg0yt)
- [x] Packaging update for openorienteering-mapper (release) (@dg0yt)
- [x] Rename files and mime type in unstable package (@dg0yt)

Symbian9:
@dg0yt, please, add Ukrainian comment inside https://github.com/OpenOrienteering/mapper/blob/master/packaging/debian/openorienteering-mapper.xml

<code>...
<comment xml:lang="uk">Спортивна карта</comment>
...</code>

dg0yt:
The duplicate icon files caused an rpmlint warning. I added `%fdupes -s ...` to OBS RPM package specs to symlink the duplicate files during packaging. In the long run, we need dedicated file type icons.

PR #760, opened 2016-09-23
Added Ukrainian comment for map MIME type

sembruk:
Issue #756


yevhenmazur:
Yes, it is correct:
"Спортивная карта" - Russian
"Спортивна карта" - Ukrainian

PR #766, opened 2016-10-03
Fixed print resolution validator

sembruk:
If `dpi` sufix in popup list translated impossible to change the print resolution.


dg0yt:
Thanks.

PR #780, opened 2016-10-19
Update translations in the Linux desktop files (fi)

sikmir:


dg0yt:
Even these translations are taken from the .ts files, i.e. Weblate (Cf.https://github.com/OpenOrienteering/mapper/commit/d3f23a87c5c879d4d7045ecf609a0c937b690269). Direct changes will be lost.

sikmir:
Thanks, I just didn't know translation of this file was integrated with Weblate.

PR #785, opened 2016-10-30
Fix inconsistent caps usage in Indonesian translation

adhikasp:
This is small fixes that correcting various inconsistent capitalization usage in Indonesian translation. For rule of thumbs, capitalization usage in Bahasa Indonesia is similiar with English.

Example:
```xml
<source>Show tip of the day</source>
<translation>Tampilkan Tips</translation>
```

became
```xml
<source>Show tip of the day</source>
<translation>Tampilkan tips</translation>
```

Mention to @arifbobidinamik, the original Bahasa Indonesia translator.

### DIFF
--- a/translations/OpenOrienteering_id.ts
+++ b/translations/OpenOrienteering_id.ts
@@ -57,7 +57,7 @@
     <message>
         <location filename="../src/gui/widgets/general_settings_page.cpp" line="92"></location>
         <source>Show tip of the day</source>
-        <translation>Tampilkan Tips</translation>
+        <translation>Tampilkan tips</translation>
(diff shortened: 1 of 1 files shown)


dg0yt:
Thanks for the pull request. I will try to merge it later today. I would suggest to make changes directly in Weblate next time. Otherwise I could get conflicts when merging the translation from Weblate into Gituhb.

@arifbobidinamik  There also are some issues with special markers. Some days ago I extended the [information for translators](https://github.com/OpenOrienteering/mapper/wiki/Translation#information-for-translators) in the wiki with more details on the subject.

Issue #358, opened 2014-07-29
Select objects using tags


dg0yt:
**gandym01 reported on Sourceforge [[tickets:#358](https://sourceforge.net/p/oorienteering/tickets/358)]:**

In future, can a tool be provided to select objects on the basis of an imported tag(s)? For example, I imported a DXF file and found, via the tag editor, that the layer number was provided as a key-value pair (key:name, value:30). The ability to select all objects with a certain key-value pair(s) would make applying symbols to newly imported DXF data (that contains many different layers) much easier and give an effective workaround to [Ticket 288](https://sourceforge.net/p/oorienteering/tickets/288/).

Also, does this mean [Ticket 337](https://sourceforge.net/p/oorienteering/tickets/337/) might already be partially achieved?

gandym

mitchellkrome:
I have made a small widget that allows for selecting objects based on the tags. Open to any comments on the layout, otherwise can make a pull request with the code.

![tag_selector](https://cloud.githubusercontent.com/assets/8371236/20820320/32ca76ba-b886-11e6-944d-d7d41c5942c9.PNG)

dg0yt:
Basically looks good to me.
 - Instead of comparison "not", I would prefer "is not".
 - Is there any precedence of relation "or"/"and"?
 - Would it make sense for relation be the first column (empty/disabled for the first row)?
 - Is this the right thing for a dock widget, or should it be a dialog instead?

(I would have implemented a pure text search with keywords, but your approach is fine as well.)

mitchellkrome:
- Is there any precedence of relation "or"/"and"?

Precedence was something I had a bit of trouble working out how to do well with this sort of layout- the way I've written it now it evaluates top down with "and" and "or" having the same precedence.

- Would it make sense for relation be the first column (empty/disabled for the first row)?

First column makes more sense - I'll give it a go.

- Is this the right thing for a dock widget, or should it be a dialog instead?

Almost everything I've seen in mapper so far used these dockable type things so I just went with that because it gave me something to base the code off which already used a table, etc (This is the first time I've messed around with Qt). QGIS has a similar "select by value" feature which is a dialog and I find that it always gets in my way when I try to use it. With a dialog do you imagine some sort of one use popup that goes away when you make the selection, or just the same thing but it can't be docked into the interface?

[...]

PR #815, opened 2016-12-13
Add ability to select objects based on tags

mitchellkrome:
Provides a solution to #358.


dg0yt:
I made a number of changes, and I will give you a short time for review before merging this to master.

I thought again about operator precedence. Maybe the best would be to handle AND and OR differently, so that you can make expression of the effective form (a&&b) || (c&&d&&e) || (f). In the dialog, we could pad the AND with space in order to visualize the precedence. 
While writing these lines I realize that we could even short-circuit the evaluation then.

mitchellkrome:
All looks good to me! Few things I didn't think/know about, will keep them in mind. 

That style of precedence is probably more useful than how I have it set up now. It would probably be best to implement this by having makeQuery return a QString, and then have a small parser in ObjectQuery which generated the query from a string representation. I can give that a go over the coming days.

PR #821, opened 2016-12-27
Add missing qtbase5-private-dev package in INSTALL.md for Linux build

bagage:
I also removed commas so that one can copy the whole line directly to the terminal.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -74,7 +74,7 @@ PROJ.4 library and of CUPS are installed. For a Ubuntu or Debian system, install
 
 For using the Qt libraries provided by the repositories, the following packages
 need to be installed:
-`qt5-default` (>= 5.2), `qttools5-dev`, `qttools5-dev-tools`, `libqt5sql5-sqlite`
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -73,8 +73,9 @@ PROJ.4 library and of CUPS are installed. For a Ubuntu or Debian system, install
 `libproj-dev` and `libcups2-dev`.
 
 For using the Qt libraries provided by the repositories, the following packages
-need to be installed:
-`qt5-default` (>= 5.2), `qttools5-dev`, `qttools5-dev-tools`, `libqt5sql5-sqlite`
+need to be installed. `qt5-default` needs at least version 5.2:
+
+`qt5-default` `qtbase5-private-dev` `qttools5-dev` `qttools5-dev-tools` `libqt5sql5-sqlite`
```
dg0yt: It is correct to add qtbase5-private-dev here, but the change of the wording about the version does not look like an improvement IMO.
(BTW the best documentation of build requirements are the control files on build.opensuse.org.)

bagage:
Updated accordingly.

bagage:
Up?

dg0yt:
Well, focus was on other things recently, but I will pick it up now.

dg0yt:
Merged. The reference got lost in another edit.

Issue #740, opened 2016-08-08
Refactoring


02JanDal:
Quite some time since I last touched the source of this project...

Since these would be quite big changes I want to discuss them a bit before I even start to consider doing any of them:

I believe that Mapper is in dire need of some refactoring. All the below is done on latest master at the time of writing (c0713b5d343bfaf9ed2ab1c89dcf7ccff7a03fdf).

### Files

Start by looking at the file level; the `src` folder contains 181 files... There are a lot of obvious possibilities for subfolders, just by looking at file names: `tools`, `gps`, `symbol`, `object`. There are also several dialogs and widgets that should probably move under `gui/dialogs` and `gui/widgets` here. Also several `fileformat_*` files that should move into the existing `fileformats`, similar for util.

Additionally there are many files that currently contain way more than a single class, `object.h` for example contains 2 large, 2 medium and a few smaller classes. These should probably also be split into separate files, which might change how the directory layout should be made.

### Classes

Next, there are many, many huge classes. You can argue about how big is to big for weeks, and everyone will have their own opinion. [This](https://swreflections.blogspot.se/2012/12/rule-of-30-when-is-method-class-or.html) blog post talks about the number 30 (30 lines per method, 30 methods per class, 30 classes per package/namespace etc.) and I believe that sounds quite sane, even though I think for a class that sounds quite high already, unless there is a lot of duplication through overloads, or implementations of well known abstract classes.

Lets look at some of the bigger classes in Mapper:

* `MapEditorController`: about a 170 methods. Yikes. While GUI classes can often become quite large with all the slots for different actions, this feels a bit extreme. 4 cores, 16 GB of RAM and still my IDE is just about usable when viewing this file.
* `Object`: about 40-50 methods. Better, but a lot of potential for improvement.
* `PathObject`: about 60 methods
* `Map`: I won't even count, above a 100 at least.

The list goes on and on.

Something many of these classes have in common is that they have groups of methods that belong together very tightly. `Object` has `tags`, `setTags`, `getTag`, `setTag` and `removeTag`. These could be moved into a separate `TagContainer` class, leaving a single `TagContainer *tags() const` in `Object`. Similar in `Map` for templates, map parts, symbols and colours.

There are also many classes that have their own `save`/`load` functions, it might make sense to extract this functionality too.

[...]

PR #822, opened 2017-01-01
Refactoring: Move files

02JanDal:
This PR moves a lot of the files previously directly under src/ into new and existing subfolders. I see this as a first step in refactoring (see also #740), as having stuff split up into smaller units makes it easier to get an overview of further work (one could now start writing unit tests focused on classes in a single directory etc., as well as attempt to untangle dependencies between directories).

There has not been a huge amount of thought put into which files go where, so suggestions for changes are welcome. Other than that this PR should be pretty straight forward to review, code changes have strictly been kept to adapt to the new locations of the files; there should therefore be _no_ changes in runtime behaviour.


.gitignore
```diff
@@ -8,3 +8,5 @@ build*/
 help/*.qhc
 3rd-party/*/download
 !3rd-party/*/download/README.txt
+.idea/
+cmake-build-debug/
```
dg0yt: These lines cannot be merged. Note that directories named build*/ are already ignored.

src/CMakeLists.txt
```diff
@@ -135,17 +135,17 @@ set(Mapper_Common_SRCS
...
+ file_formats/file_format_registry.cpp
+ file_formats/file_import_export.cpp
+ file_formats/file_format_native.cpp
+ file_formats/file_format_ocad8.cpp
+ file_formats/file_format_xml.cpp
+
+ file_formats/ocad/ocd_file_export.cpp
+ file_formats/ocad/ocd_file_format.cpp
+ file_formats/ocad/ocd_file_import.cpp
+ file_formats/ocad/ocd_types.cpp
```
dg0yt: Moving out the ocd_types* is okay. I would prefer the ocd_file_* files not to be moved. 

Apart from this, spelling for new files, directories and classes must be "ocd".

src/CMakeLists.txt
```diff
@@ -62,17 +62,11 @@ set(Mapper_Common_SRCS
 
  settings.cpp
 
- map.cpp
- map_part.cpp
+ map/map.cpp
+ map/map_part.cpp
```
dg0yt: Isn't this core? (At least once the remaining Qt Widgets dependencies got removed.)

src/CMakeLists.txt
```diff
@@ -187,6 +181,13 @@ set(Mapper_Common_SRCS
  gui/symbol/symbol_setting_dialog.cpp
  gui/symbol/symbol_properties_widget.cpp
  gui/symbol/symbol_point_editor.cpp
+
+ gui/map/map_widget.cpp
+ gui/map/map_editor.cpp
+ gui/map/map_editor_activity.cpp
+ gui/map/map_dialog_new.cpp
+ gui/map/map_dialog_scale.cpp
+ gui/map/map_dialog_rotate.cpp
```
dg0yt: I would rather not introduce a "map" subdirectory in gui at the moment.

dg0yt:
Some general thoughts:

1. Plural vs singular directory names: Either file_formats, templates, objects, symbols, tools, or file_format, template, object, symbol, tool. (But please don't "maps", "guis").

2. Object, Symbol, Map, Renderable are either "core" (MapColor already is!), or may together form "map". Both should be usable (testable) without Qt Widgets.

.gitignore
02JanDal: That wasn't intended to be committed, will be removed.

src/CMakeLists.txt
02JanDal: Given that the import/export of newer OCAD files consists of 6 files (excluding the ocad_types_* files) and they don't follow the file_format_* naming scheme of other formats I think keeping them in a separate directory makes sense.

Will change ocad->ocd, would be interested in knowing if there is there some specific reason for that spelling though?

src/CMakeLists.txt
02JanDal: Why not? They are GUI classes directly related to the map.

src/CMakeLists.txt
dg0yt: The file_format_* naming scheme of the other files is wrong. It is a poor substitute for using folders (which we do add now). In contrast, ocd_file_* were created by me with file names which match the class names, in a folder which is meant to take up all file formats, on the occassion of refactoring.

So now you know that you have to rename most files which you move to a folder.

src/CMakeLists.txt
dg0yt: While we had to few folders in the past, we shouldn't fall into the other extreme now. There is not much to gain, but a lot more "../" to write (`#include "../../core/map_coord.h"`)

02JanDal:
I have fixed some of the things pointed out.

* Provided this gets merged in the not so distant future I will continually rebase it once there are new changes, additionally this should likely get squashed before merge.

* Plurality: The general idea was plural for what there's multiple of (file formats, symbols, tools, etc.) but singular for others (object, map, etc.), but I see I didn't really see that through. Has been fixed.

* Object, Symbol, Map, Renderable, MapColor, etc. all moved to `core/map`

src/CMakeLists.txt
02JanDal: I don't believe this is to many. To solve the `../` issue I think we should start using absolute paths (from `src`, so `"../../core/map_coord.h"` becomes `"core/map_coord.h"`, regardless of which file it is included from). If you are ok with that solution I can take on converting paths as part of this or a separate PR.

dg0yt:
> There has not been a huge amount of thought put into which files go where ...

> Provided this gets merged in the not so distant future I will continually rebase it once there are new changes, additionally this should likely get squashed before merge.

Although the first statement wasn't encouraging, I wouldn't take the time to review if I wouldn't be willing to merge. The next release should be 0.7.0, so there is no better point in time soon.

>  Object, Symbol, Map, Renderable, MapColor, etc. all moved to `core/map`

`core` IS `core/map` IMO. Maybe some of the current files don't belong there.

~~~
src/
src/core
src/core/map               - files belong into src/core
src/core/map/symbols       - src/core/symbols at best
                           - for convenience, src/core/symbol.h could
                             #include "core/symbols/symbol.h"
src/file_formats
src/file_formats/ocd
src/gdal
src/gui
src/gui/map
src/gui/symbol
src/gui/widgets
src/libocad
src/mingw
src/printsupport
src/printsupport/patches
src/printsupport/qt-5.2.1
src/printsupport/qt-5.5.1
src/qmake
src/sensors
src/templates
src/tools
src/undo
src/util
~~~

dg0yt:
> There has not been a huge amount of thought put into which files go where ...

> Provided this gets merged in the not so distant future I will continually rebase it once there are new changes, additionally this should likely get squashed before merge.

Although the first statement wasn't encouraging, I wouldn't take the time to review if I wouldn't be willing to merge. The next release should be 0.7.0, so there is no better point in time soon. However, I can't merge the tools part at the moment, so I might start with cherry-picking some commits, or merging an intermediate state. That's why you should edit your previous commits, rather than piling new ones on top. Squashing and merging as a whole is unlikely to happen from the current state.

>  Object, Symbol, Map, Renderable, MapColor, etc. all moved to `core/map`

Maybe it is better to discuss the whole picture:

~~~
src                        - The root, which could be added to the include path
                             in order to allow #include "core/map_coord.h"
src/core                   - Core map stuff which doesn't belong anywhere else
                             (Some existing files might be moved out.)
src/core/map               - These files belong into src/core, src/core/objects,
                             src/core/renderables
src/core/map/symbols       - src/core/symbols at best
                           - For convenience, src/core/symbol.h could
                             #include "symbols/symbol.h"
                             Similar for object.h, renderable.h.
src/file_formats           - Just use the existing src/fileformats.
src/file_formats/ocd       - Not needed, but maybe there could be
                             src/fileformats/ocd for ocd_types*
src/gdal                   - All GDAL stuff (maybe moves to src/templates/gdal)
src/gui                    - Complex GUI elements
src/gui/map                - Complex GUI elements related to map
src/gui/symbol             - Complex GUI elements related to symbols
src/gui/widgets            - GUI widgets, limited complexity
src/libocad                - libocad (OCD (legacy) export and legacy import)
src/printsupport           - Printing, esp. modified Qt PrintSupport
src/sensors                - All sensors stuff* (GPS, compass, etc.)
src/templates              - All templates stuff*
src/tools                  - Most tools*
src/undo                   - Undo support
src/util                   - Various utilities.

*) Mixes data manipulation and GUI.
~~~

dg0yt:
I pushed my pending changes to the tools. You will probably have to redo the tools part now. If you don't plan to do it soon, please drop a line here.

02JanDal:
Decided to redo the commits from scratch (really happy for refactoring tools in modern IDEs right now).

While initially slightly straying away from your list (which is similar to what I hoped would be the outcome of #740) I ended up following it very closely.

Slightly of topic: An IRC channel or similar would be very nice to have to discuss bigger changes like this, Github/mail/etc. fell to bulky. Has that ever been considered? Could also be used for support etc.

test/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ endfunction()
 # Include generated files (moc ouput, build configuration)
 include_directories("${CMAKE_CURRENT_BINARY_DIR}/../src")
 include_directories("${CMAKE_CURRENT_BINARY_DIR}")
+include_directories("${CMAKE_SOURCE_DIR}/src")
```
dg0yt: Use `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR`

src/CMakeLists.txt
```diff
@@ -71,16 +71,16 @@ set(Mapper_Common_SRCS
...
- symbol_point.cpp
- symbol_line.cpp
- symbol_area.cpp
- symbol_text.cpp
- symbol_combined.cpp
+ core/symbols/symbol.cpp
+ core/symbols/symbol_dialog_replace.cpp
+ core/symbols/symbol_setting_dialog.cpp
+ core/symbols/symbol_properties_widget.cpp
+ core/symbols/symbol_point_editor.cpp
```
dg0yt: Widgets and dialogs should go to `gui/...` IMO

test/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ endfunction()
 # Include generated files (moc ouput, build configuration)
 include_directories("${CMAKE_CURRENT_BINARY_DIR}/../src")
```
dg0yt: For consistency with the line inserted below, `include_directories("${PROJECT_BINARY_DIR}/src")` ?

src/CMakeLists.txt
02JanDal: Ups, forgot about those.

02JanDal:
I guess I should have come with some more concrete examples in the issue, will do that in the future. Do you prefer slightly bigger PRs with individual commits like this one, or should I split up stuff into further PRs (example: splitting files (like the symbol dialogs) with one PR or one commit per directory?).

dg0yt:
Concretes examples always help :-)

A single PR with individual commits is fine. However, the project should compile after each single commit (for the case of bisecting). "Add src/ to the include path to get cleaner #includes" probably needs to go first.

02JanDal:
WTF git, why on earth did it reorder that?

dg0yt:
Maybe it is just the Github display. I didn't check locally.

02JanDal:
Seems to be Github, both local and [here](https://github.com/02JanDal/mapper/commits/refactor_1) is correct.

test/CMakeLists.txt
```diff
@@ -120,8 +120,9 @@ endfunction()
 
 
 # Include generated files (moc ouput, build configuration)
-include_directories("${CMAKE_CURRENT_BINARY_DIR}/../src")
-include_directories("${CMAKE_CURRENT_BINARY_DIR}")
+include_directories("${PROJECT_BINARY_DIR}/src")
+include_directories("${PROJECT_BINARY_DIR}")
```
dg0yt: No, this remains CMAKE_CURRENT_BINARY_DIR (for files generate in test subdir of binary tree).

02JanDal: Now you confuse me... https://github.com/OpenOrienteering/mapper/pull/822#pullrequestreview-15640313

dg0yt: I noticed ;-) Take a step back. We need to include:
- 123: files generated in build tree /src
- 124: files generated in build tree /test
- 125 (new): files in source tree /src

So when using PROJECT_XXX_DIR for /src in line 125 (src in source tree, "line inserted below"), I suggested to also use PROJECT_XXX_DIR for /src in line 123 (src in build tree).
For the current binary directory (line 124), there is no need to take the way over PROJECT_XXX_DIR.

02JanDal:
I hope the qmake change is enough, I don't have an Android toolchain set up right now. In a later part of the refactor I'd probably see about reducing the android part to just a single .cpp file, and the rest a library/libraries. Of course being able to skip qmake completely would be best, but I assume deployment with just CMake is non-trivial compared to using qmake.

02JanDal:
I'll see if I can get it to build later today/tomorrow, otherwise I'll leave it to you.

dg0yt:
API-docs updated. Still a wild network of dependencies: http://www.openorienteering.org/api-docs/mapper/dir_68267d1309a1af8e8297ef4c3efbcdba.html

02JanDal:
Well, I'd assume it'd to look worse than before since now it's actually visible in that graph what parts belong together.

dg0yt:
You are right, a single dir would look much cleaner ;-)

I'm now fixing the gdal folder. It is an optional part of the build, so we missed it.

Issue #517, opened 2015-09-17
Submit to F-Droid


simison:
Please. :-)

https://f-droid.org/

dg0yt:
I would be glad to see Mapper on F-droid. But Mapper isn't a simple Android app. As a native app, it has a quite heavy dependency on the Qt framework. So far I haven't found other apps which could serve as a guide, nor sufficient documentation. This would take away time from other issues.

But it is a nice isolated task for volunteers...

rugk:
:+1: for this. And I think the Qt dependency should not be a problem as long as the apk is buildable from source. :smiley: 

BTW here is a HowTo: https://f-droid.org/wiki/page/Inclusion_How-To

dg0yt:
:+1 from me, too. 

There is no lack of How-To. There is lack of people contributing time and results. Building for F-Droid is perfectly separate task.

rugk:
What about a "good volunteer task" label or something similar? :smiley:

dg0yt:
We had such a label... But I didn't find it 5 minutes ago...

rugk:
Probably just the default ("help wanted"](https://github.com/OpenOrienteering/mapper/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)?

bagage:
I am almost done on this, but I'd like one bug to be fixed before submitting it to F-droid. Currently if you build mapper using the following (which is basically what Fdroid robot will do):

```
export Qt5_android=$QT_PATH/android_armv7
$Qt5_android/bin/qmake
make
```

You end up with the following `doxygen` error:

```
oolean-merge-holes.png
CMake Error at cmake_install.cmake:42 (file):
  file INSTALL cannot find
  "/tmp/mapper/doc/manual/manual/html/graph_legend.png".
```

Currently I'm "fixing" it by running `make || cp doc/manual/manual/html/tab_a.png doc/manual/manual/html/graph_legend.png && make`, which is not very proper. But I don't get who is generating this manual and from where this `graph_legend.png` should come from… Any help would be welcome :).

dg0yt:
`graph_legend.png` sounds like a file coming with doxygen. Check the doxygen version, check the scripts in `doc/manual/`

bagage:
OK, this is a problem with doxygen packaging on Debian then. I am forcing the build of doxygen for fdroid. PR in progress [here](https://gitlab.com/fdroid/fdroiddata/merge_requests/1922).

[...]

PR #823, opened 2017-01-03
Update doxygen version to 1.8.13 to fix issue with flex 1.6.X

bagage:
On debian, doxygen is throwing an error, see [upstream bug](https://trac.macports.org/ticket/49881). Bumping doxygen to 1.8.13 fixes the problem.

### DIFF
--- a/3rd-party/doxygen/CMakeLists.txt
+++ b/3rd-party/doxygen/CMakeLists.txt
@@ -23,7 +23,7 @@ cmake_minimum_required(VERSION 2.8.3)
 
 # Configuration options
 
-set(DOXYGEN_VERSION_DEFAULT 1.8.8)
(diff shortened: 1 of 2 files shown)


bagage:
Bumping this. Required for #517.

dg0yt:
You already have got a patch in F-Droid, and it will be dropped from the master branch here anyway... So why should I merge now? The 0.6.8 PR is meant to give you the chance to have an update without changing things which do work.

bagage:
Yeah, I was doing some cleanup of my existing commits and saw it, but then I cloned superbuild and saw that you already bumped it… I'll try to figure how this work.

dg0yt:
There will be documentation how to use the superbuild but I'm still ironing out some issues.

PR #829, opened 2017-01-26
Minor fixups

sikmir:


src/fileformats/ocd_file_import.cpp
```diff
@@ -1046,7 +1046,7 @@ OcdFileImport::OcdImportedLineSymbol* OcdFileImport::importLineSymbolBase(const
 			
 			if (attributes.end_length && attributes.end_length != attributes.main_length)
 			{
-				if (attributes.main_length && 0.75 >= attributes.end_length / attributes.main_length)
+				if (attributes.main_length && 0.75 >= (double)attributes.end_length / attributes.main_length)
```
dg0yt: This adds an old-style cast (-wold-style-cast). I will amend this in the merge.

PR #835, opened 2017-02-04
Code coverage generation and upload to codecov.io

02JanDal:
This commit adds code coverage generation when building with GCC or Clang
on Unix-based systems.

Enable it by setting Mapper_COVERAGE_BUILD to ON (the default when
Mapper_DEVELOPMENT_BUILD is set).

The main interesting target is coverage_html, which will run tests and
generate a HTML coverage report in <builddir>/coverage/html. Additionally
there is a coverage_codecov target that will upload coverage data to
codecov.io (this needs to either be run on a CI or with an explicit repo
token to work).

In case you are curious, here is the coverage report for 553223f671c765a5c75c894d299d43ea19e42e59: https://static.02jandal.xyz/mapper-coverage-20170204/
And here is the same at codecov.io: https://codecov.io/gh/02JanDal/mapper/branch/coverage

(for the codecov part to be useful it would need to be setup on some CI)


CMakeLists.txt
```diff
@@ -125,6 +125,8 @@ option(Mapper_AUTORUN_SYSTEM_TESTS "Run the system tests as part of the Mapper_T
 option(Mapper_AUTORUN_MANUAL_TESTS "Run the system tests as part of the Mapper_Test target" OFF)
 mark_as_advanced(Mapper_AUTORUN_SYSTEM_TESTS Mapper_AUTORUN_MANUAL_TESTS)
 
+option(Mapper_COVERAGE_BUILD "Builds with coverage support" ${Mapper_DEVELOPMENT_BUILD})
```
dg0yt: Does it come without negative side effects? I wonder how the coverage instrumentation affects the debugging experience.

cmake/EnableCoverage.cmake
```diff
@@ -0,0 +1,118 @@
...
+			)
+
+			set(codecov ${coverage_dir}/codecov.sh)
+			file(DOWNLOAD https://codecov.io/bash ${codecov})
+			add_custom_target(coverage_codecov
+				DEPENDS coverage_capture
+				COMMAND ${codecov} -f ${info_file} -X search -X gcov
+				VERBATIM
+				COMMENT "Uploading coverage data to codecov.io"
+			)
```
dg0yt: I dislike the idea of running a unverified executable which is downloaded everytime CMake is run. I would suggest to delegat this download to ExternalProject, with an SHA256 sum for verification.

CMakeLists.txt
02JanDal: During debugging you can occasionally step into one of the internal functions, this is extremely rare (maybe happened once or twice to me) and simply stepping-in again will put you where you want. There are also some cases where the .gcda files (the ones containing line and branch counts) get corrupted (can happen at crashes, but not always) which simply means that you get a couple hundred lines of warnings printed to stderr, quite annoying but not a huge issue.

In general I would only turn this on when actually needing access to coverage data (while trying to improve it, or after some bigger changes to make sure there is no regression in percentage). So this should probably be default off.

cmake/EnableCoverage.cmake
02JanDal: That's true. It's available [here](https://github.com/codecov/codecov-bash/blob/master/codecov), so shouldn't be hard to do.

02JanDal:
> Did you consider using one of the existing 3rd-party coverage modules for CMake? I found
> https://github.com/RWTH-ELP/CMake-codecov
> https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake
This is the 3rd or 4th project I'm adding this kind of stuff to, and the times I've tried using existing modules have usually resulted in me basically redoing the entire thing anyway. I can check though, maybe there've improved or are new so that I didn't consider them last time. On a quick check of those two the first one looks kinda promising (supports clang, includes ways to generate for multiple targets), but does way more than it should have to (and as a result is significantly more complicated), the second one doesn't support clang and only provides half of what EnableCoverage.cmake does.

I might be a bit to set in what I want (one thing providing to much and the other to little...), but there really is no need for manually going through source files and the most useful report is the one for all test suites.

> Did you consider using a distinct CMAKE_BUILD_TYPE for coverage?
Kinda, partly I personally prefer the current approach and then there are some issues with adding custom build types when working with some generators (xcode and VS IIRC, boils down to having to use CMAKE_CONFIGURATION_TYPES which has been quite volatile in it's behavior between versions. I think that would be fixed by setting the required version to some 3.x version (which might make sense anyway, 2.8 is ancient) though).

dg0yt:
I suggest you simply remove the upload part from this commit, and re-add it when adding Travis.

We may already drop CMake 2.8. It is a small set of older Linux distributions where CMake 3.0 is not readily available: Ubuntu 14.04, Mint 17, openSUSE 13.2, cf. https://github.com/OpenOrienteering/mapper/wiki/Target-systems. (And I already build CMake 3.x for openSUSE 13.2 for the Windows superbuild.)

dg0yt:
We do now require CMake 3.0. (In fact, only Debian 8.0 is still on 3.0.)

PR #836, opened 2017-02-04
Make the app restartable on language change

sikmir:
Make it possible to ask "Restart Now"/"Restart Later" on language change.


src/gui/main_window.cpp
```diff
@@ -67,6 +67,8 @@ constexpr int MainWindow::max_recent_files;
 
 int MainWindow::num_open_files = 0;
 
+int const MainWindow::EXIT_CODE_REBOOT = -123456789;
```
dg0yt: I would prefer a value more close to 0. And this CAPITAL_STYLE looks very much like an preprocessor macro.

src/gui/main_window.h
```diff
@@ -72,6 +72,7 @@ Q_OBJECT
 	/** Returns the application's localized name. */
 	QString appName() const;
 	
+	static int const EXIT_CODE_REBOOT;
```
dg0yt: This must become constexpr. Even an inline constexpr function could do this perfectly.

src/gui/widgets/general_settings_page.cpp
```diff
@@ -30,6 +30,7 @@
 #include <QFormLayout>
 #include <QLineEdit>
 #include <QMessageBox>
+#include <QPushButton>
```
dg0yt: Is this used?

src/gui/widgets/general_settings_page.cpp
```diff
@@ -173,17 +174,43 @@ void GeneralSettingsPage::apply()
...
+				messageBox.setText(tr("The program must be restarted for the language change to take effect!"));
+				messageBox.setInformativeText(tr("Do you want to restart now?"));
+				messageBox.setIcon(QMessageBox::Question);
+				auto yesButton = messageBox.addButton(tr("Restart Now"), QMessageBox::YesRole);
+				messageBox.addButton(tr("Restart Later"), QMessageBox::NoRole);
+				messageBox.exec();
+				if (messageBox.clickedButton() == yesButton)
+				{
+					qApp->exit(MainWindow::EXIT_CODE_REBOOT);
+				}
```
dg0yt: I would suggest to deduplicate the message box part. This requires to solve the different need for tr(). Maybe the easiest way is to put this in a member function which takes a function pointer named tr, receiving either a pointer to tr(), or a no-op lambda.

src/main.cpp
```diff
@@ -47,7 +47,7 @@
 #include "util/backports.h"
 #include "util/recording_translator.h"
 
-int main(int argc, char** argv)
+int AppMain(int argc, char** argv)
 {
```
dg0yt: "AppMain" is not an acceptable name for a free function. It must start with a lower case letter.

src/main.cpp
```diff
@@ -156,6 +156,17 @@ int main(int argc, char** argv)
 	return qapp.exec();
 }
 
+int main(int argc, char** argv)
+{
+	int currentExitCode = 0;
```
dg0yt: This is not the code style use for variables. Use underscores, not  camel case.

src/gui/widgets/general_settings_page.cpp
sikmir: Yes, it's used implicitly, MessageBox::addButton(..) returns QPushButton*. Without this include, we get compile error: comparison between distinct pointer types ‘QAbstractButton*’ and ‘QPushButton*’ lacks a cast.

dg0yt:
Shall I try to implement the deduplication part first?

sikmir:
@dg0yt Actually, I can do it myself just a bit later (since I don't have free time right now).

sikmir:
I'll go back to this PR on weekends, hope to catch the 0.7.0 release.

sikmir:
@dg0yt I can't find the proper way to close all windows with unsaved changes. Any ideas?

dg0yt:
IIRC this is where I stopped any past attempts because it gets tricky...

[QApplication::closeAllWindows](http://doc.qt.io/qt-5/qapplication.html#closeAllWindows) will close all windows, but in random order and only until one window does not accept the close event (i.e. user does not allow to close the window). 
- For simplicity, I wouldn't mind if we don't switch the language if one window does not accept the close event.
- To find out if there are open windows left, you can probably call QApplication::topLevelWidgets().
- In any case, you will have to get the paths of all open documents first. Cf. MainWindow::findMainWindow. 
- After the switch, you will have to open exactly these documents, not the documents passed in the command line arguments.

I will probably move the language switch dialog change to a separate commit, with modifications. As stated, I would prefer a lambda to a macro.

dg0yt:
The best way for restart would be to reopen all main windows without going through saving / loading the map files.

This would roughly mean, for each main window, to save the state (including dock widgets etc.), detach the "controller" from the window, close the old window, open a new window, attach the controller, and restore the window state. 

There might be quite a few missing pieces on this road. (That's why we don't have this feature yet.)

Issue #831, opened 2017-01-29
Ctrl+C, Ctrl+V


999999333:
Version 0.6.7., windows,  have broken copy/paste within one map, crs UTM.
It will paste the item somewhere far away out of map area. try yourself

999999333:
![before](https://cloud.githubusercontent.com/assets/4051127/22407525/5e45b312-e668-11e6-8c4c-213d463b0113.PNG)
![paste](https://cloud.githubusercontent.com/assets/4051127/22407526/5e496ed0-e668-11e6-8c4e-cbf5351eedb6.PNG)

Pictures

dg0yt:
Confirmed.

PR #838, opened 2017-02-15
Tentative Release 0.6.8

dg0yt:
Release 0.7.0 will have significant changes in source code structure and build system, requiring more effort than usual from distribution package maintainers (@sikmir, @gaudenz, @bagage). That's why I'm proposing to offer another 0.6.x release.

While I was planning to only provide a simplified fix for #831, I found a lot more changes which either fix serious bugs, or add value without risk (user manual).

To be clear: 0.7.0 is to be released soon, too. But at least the F-Droid package is likely to need significant changes.


bagage:
Can you explain a bit more why/what gonna change exactly? I'm not against updating f-droid version but it would save a bit of time to know what I should look for. Is there some 0.7.0 branch somewhere I can test building?

dg0yt:
@bagage: As indicated in #517, there is now the separate superbuild repository, so the 3rd-party build rules will be removed from the Mapper repository. (Adding GDAL with its multiple dependencies would add too much complexity.) There is no 0.7.0 branch: the master branch is what received latest development, and that's why you see cherry-picks here. From my point of view, the APK from F-Droid is incomplete anyway, see #517.

bagage:
OK, I'll try to make master build again then.

PR #840, opened 2017-02-27
Instruct QT to apply C++14 in src/

lpechacek:
Hit a problem building Mapper in Qt Creator after ff7d5669fb3b (Require C++14).


src/core/crs_template_implementation.cpp
```diff
@@ -21,6 +21,8 @@
 
 #include "crs_template_implementation.h"
 
+#include <memory>
```
dg0yt: This is already in crs_template_implementation.h.

src/src.pro
```diff
@@ -4,7 +4,7 @@
 
 TEMPLATE = app
 TARGET   = Mapper
-CONFIG  += c++11
+CONFIG  += c++14
```
dg0yt: True, but we must fix src/src.pro.in first, src.pro is generated from it when CMake runs.

lpechacek: Thanks for reminder.  I keep forgetting about it. :)

PR #854, opened 2017-04-11
Fix CRT symbol mapping when symbol has a minor number

xc-racer99:
Currently the code won't match up symbols correctly if the target symbol has a minor number due to a typo.


dg0yt:
Thanks.

PR #860, opened 2017-04-21
Add ISOM 2017 symbol set

yevhenmazur:
Add ISOM 2017 symbol set (according to ISOM 2017 v1 document from 04/04/2017)

### DIFF
--- a/symbol sets/src/ISOM2017_15000.xmap
+++ b/symbol sets/src/ISOM2017_15000.xmap
@@ -1,0 +1,10029 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="6">
+    <notes>ISOM 2017
+Symbols for OpenOrienteering Mapper
+Created by Jukka Liikari and Håvard Tveite, IOF Map Commission
+</notes>
+    <georeferencing scale="15000">
+        <projected_crs id="Local"/>
+    </georeferencing>
+    <colors count="27">
+        <color priority="0" name="Purple" c="0.150" m="1.000" y="0.000" k="0.000" opacity="1.000">
+            <spotcolors>
+                <namedcolor>Purple</namedcolor>
+            </spotcolors>
+            <cmyk method="custom"/>
+            <rgb method="custom" r="0.850" g="0.000" b="1.000"/>
+        </color>
+        <color priority="1" name="White background" c="0.000" m="0.000" y="0.000" k="0.000" opacity="1.000">
+            <cmyk method="custom"/>
+            <rgb method="custom" r="1.000" g="1.000" b="1.000"/>
+        </color>
+        <color priority="2" name="All color separations" c="1.000" m="1.000" y="1.000" k="1.000" opacity="1.000">
+            <cmyk method="custom"/>
+            <rgb method="custom" r="0.000" g="0.000" b="0.000"/>
+        </color>
+        <color priority="3" name="Covering white" c="0.000" m="0.000" y="0.000" k="0.000" opacity="1.000">
+            <cmyk method="custom"/>
+            <rgb method="custom" r="1.000" g="1.000" b="1.000"/>
+        </color>
+        <color priority="4" name="Black" c="0.000" m="0.000" y="0.000" k="1.000" opacity="1.000">
+            <spotcolors>
+                <namedcolor>Black</namedcolor>
+            </spotcolors>
+            <cmyk method="custom"/>
+            <rgb method="custom" r="0.000" g="0.000" b="0.000"/>
+        </color>
+        <color priority="5" name="Black for rock" c="0.000" m="0.000" y="0.000" k="1.000" opacity="1.000">
+            <spotcolors>
+                <component factor="1" spotcolor="4"/>
+            </spotcolors>
+            <cmyk method="custom"/>
+            <rgb method="custom" r="0.000" g="0.000" b="0.000"/>
+        </color>
+        <color priority="6" name="Street infill" c="0.000" m="0.280" y="0.500" k="0.090" opacity="1.000">
+            <spotcolors knockout="true">
+                <component factor="0.5" spotcolor="11"/>
+            </spotcolors>
+            <cmyk method="spotcolor"/>
+            <rgb method="spotcolor" r="0.910" g="0.681" b="0.500"/>
+        </color>
+        <color priority="7" name="Street borderlines" c="0.000" m="0.000" y="0.000" k="1.000" opacity="1.000">
+            <spotcolors>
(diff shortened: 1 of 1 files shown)


symbol sets/src/ISOM2017_15000.xmap
```diff
@@ -0,0 +1,10029 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="6">
+    <notes>ISOM 2017
+Symbols for OpenOrienteering Mapper
+Created by Jukka Liikari and Håvard Tveite, IOF Map Commission
```
dg0yt: This looks like Jukka and Havard hold copyright and need to agree that OpenOrienteering may redistribute this file under its terms.

symbol sets/src/ISOM2017_15000.xmap
```diff
@@ -0,0 +1,10029 @@
...
+                <namedcolor>Purple</namedcolor>
+            </spotcolors>
+            <cmyk method="custom"/>
+            <rgb method="custom" r="0.850" g="0.000" b="1.000"/>
+        </color>
+        <color priority="1" name="White background" c="0.000" m="0.000" y="0.000" k="0.000" opacity="1.000">
+            <cmyk method="custom"/>
+            <rgb method="custom" r="1.000" g="1.000" b="1.000"/>
+        </color>
+        <color priority="2" name="All color separations" c="1.000" m="1.000" y="1.000" k="1.000" opacity="1.000">
```
dg0yt: This is not a regular color and shall not exist in Mapper file format.

yevhenmazur: Ok. I can see two ways from here.
The first one is:
1. We are asking for permission from @havatv and Jukka Liikari to distribute file isom-201x-finaldraft-11-2015-inofficialsymbols.omap under GNU/GPL v3 as a part of OpenOrienteering Project. And if they do not mind
2. I'm creating new pull request with this file + my commits
3. OpenOrienteering community is working and polishing this file

The second way is:
1. OpenOrienteering community is creating new symbol set, based on existing ISOM2000.

I don't really like the second way, because a lot of work already done. Many symbols was totally created from scratch, but licensing collision means that I can't just copy-paste them from one file to another.

dsc_0620-1

@dg0yt please approve one of the options before I do something else unnecessary

@999999333 - For your information

symbol sets/src/ISOM2017_15000.xmap
havatv: I created the Open Orienteering Mapper symbol set for ISOM2017 based on an OCAD symbol set that Jukka and several others have contributed to through the years.  I do not claim copyright.  The symbol set was distributed from the IOF to the member federations during the ISOM201X project without any copyright notice.
Please have in mind that the symbol set was created for demonstration purposes. It has not been thoroughly checked and may therefore contain serious mistakes.  It also does not reflect changes that were made before the publishing of the ISOM201X.
Please contact me on email if there are other questions regarding the copyright.

Håvard Tveite

havatv:
The name of the symbol set should preferably be "ISOM2017", and not "ISOM2017 15000", since the ISOM2017 symbol set was designed to be used without modifications for all map scales (a 1:10000 print of the map will be a pure enlargement of the 1:15000 version).  This also applies to the course overprint symbols.
For the ISOM2017, the mapper shall not have to worry about map scale before preparing for printing.

lpechacek:
FWIW, I'm sharing @dg0yt's opinion that '17 symbol set should be derived from the existing ISOM 2000 template.

havatv:
I agree that using the ISOM2000 template as the starting point would be the best approach.

dg0yt:
At this occassion, we should remove course setting symbols from the ISOM 2017 Map symbol set, and update the Course Design symbol set.

(Maybe we need to leave Purple and a symbol using this color in the Map symbol set, so that merging of files remains easy.)

yevhenmazur:
I'm joining to work on implementation based on existing ISOM2000 template, as discussed above.

lpechacek:
To avoid effort duplication... @yevhenmazur, are you working on the symbol definitions or shall I proceed? [In my set](https://github.com/lpechacek/mapper/blob/isom2017/symbol%20sets/15000/ISOM-2017_15000.omap) there is majority of man-made features still unfinished.

yevhenmazur:
I'm working on it. You can track the progress in my branch [isom2017](https://github.com/yevhenmazur/mapper/tree/isom2017). Today I'm going to finish all 2xx symbols; 3xx and 4xx - during the weekend.
The symbol set is just a .xml file under the version control, so we can use the power of git and work together. I've created new pull request (https://github.com/OpenOrienteering/mapper/pull/872) for taking this file in to upstream repository, but still no attention there.

Also please note - there is much more work than just finishing of man-made features. In 1xx group I have had to adjust 8 from 18 symbols. And I think these 8 symbols should be reviewed and approved by other person.

dg0yt:
#872 has attention (I get an e-mail for every post anyway), but I cannot comment unless I have enough time to review it. On the other hand, this is something which can be reviewed by a much wider audience than the normal contributions.

lpechacek:
@yevhenmazur thanks for taking the ISOM2017 symbol set further. I had a brief look at the symbols and color table and it looks good to me. I especially like the changes which originally slipped my attention like the possibility of rounded ends for passable cliff face. Let me know if you want to finish the initial draft yourself or want to offload parts of it on me.

FTR, today I looked into the element clipping options for boulder fields and stony ground. The best looking choice was (unsurprisingly) "center": 4270535adfc2a3e5cc7c48f63c9b764d31b338ed in lpechacek/isom2017

yevhenmazur:
I'm going to finish the symbol set tomorrow (the man-made features almost finished and this change will be pushed in a few hours). Nevertheless I want to ask you to review all modified by me symbols. The list of such symbols can be find in messages to my commits.
As well I'm not sure in description accuracy for every symbol...

yevhenmazur:
>The sizes of the overprinting symbols are given for the map scale 1:15 000. For larger map scales the symbols shall be enlarged proportionally (to 150% for 1:10 000, to 300% for 1:5 000).

@dg0yt, what is the reason to drop course setting symbols from the ISOM 2017 Map symbol set?
It is very conveniently quickly mark a start of next event directly on the map, or even plan single course for personal training...
And in the other hand, Course Design symbol set loses its value since [Purple Pen supports omap/xmap as a background map](http://purplepen.golde.org/change_summary.htm) directly.

dg0yt:
The reason for my proposal is "Separation of concerns". Map drawing is a different activity from course planning. I don't think it is inconvenient to use a different symbol set for planning a course. However, this is open to discussion.

Purple Pen is a Windows program. In the long run, we need course setting in Mapper.

PR #862, opened 2017-04-25
Area symbol: Prepare alternatives to pattern clipping

dg0yt:
Alternatives added in last commit (272e0b3).
Includes cleanup and modernization of this AreaSymbol and AreaSymbol::FillPattern.

### DIFF
--- a/src/core/symbols/area_symbol.cpp
+++ b/src/core/symbols/area_symbol.cpp
@@ -293,6 +293,7 @@ void AreaSymbol::FillPattern::createRenderables(const AreaRenderable& outline, f
 	
 	// Adjust extent wrt fill pattern size
 	auto extent = outline.getExtent();
+	auto point_extent = QRectF {};
 	switch (type)
 	{
 	case LinePattern:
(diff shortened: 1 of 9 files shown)


dg0yt:
Limitation of this implementation: It looks only at (the center of,  the points of) the extent rectangle of a fill pattern. This rectangle may inaccurately reflect (the center of, the points of) the actual graphical elements, especially when the pattern is composed of multiple elements.

In the following screenshot, every element is a distinct fill pattern. At least the "center" works fairly well. In the other non-clipping modes, there are missing and excess elements, due to the imprecision of to the extent rectangle (cf. red boxes). Solving this would need more sophisticated work with the graphical primitives.
alternatives-to-clipping

dg0yt:
Now handling each element in a pattern individually. And red boxes solved: Now looking at the actual coords. Dot/circle handled simplified (left, right, top, bottom coordinate). Test file attached.

[clipping-alternatives.omap.zip](https://github.com/OpenOrienteering/mapper/files/988987/clipping-alternatives.omap.zip)

PR #872, opened 2017-05-09
Add ISOM 2017 symbol set (work in progress)

yevhenmazur:
It is still in work-in-progress state, but I think we need this file in upstream repository (maybe not in master branch).

What is done: all landform symbols (1XX) have been reviewed/adjusted.
The symbols 103 104 104.0.1 105 106 108 111 112 were modified. It would be nice if they were viewed by another person.


lpechacek:
Quick comments on the latest commit. AFAIU the standard now defines only single "road" symbol. There is no longer distinction between "major" and "minor" road. That should be reflected in the symbol set by providing "502 Wide road, minimum width" plus "502.1 Dual carriage road, minimum width" and let users derive wider variants as they need.

Narrow ride (508) with the various outlines should IMHO draw over the background green in thickets.

Ad interaction between Paved area, with outline (501) and Wide road (502). Should the 50% brown road infill draw over the black Paved area outline? Yes would make combining these two symbols easier but would also require additional color.

Continuing the review.

yevhenmazur:
Symbol names for 502, 502.1 and 502.2 have been updated according to current ISOM in https://github.com/OpenOrienteering/mapper/pull/872/commits/3d2661aae4fd4eccf55d65d9ab53bc0c7b6fbde6

Could you please describe in more detail what do you mean talking about 508.

About 501 and 502 - I'll play with it. Now it is implemented the same as for ISSOM_2007, but one extra color it is not so heavy option, I think.

lpechacek:
> Narrow ride (508) with the various outlines should IMHO draw over the background green in thickets.

Current status is like this:
image

Narrow ride outline should overlay the green background:
image

Otherwise I see the outline of limited usability.

dg0yt:
Libor, I can merge this in a temporary branch and grant write access if you cannot push to the isom2017 branch on yevhenmazur/mapper.

Merging to a main branch (dev, master) can happen later.

yevhenmazur:
I've added @lpechacek and @999999333 as a collaborators in to my fork repository, just in case.

yevhenmazur:
> Ad interaction between Paved area, with outline (501) and Wide road (502). Should the 50% brown road infill draw over the black Paved area outline? Yes would make combining these two symbols easier but would also require additional color.

Fixed even without adding extra color
>Narrow ride outline should overlay the green background

Fixed in https://github.com/OpenOrienteering/mapper/pull/872/commits/13dbef0ed3c439da9cbdf64a73e79cdb7031b955

Not sure in the new color names... Green 20% and Green 50% already have 'Knockout' property, so Opaque Green 20% and Opaque Green 50% poorly characterizes the new colors.

lpechacek:
Created pull request with my changes targeted at yevhenmazur/isom2017. Although I've been granted write permission for the repository, I consider impolite to push changes without prior discussion. At this stage I'm fine with the color table and the (changed) symbol set.

yevhenmazur:
Sorry that I did not reply for a while... I have a few comments. I'll issue them in a few hours.

yevhenmazur:
Most of my comments are no longer relevant. I'm almost happy with current state.

Only the absence of a separate symbol for a reduced large boulder introduces confusion - it is not easy to use symbols correctly, if single symbol indicates different, but similar objects...

Also, I propose to switch default tag lines direction for symbols 201.0.1 (Impassable cliff, minimum size) and  202.1.1 (Cliff, minimum size)
2017-06-19-201107_scrot

dg0yt:
> Only the absence of a separate symbol for a reduced large boulder introduces confusion - it is not easy to use symbols correctly, if single symbol indicates different, but similar objects...

I looked into ISOM2017, and I support the proposal the idea of adding a separate symbol in order to "reduce the size of the symbol ... for some of the boulders".

> Also, I propose to switch default tag lines direction for symbols 201.0.1 (Impassable cliff, minimum size) and 202.1.1 (Cliff, minimum size)

Good point from the usability point of view. However, this is giving trouble during map conversion. It can be mitigated by providing the legacy version in a hidden symbol.

I will merge my pull requests, add some changes, push and trigger new unstable builds. This will become the baseline for further reviews and changes before 0.7.1

dg0yt:
>> Also, I propose to switch default tag lines direction for symbols 201.0.1 (Impassable cliff, minimum size) and 202.1.1 (Cliff, minimum size)
>
> Good point from the usability point of view. However, this is giving trouble during map conversion. It can be mitigated by providing the legacy version in a hidden symbol.

203 Rocky pit or cave has a similar orientation issue when the symbol shall point into the cave.

dg0yt:
>> Only the absence of a separate symbol for a reduced large boulder introduces confusion - it is not easy to use symbols correctly, if single symbol indicates different, but similar objects...
>
> I looked into ISOM2017, and I support the proposal the idea of adding a separate symbol in order to "reduce the size of the symbol ... for some of the boulders".

Looking at it again in the editor, I'm about to change my mind... I will try to cover it with changes in wording.

PR #873, opened 2017-05-12
Update man page

sikmir:


doc/man/Mapper.1
```diff
@@ -37,4 +82,4 @@ Extended documentation for
 .B Mapper
 is available from the program's help menu and also online at:
 .br
-<http://openorienteering.github.io/mapper-manual/>
+<http://www.openorienteering.org/mapper-manual/>
```
dg0yt: BTW the AUR package has an old (redirected) homepage link, too - cf. https://repology.org/metapackage/openorienteering-mapper/packages

sikmir: Updated.

PR #874, opened 2017-05-12
AboutDialog: Make copyright string translatable

sikmir:


dg0yt:
So far, I shied away from translating the copyright string because I want to avoid possible deviations in meaning caused by translation. I have no idea how many programs do that - my Firefox is in German locale, but about.licences is in English.

sikmir:
@dg0yt Actually many programs/products do that, for examples google by "Tekijänoikeus (c)", "Авторское право (c)" or "Kopirajto (c)". That seems more consistent to me. But up to you.

Symbian9:
[code]Copyright © ...[/code] is widely used in the world without translating (such as phisical constats e, g, ...)

I don't think this should be translated in other languages. All license documentation must be in English. For all other things there is Wikipedia and Internet.

PR #878, opened 2017-05-17
Esperanto translation for Qt

sikmir:


sikmir:
@dg0yt Could you trigger resource update on Weblate as well, please.

dg0yt:
Done.

PR #884, opened 2017-06-05
Merge ISOM2017 symbol set

dg0yt:
Merged work from existing ISOM and ISSOM symbol sets, and from pull requests by @yevhenmazur and contributions by @lpechacek.


dg0yt:
@yevhenmazur and @lpechacek, please review if your work is properly included and credited in my pull request. I'm about to finish improved CRT file support, so I want to merge this soon.

Regarding map conversion to ISOM, I consider adding some special symbols which for objects which need manual review after conversion. Applications include:
- ISOM2000 104 Slope line (ISOM2017: regular vs. form line)
- ISOM2000 106 Earth bank, very high (ISOM2017: earth bank vs. impassable cliff)
- ISOM2000 204 Pit (ISOM2017: combined with cave, must check orientation in Mapper)
- ISOM2000 302 Pond (ISOM2017: Impassabable body of water WITH bank line and minimum size)
- ISOM2000 50x Various road under construction (ISOM2017: removed)
- ISOM2000 531 Firing range (ISOM2017: removed)
- ISOM2000 532 Grave (ISOM2017: removed)

Using explicit, hidden symbols, it is possible to keep these symbol out of new maps, but highlight them as needing review when converting existing maps. (That is the reason why there are certain hidden symbols in the ISMTBOM symbol sets.)

lpechacek:
I've reviewed the symbol set and it looks good to me. Further I looked into the CRT. There I have a few comments.

- Agree that there should be (intermediate) replacement symbols for dropped entities. They carry information which should be retained in the map for post-processing.
- There should be two versions of ISOM2017 203 Rocky pit or cave, rotatable and non-rotatable, for the previous two ISOM2000 symbols. Use case is map rotation upon grivation changes.
- Major power line ISOM2000 511 should convert to 517 - the symbol **without** squares representing masts. 517 has about the same footprint and look as 511 had.
- Use case for CRT entries 702->702/Control point and 708->519/Passage is unclear to me.

dg0yt:
> Use case for CRT entries 702->702/Control point and 708->519/Passage is unclear to me.

702 was left to protect to protect the overprinting color.
708->519  was an automatic match by name which made it is way to the file by accident.

dg0yt:
Correction: 708->519 is not in the file. It is the default suggested match by name. The only fix is to not make any suggestions automatically.

Issue #606, opened 2016-02-25
Import OpenStreetMap data with styles


kolesar-andras:
Imported .osm files become gray lines. Points are not imported at all. Do you plan to extend OSM importer to read tags and apply styles to objects during import? Or will you accept a pull request that implements this feature?

If not, I should find or create a converter that creates .omap files from OSM data. Do you know such a converter? I don't want to reinvent the wheel if not necessary.

dg0yt:
In the stable releases, I would expect OSM vectors to become purple line, and tags are imported, too. However, there is nothing yet what would assign styles. 

In the unstable snapshots, OSM import might already use GDAL/OGR, and formatting would be according to the GDAL defaults (black, gray). Tags are partially imported, depending on [the configuration](http://gdal.org/drv_osm.html). This is work in progress. I'm about to add GDAL configuration options to the settings. In the end, I will probably remove Mapper's native support for DXF, GPX and OSM in favour of GDAL based support. This code directly creates first class Mapper objects (Map, Symbol, Object).

I gladly accept pull requests, but I'm unlikely to accept ad-hoc solutions which will be difficult to maintain or extend in the long run. When you want to work on assigning styles, we need to discuss the features and design:
 - Which type of style definition will we (re-)use?
   - Is it common?
   - Are the existing definitions for orienteering-style maps?
 - Can this feature be generalized for other vector formats?

My ideas include to expose the map objects to a Javascript environment where arbitrary user scripts could do what they want.

wanacode:
Someone wrote a basic bash script which may be a starting point for a converter https://github.com/OpenOrienteering/mapper/issues/178#issuecomment-135996873

dg0yt:
That script doesn't do much more than splitting osm files. It does neither assign existing symbols nor create new ones.

kolesar-andras:
Thank you for you answer. Planned JavaScript API is an interesting feature that would allow users to automate tasks. Howewer, a beginner user would rather need a one-step (builtin) solution for importing OSM data. Either by opening an .osm file downloaded from OpenStreetMap website or by selecting a bounding box of target area on a slippy map, just [like in JOSM](http://learnosm.org/images/josm/josm_download-dialog.png).

Styling could happen by a builtin but configurable JavaScript solution or a json/xml/etc configuration file. This solution can be even more general:

* not for only .osm files but any OGR data source that has OSM tags
* not only for OSM tags but mapping any combination of values in data source to styles

[...]

PR #892, opened 2017-06-18
Elaborate and leverage ObjectQuery and CRT file support

dg0yt:
Until now, CRT file support was limited to DXF import. These changes leverage ObjectQuery for powerful symbol replacement and CRT file support:
- CRT files may be loaded during symbol set replacement operations.
- CRT files may be loaded at any time, searching for symbol numbers or any tag contents. In the future, there could be complex queries, e.g. for complex OSM tag combinations.
- Symbol replacement/assignments may be saved to CRT files.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -64,6 +64,7 @@ set(Mapper_Common_SRCS
   core/objects/boolean_tool.cpp
   core/objects/object.cpp
   core/objects/object_query.cpp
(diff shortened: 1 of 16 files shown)


dg0yt:
With the proposed changes, complex queries can be applied to object tags (cf. tag selection), as needed for #606:

~~~
505    highway = track
504    highway = track AND (smoothness = good OR smoothness = intermediate)
...
~~~


osm-crt

PR #900, opened 2017-07-13
OSM import with GDAL (+CRT)

sembruk:
Some fixes.


Symbian9:
http://wiki.openstreetmap.org/wiki/IOF_mapping

PR #909, opened 2017-08-04
Issue warning for out of bounds objects during OCAD 8 export

lpechacek:
Problem reported at https://www.facebook.com/groups/485564718218028/1397750290332795/

### DIFF
--- a/src/fileformats/ocad8_file_format.cpp
+++ b/src/fileformats/ocad8_file_format.cpp
@@ -1533,7 +1533,8 @@ double OCAD8FileImport::convertTemplateScale(double ocad_scale)
 OCAD8FileExport::OCAD8FileExport(QIODevice* stream, Map* map, MapView* view)
  : Exporter(stream, map, view),
    uses_registration_color(false),
-   file(nullptr)
+   file(nullptr),
+   coord_bounds_warned(false)
 {
 	ocad_init();
 	encoding_1byte = QTextCodec::codecForName("Windows-1252");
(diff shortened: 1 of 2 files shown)


dg0yt:
Hm, this adds a warning. But what users need is the coordinates to be adjusted so that (at least parts of) the objects become visible?

lpechacek:
Strictly speaking, in the FB post the user wondered why is his map invisible in OCAD. In turn we both learned that OCAD viewport is artificially limited to 4x4 meters. There was no suggestion for Mapper to handle the export gracefully. As to my understanding, trying to resolve the situation is an extra step involving transformation of all map objects, recalculation of georeferencing parameters and checking that the transformed map fits into the limit. Honestly, I'd rather invest my limited time budget into proper georeferencing export (issue 596) and line vectorization (issue 833). That is why I implemented only the warning.

If you want, you can assign the above referenced issues to me.

dg0yt:
Okay, I can look into the transformation. I implemented something similar for OMAP import (partially for viewport reasons, too).
We may need a warning anyway, as soon as a map is larger than 4x4 m. But just a warning isn't enough, people tend to ignore them.

dg0yt:
Alternative (coord adjustment) in #916.

lpechacek:
Side note: OCAD must be facing the same issue. The Professional edition has 80x80m view port while the Standard edition and older OCAD releases are limited to 4x4m. http://www.ocad.ch/OCAD10/OCAD10_GettingStarted_English.pdf - warning paragraph in chapter Georeference the Map

lpechacek:
Superseded by #916.

dg0yt:
> Side note: OCAD must be facing the same issue. The Professional edition has 80x80m view port while the Standard edition and older OCAD releases are limited to 4x4m.

OCAD Pro and Std seem to have a tool for centering the map content, according to the OCAD wiki. It even states: "If a map should be visible in all OCAD editions then the entire map must fit in the 4x4m drawing area." Cf. http://www.ocad.com/wiki/ocad12/en/index.php?title=Map_Transform

Issue #910, opened 2017-08-06
Cut away tool broken in latest unstable


xc-racer99:
### Steps to reproduce
1. Draw a line symbol
2. Cover part of it with an area symbol
3. Attempt to use the "Cut away" tool and press enter

### Actual behaviour
The line symbol is entirely deleted.  

### Expected behaviour
Only the part of the line symbol that is covered should be deleted (as in previous versions, including 0.7.0)

### Configuration
Mapper Version: 0.7.1 (latest unstable - last git commit is https://github.com/OpenOrienteering/mapper/commit/c9e89e9131aef25d162df53944daf756f37e1c8a)
Operating System: Linux and Windows

### Additional info
Based on a git bisect, this regression appears to have been introduced in https://github.com/OpenOrienteering/mapper/commit/9bb9f2542b59cc66b1824765496af53f0183b2c6

999999333:
👍

PR #913, opened 2017-08-09
Fix Cut out tool

xc-racer99:
Broken since https://github.com/OpenOrienteering/mapper/commit/9bb9f2542b59cc66b1824765496af53f0183b2c6

Basically, in finish(), new_objects didn't have anything in it so the cut objects weren't being added back in.  I'm definitely not a C++ expert, but this fixes the issue I was seeing.

Solves #910

### DIFF
--- a/src/tools/cutout_tool.cpp
+++ b/src/tools/cutout_tool.cpp
@@ -177,6 +177,7 @@ struct PhysicalCutoutOperation
 	, cutout_object(cutout_object)
 	, cut_away(cut_away)
 	{
+		new_objects = new std::vector<PathObject*>();
 		add_step = new AddObjectsUndoStep(map);
 		delete_step = new DeleteObjectsUndoStep(map);
 	}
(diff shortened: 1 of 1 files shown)


dg0yt:
Your bisecting already gave the right clue, however I came to another solution (which is not pushed yet). The problem is that a function object (here: `PhysicalCutoutOperation`) is copied when a `std::function` object is created (here: implicitly in the call to `applyOnAllObjects`). But `PhysicalCutoutOperation` is not implemented in a copyable way... 

Your proposal solves that by turning `new_objects` into a pointer, thus making all copies share the same list of objects. But it adds a memory leak. In modern C++ you generally avoid error-prone direct uses of `new` and `delete`. (`PhysicalCutoutOperation` is not a good example in this regard).

My solution is to explicitly forbid copying (to make wrong use impossible) and to explicitly use a "copyable reference" in the call to `applyOnAllObjects`.

dg0yt:
(And of course I found a few more instances of `applyOn...` where copying the functor shall be avoided. I was not aware of copying taking place.)

xc-racer99:
Great, thanks for the fix and the explanation of what was wrong with mine.

Issue #750, opened 2016-09-02
Position of line pattern changes at export to .ocd format


dl3sdo:
### Steps to reproduce
1. Open sample complete map.omap
2. Save as .ocd
3. Open saved .ocd

### Actual behaviour
The magnetic north lines are drawn as a pattern in Mapper.
At the exported .ocd map the lines are shifted to the right, thus being
unaligned with the north arrows.

### Expected behaviour
Lines should stay their original position

### Configuration
Mapper Version: 0.6.5
Operating System: Win7

dg0yt:
Well, I noticed this when dealing with #749. Unfortunately, the ocd format seems to offer no way to preserve the position of object patterns. Mapper already emits a warning on ocd export.

In the complete map example, it is the north line pattern which is not used like that with ocd symbol sets. You could replace this very easily. However, I used the ability to determine the pattern position very heavily in a vineyard map, to improve readability. Here I don't see a solution at all.

PR #916, opened 2017-08-10
OCAD8FileExport: Move coords into OCAD 8 drawing area

dg0yt:
Do not just issue a warning, but move coordinates as needed to make objects visible in the bounds of the OCAD drawing area.

### DIFF
--- a/src/fileformats/ocad8_file_format.cpp
+++ b/src/fileformats/ocad8_file_format.cpp
@@ -1905,6 +1905,16 @@ MapCoord OCAD8FileExport::calculateAreaOffset()
 			addWarning(tr("Some coordinates remain outside of the OCAD 8 drawing area."
 			              " They might be unreachable in OCAD."));
 		}
+		
+		if (area_offset.manhattanLength() > 0)
+		{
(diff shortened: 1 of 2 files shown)


dg0yt:
(I didn't test with templates.)

lpechacek:
I tested with a georeferenced map with templates (attached) by adding a text object 3000 mm away from the origin. Export went well, file was displayed correctly in OCAD 8. Adding another object on the opposite side yielded warning about out of bounds object and map was not transformed. Code looks good to me as well, modulo the whitespace cruft. Thanks!
[out-of-bounds-test.zip](https://github.com/OpenOrienteering/mapper/files/1217348/out-of-bounds-test.zip)

dg0yt:
3cbc0a4 fixes a use-after-delete error (detected by ubsan) which was introduced by the usage of object duplicates. This was uncovered by trying your test file. Thanks for sharing.

There is one issue with moving coordinates like this: In OCD file format, we cannot adjust the offset of patterns (undergrowth etc.) per object AFAIK. So during this export, pattern elements may move relative to its surroundings. OTOH if offsets were set in Mapper (Rotate pattern tool), this effect did already exist in OCD export (#750).

PR #917, opened 2017-08-10
Minor fixes in symbol sets

sikmir:
- fixed extra space at the end of the source string (bulk update)
- fixed missing dot at the end of the source string (bulk update)
- added da,eo,nl to `map_symbols_translations`


translations/CMakeLists.txt
```diff
@@ -57,9 +57,12 @@ set(Mapper_translations_template
 # Translation files which are updated by SymbolSetTool.
 set(map_symbols_translations
   map_symbols_cs.ts
+  map_symbols_da.ts
   map_symbols_de.ts
+  map_symbols_eo.ts
```
dg0yt: This file is not in Git and thus breaks the build.

sikmir: Yes, sync with Weblate is required.

PR #919, opened 2017-08-12
Set version to 0.8.0

dg0yt:
Since 0.7.0, there has been a lot of new features:

- modified pattern clipping (change in file format)
- ISOM 2017 symbol set
- symbol set translations
- Find dialog with object queries
- advanced CRT file support with object queries
- CRT files for ISOM2017 and OSM
- modified dash symbol scaling (change in file format)

plus a lot of refactoring and fixes. Cf. https://github.com/OpenOrienteering/mapper/issues?utf8=%E2%9C%93&q=milestone%3Av0.7.1%20. (Documentation is lagging as usual...)

So I would suggest to use 0.8.0, and to repeat the deprecation warnings (file formats, OS X 10.7) rather than activating these changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ include(FeatureSummary)
 
 # Project declaration
 
-project(Mapper VERSION 0.7.1 LANGUAGES CXX C)
+project(Mapper VERSION 0.8.0 LANGUAGES CXX C)
 
 set(Mapper_COPYRIGHT "(C) 2012-2017 The OpenOrienteering developers")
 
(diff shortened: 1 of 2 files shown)


Symbian9:
> Since 0.7.0, there has been a lot of new features... So I would suggest to use 0.8.0

Fully agree with You, dg0yt!

sikmir:
I'm going to create another one PR prior 0.8.0 with various fixes of symbol sets.

dg0yt:
Okay, the release will have to wait for some more testing anyway. The current dev branch - and the unstable builds - contain a big chunk of non-functional changes (mostly modernization, but also corrections) which I would like to have in 0.8.0 to ease further development. So testing should focus on dev. Only if it turns out be too broken, 0.8.0 will be based on the current state of the master branch. (In the worst case, I would reset even reset dev to master.)

dg0yt:
Modernization changes: https://github.com/OpenOrienteering/mapper/compare/0d74462...f9e025b

And first result of using keyword override: 0fef1248

PR #922, opened 2017-08-18
doc/man/Mapper.1: Add ISOM 2017 symbol set

sikmir:


doc/man/Mapper.1
```diff
@@ -20,7 +20,7 @@ It comes with predefined symbol sets implementing the IOF standards:
 .sp -1
 .IP \(bu 2.3
 .\}
-ISOM 2000 (forest) (1:15000, 1:10000)
+ISOM 2000/2017 (forest) (1:15000, 1:10000)
```
dg0yt: I wonder if 2000/2017 is the best way to refer to both versions (alternative: separate ISOM2017 entry).
And not that I chose to consistently drop the space in ISOM2017 et al. 
However, IOF is not 100% consistent.

sikmir: Updated.

PR #926, opened 2017-08-19
Add missing Qt messages

sikmir:


sikmir:
@dg0yt By the way, do you know a way how to extract qt translations, limited to contexts relevant to mapper? In other words, how to regenerate qt_template.ts?

dg0yt:
I ended up creating the file by hand. For a more systematical approach, you could build Mapper with ```Mapper_DEBUG_TRANSLATIONS``` defined. Then class ```RecordingTranslator``` (in src/util/recording_translator.h) would catch all missing translations when you do a representative piece of work in Mapper. However, it does not record at the moment - it just prints the data to the console.

PR #928, opened 2017-08-19
Simplify logic expressions

sikmir:
Changelog:
- replace '(ptr1 && !ptr2) || (!ptr1 && ptr2)' with 'bool(ptr1) != bool(ptr2)'
- replace '!A || (A && B)' with '!A || B'
- protect copy operators against self-assignment

### DIFF
--- a/src/util/matrix.h
+++ b/src/util/matrix.h
@@ -77,6 +77,9 @@ public:
 	/** Assignment */
 	void operator=(const Matrix& other)
 	{
+		if (this == &other)
+			return *this;
+		
 		delete[] d;
 		n = other.n;
 		m = other.m;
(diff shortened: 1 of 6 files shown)


src/util/matrix.h
```diff
@@ -77,6 +77,9 @@ class Matrix
 	/** Assignment */
 	void operator=(const Matrix& other)
 	{
+		if (this == &other)
+			return *this;
```
dg0yt: Triggered FTBFS.
This ```operator=``` was returning ```void```, not ```Matrix&```. I'm fixing it.

sikmir: Sure, my mistake, thanks!

PR #930, opened 2017-08-20
Fix spacing between numbers and units of measure

sikmir:
With regards to typography, English style guides say:
- [percent](https://en.wikipedia.org/wiki/Percent_sign#Spacing) (%) and [degree](https://en.wikipedia.org/wiki/Degree_symbol#Typography) (°) signs follow the number without any space
- all other units of measure (mm, min, etc.) &mdash; with space

I think the same rules should apply for UI output, but it's discussable.

TODO: other languages have other rules at least for spacing in front of the percent sign, so ideally it should be moved to translations.

Examples (before):
screenshot_a1
screenshot_a2
Examples (after):
screenshot_b1
screenshot_b2


dg0yt:
Looks to me like this really needs to be addressed in translation. OTOH there is no need to hurry.

dg0yt:
From the Wikipedia links, I see that we do not only have to deal with the space, but also with suffix vs. prefix:
- English: 5%
- German: 5 %
- Turkish: %5

To handle this in translation, we could use ```Unit::tr("5%")```. The translator would see "5%" and use the proper style (e.g. "5 %"). The calling function would need to lookup the location of '5' and then set prefix and suffix (spinbox), or replace '5' with the actual number for normal output.

PR #962, opened 2017-09-13
PrintWidget: dpi_combo fix

sembruk:
Сhange the resolution when user select one of the default values.


dg0yt:
Merged with 9da3132, thanks.

Issue #398, opened 2014-09-16
line segment length and angle on mobile version


dg0yt:
**krticka reported on Sourceforge [[tickets:#398](https://sourceforge.net/p/oorienteering/tickets/398)]:**

One important function I am missing for mapping. Sometimes straight linear objects (narrow rides, minor water channels) are not visible in orthophoto, but I can find them in terrain, hidden under the trees. When I was mapping on desk, I measured directions of these straight objects, and draw their direction on the desk. Especially in case of many straight narrow rides it was helpful, you don't need to walk them all. 
When I am drawing in mobile version I am missing such functionality. I can measure direcions (angle) of narrow rides with sighting compass, but how to precisely draw it in mobile Mapper? Solution can be to show length and angle of actual line segment when drawing, see attachment (sample is from desktop version).

![segmentlengthangle](https://cloud.githubusercontent.com/assets/13567791/9568140/c0b88ee4-4f41-11e5-81ed-6ec98fa61168.gif)

Osin01:
Hi, I would also took advantage of this functionality. I am used to measure directions and distances precisely by compass with optical visor and laser range finder and draw it on a desk with protractor. I measure not only linear objects, but also distant point objects.  It is possible to make an accurate map even without good templates. It's actually pretty quick because I do not have to perform additional repairs of drawing. In Ocad I can use numeric mode of drawing, but as proposed by krticka, viewing angle and distance lines when drawing a line would be propably quicker on tablet. It should be enough to display integers for reasonable accuracy. If it was easier, direct input of distance and angles would also be very helpful.

krticka:
Well, I must say that still I would like to see this functionality implemented (and quite many other users as I know). For example last two days I was working in valley with problematic GPS signal and with many features. Almost no information from orthophoto. The result is that I will need to switch to classic way of mapping with pens and mylar as I am not able to process this area properly with GPS signal only. Precise lengths and angles are still needed in my point of view.

langrjan:
I have to agree with Osin01 and krticka. I moved from traditional way to digital way of mapping last year. Mentioned functionality is missing me. I consider such tool as very crucial and useful for field work mapping. There are many situations when GPS signal has problematic or unsufficient quality and further measurements like distance and azimuth between objects are needed.

akom5224:
[...]

PR #985, opened 2017-10-06
Show line segment length and azimuth while drawing

lpechacek:
The feature can be activated by pressing Ctrl-Space or "Info" button in mobile UI.

References: Issue #398

### DIFF
--- a/src/tools/draw_path_tool.cpp
+++ b/src/tools/draw_path_tool.cpp
@@ -46,6 +46,7 @@ DrawPathTool::DrawPathTool(MapEditorController* editor, QAction* tool_button, bo
 , finished_path_is_selected(false)
 , cur_map_widget(mapWidget())
 , angle_helper(new ConstrainAngleToolHelper())
+, azimuth_helper(new AzimuthInfoHelper(cur_map_widget, active_color))
 , snap_helper(new SnappingToolHelper(this))
 , follow_helper(new FollowPathToolHelper())
 , key_button_bar(nullptr)
@@ -61,6 +62,7 @@ DrawPathTool::DrawPathTool(MapEditorController* editor, QAction* tool_button, bo
 	following = false;
 	picking_angle = false;
 	picked_angle = false;
+	show_azimuth_info = false;
 	draw_dash_points = false;
 	shift_pressed = false;
 	ctrl_pressed = false;
@@ -87,6 +89,7 @@ void DrawPathTool::init()
 		key_button_bar->addPressKey(Qt::Key_Return, tr("Close"));
 		key_button_bar->addModifierKey(Qt::Key_Shift, Qt::ShiftModifier, tr("Snap", "Snap to existing objects"));
 		key_button_bar->addModifierKey(Qt::Key_Control, Qt::ControlModifier, tr("Angle", "Using constrained angles"));
+		key_button_bar->addPressKeyWithModifier(Qt::Key_Space, Qt::ControlModifier, tr("Info", "Show segment azimuth and length"));
 		key_button_bar->addPressKey(Qt::Key_Space, tr("Dash", "Drawing dash points"));
 		key_button_bar->addPressKey(Qt::Key_Backspace, tr("Undo"));
 		key_button_bar->addPressKey(Qt::Key_Escape, tr("Abort"));
@@ -443,8 +446,13 @@ bool DrawPathTool::keyPressEvent(QKeyEvent* event)
 		return true;
 		
 	case Qt::Key_Space:
-		draw_dash_points = !draw_dash_points;
-		updateStatusText();
+		if (event->modifiers() & Qt::ControlModifier)
+			show_azimuth_info =! show_azimuth_info;
+		else
+		{
+			draw_dash_points = !draw_dash_points;
+			updateStatusText();
+		}
 		return true;
 		
 	case Qt::Key_Control:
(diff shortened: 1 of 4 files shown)


dg0yt:
It needs some more work before it becomes really useful with touch input (Android) where we track movements only with finger=button down. However, I need to somewhat cleanup in the tool classes first, any maybe make the state and state changes more explicit, instead of burried in dozens of bools and if/else.

lpechacek:
The code looks much neater now and I like that map rotation is taken into account. I have a comment though.

Info deactivation in finishDrawing() and abortDrawing() was intentional. I use the display only after taking azimuth using [hand-bearing compass](http://www.suunto.com/cs-CZ/Products/Compasses/Suunto-KB-14/Suunto-KB-14360R1/) and draw single straight line segment. I don't need it otherwise, therefore automatic deactivation is convenient for my use case. I'd be happy to hear opinions from more experienced mappers here. @krticka, @ollesmaps, @yevhenmazur, @zlenhart, @akom5524 what is your preference? Should the azimuth display deactivate upon finishing the line or should it remain active until you click the button again?

Regarding the field experience, I'm happy with the way the feature works on Nvidia Shield tablet with stylus. The only downside I see now is that the mobile UI button does not reflect the azimuth display state. That's, however, the same as with the Dash button, which retains its look regardless the dash point status.

dg0yt:
> Info deactivation in finishDrawing() and abortDrawing() was intentional. 

I was expecting something like that. Is it disturbing to leave it turned on (until explicitly deactivated)?

> The only downside I see now is that the mobile UI button does not reflect the azimuth display state. That's, however, the same as with the Dash button, which retains its look regardless the dash point status.

That's a short coming of the KeyButtonBar as it is now. Maybe to be fixed now..

lpechacek:
> I was expecting something like that. Is it disturbing to leave it turned on (until explicitly deactivated)?

I need to test the current version in the field before answering your question. But as adding the automatic deactivation is trivial, the feature can be made configurable and there was no immediate feedback from other users, I'm for merging the code as it is now.

dg0yt:
Okay.

dg0yt:
Still need to update the azimuth info when pressing Backspace on a curved segment.

krticka:
On desktop seems to work OK, thank you!
@lpechacek Is there latest unstable build for android available? I will test it in terrain.

dg0yt:
I hope I can provide a new 0.8.0 beta release in the next days.

lpechacek:
> @lpechacek Is there latest unstable build for android available? I will test it in terrain.

I'm using [http://liborpechacek.wz.cz/oomapper/mapper-issue398-g0721f412f09c.apk](http://liborpechacek.wz.cz/oomapper/mapper-issue398-g0721f412f09c.apk) for my field work.

akom5224:
Hello,

thanks from me as well for implementing this tool!

Just a brief reply to the question on automatic deactivation of the tool 
(even though I would not call myself as "more experienced mapper" ;-)). 
I think that both (automatic deactivation or not) are useful in 
different situations and the preference will highly depend on a user. My 
(relatively mild) preference is NOT to deactivate it automatically. The 
reason is that when I want to draw azimuth lines then often I need to 
draw more of them (e.g., to fix a point object from two different 
positions) and automatic deactivation would be annoying in such 
situations. Once I am done with azimuth lines, I can switch it off 
manually and find this more handy than to switch it on for each single line.

Best wishes,
Arnost (akom5224)


Libor Pecháček schreef op 10/31/2017 om 08:41 PM:

lpechacek:
Thanks for describing your use case, Arnošt!

A brief update from my side. I spent Sunday morning in the field, with Mapper built from commit 0721f41, and found current behavior mostly OK. The only annoying part was that I didn't see whether the azimuth display is on or off before starting a line. That part is resolved in commit 333fd04 and I'm going to test the updated version.

yevhenmazur:
Could you please provide the actual build? I want to join to the testing of this feature.

lpechacek:
Just build today's master (214afa6abc18) for my testing: http://liborpechacek.wz.cz/oomapper/mapper-issue398-g214afa6abc18.apk. HTH

lpechacek:
@krticka, @yevhenmazur, go ahead and grab the official build at https://github.com/OpenOrienteering/mapper/releases/tag/v0.7.92. I'm going to delete the development builds soon.

krticka:
Works OK for me!
Probably info about angle can be useful also when you want to repair direction/rotation of existing object. Now there is no option how to find it as measure tool shows only length for line objects, perimeter and area for polygon object and nothing for point object.

lpechacek:
I'm also missing the azimuth display in edit mode. Unfortunately, the code structure for path editor is very different from the draw path tool so I had no good idea how to stick the azimuth display into it.

dg0yt:
Some tools are really ugly and difficult to extend. This also blocks better touch input handling. I started porting DrawPathTool to use explicit states and input/action handling (state machine) as part of merging the azimuth info contribution, but I decided to hold the more challenging parts back for integration after v0.8.0.

ollesmaps:
Here I come for testing. I can't find the Info button that shall activate it though (Android). Thanks for hint.

krticka:
Are you using beta? [https://github.com/OpenOrienteering/mapper/releases/tag/v0.7.92](https://github.com/OpenOrienteering/mapper/releases/tag/v0.7.92)

ollesmaps:
Yes

akom5224:
Info button appears on bottom of the map if you (a) choose a line object 
and (b) choose a straight line mode of drawing. It seems that it then 
remains activated even for all subsequent line objects (till deactivated).


Aleš Hejna schreef op 11/30/2017 om 12:46 PM:

ollesmaps:
Great enhancement! Thank you!
It best works with a special help symbol (thin purple line?) that you stick in your location and draw there and back, there and back till you receive your angle acquired by compass.  Than copy the angle with the right symbol.
If it worked in Edit mode this "thereandbacking" would be unneeded. 

I have not used it during fieldwork though. Will report later.

ollesmaps:
Now I see the numbers get activated also when drawing dragging (thanks @zlenhart for pointing it out). Brilliant!

Zerbembasqwibo:
Yes close it.

PR #1009, opened 2017-11-11
Change package list to space separated and mark as code block

sgnewson:
A space separated list is easier to copy-paste into a terminal, whereas the comma separated list would not run (for example with `sudo apt install cmake, ...`


dg0yt:
I like the text document do be readable, with these items grouped logically. (And personally, I like to review more closely what I install on a computer rather than copy-and-pasting what doesn't fit on the screen.)
But this has been proposed before, and I see the benefit. How about preceeding line breaks with backslashes?

PR #1010, opened 2017-11-12
Add QtCreator *.autosave to .gitignore

sgnewson:


dg0yt:
Is this an issue with a current version of Qt Creator? I can't remember having seen .autosave files.

sgnewson:
I'm using:
```Qt Creator 4.4.1
Based on Qt 5.9.2 (GCC 5.3.1 20160406 (Red Hat 5.3.1-6), 64 bit)

Built on Oct 4 2017 04:14:10

From revision 6afdb8bdf9```

on Ubuntu 16.04

dg0yt:
Ah, this from the About dialog. I have got the same revision, and the headline says Qt Creator 4.4.1.
I'm still surprised that I don't see .autosave files.

dg0yt:
I verified that Qt Creator creates temporary .autosave files. But these files are removed as soon as the file is saved regularly. That's the reason why I didn't notice them: When I look at git, almost everything is properly saved, because this is done at every build. That's also the reason why I will not merge this pull request:
The presence of a .autosave file indicates that there are unsaved changes. So the presence of a .autosave file strongly indicates that you will miss some of the changes you made. You will loose this hint when we add .autosave to .gitignore.

In addition, most changes are modifications to existing files. The `-u` parameter to `git` easily prevents accidently adding local files which are new or not meant to be committed. I only drop `-u` if I really want to add a new file.

sgnewson:
Makes sense, thanks for the explanation.

PR #1013, opened 2017-11-23
Fix Swiss CH1903 OCAD zone import

lpechacek:
It is unclear where did the OCAD code 14/000 come from. Swiss Grid / CH1093 is assigned 14/001 these days.


dg0yt:
Merged.

Issue #825, opened 2017-01-06
"Tag window" - unified window for Tag Editor & Tag Selector widgets


Symbian9:
### Steps to reproduce
1. Run OOM
2. Create or open map
3. Open "Tag Editor" & "Tag Selector"

### Actual behaviour
Currently there are two different widgets for working with objects tags

### Expected behaviour
Unify those widgets in single "Tag window" widget (like "Symbol window", "Template Window", "Color window")

### Configuration
Mapper Version: 0.6.7
Operating System: all supported

P.S.: Also check widget naming because currenty it's "Tag Selection" and "Tag editor" names, but in manual there are "Tag Selector" and "Tag Editor"
* http://www.openorienteering.org/mapper-manual/pages/

dg0yt:
Note that you may already merge the windows to a tabbed group (unless you are on an old system) by drag-and-drop of one window onto the other:

![unified-window](https://cloud.githubusercontent.com/assets/13567791/23586778/002dc580-019d-11e7-9956-0bcc8b6308d7.png)

dg0yt:
The tag selector may become fairly obsolete by the new find dialog which accepts more complex queries, e.g. highway = "track" AND (foo = bar OR x != y).

Symbian9:
Has you a new UI mockup for it?

dg0yt:
It is in the "unstable" builds...

Symbian9:
@dg0yt, fix menu entry for "Tag Editor" & "Tag Selector"

Symbian9:
@dg0yt, in 'unstable' builds menu entry for "Tag Editor" & "Tag Selector" still not fixed.

Is it related to locales or is it mistake in sources?

Symbian9:
@dg0yt, please, fix this issue in menu and widget naming!
![Tag Editor & Tag Selector](http://hnng.moe/f/WaF)

PR #1015, opened 2017-11-23
Fix tag's widgets naming

Symbian9:
Close https://github.com/OpenOrienteering/mapper/issues/825

### DIFF
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -905,9 +905,9 @@ void MapEditorController::createActions()
 	open_template_act = newAction("opentemplate", tr("Open template..."), this, SLOT(openTemplateClicked()), nullptr, QString{}, "templates_menu.html");
 	reopen_template_act = newAction("reopentemplate", tr("Reopen template..."), this, SLOT(reopenTemplateClicked()), nullptr, QString{}, "templates_menu.html");
 	
-	tags_window_act = newCheckAction("tagswindow", tr("Tag editor"), this, SLOT(showTagsWindow(bool)), "window-new", tr("Show/Hide the tag editor window"), "tag_editor.html");
+	tags_window_act = newCheckAction("tagswindow", tr("Tag Editor"), this, SLOT(showTagsWindow(bool)), "window-new", tr("Show/Hide the tag editor window"), "tag_editor.html");
 
-	tag_select_window_act = newCheckAction("tagselectwindow", tr("Tag Selection"), this, SLOT(showTagSelectWindow(bool)), "tag-selector", tr("Show/Hide the tag selection window"), "tag_selector.html");
+	tag_select_window_act = newCheckAction("tagselectwindow", tr("Tag Selector"), this, SLOT(showTagSelectWindow(bool)), "tag-selector", tr("Show/Hide the tag selector window"), "tag_selector.html");
 	
 	edit_tool_act = newToolAction("editobjects", tr("Edit objects"), this, SLOT(editToolClicked()), "tool-edit.png", QString{}, "toolbars.html#tool_edit_point");
 	edit_line_tool_act = newToolAction("editlines", tr("Edit lines"), this, SLOT(editLineToolClicked()), "tool-edit-line.png", QString{}, "toolbars.html#tool_edit_line");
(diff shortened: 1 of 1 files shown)


src/gui/map/map_editor.cpp
```diff
@@ -905,9 +905,9 @@ void MapEditorController::createActions()
 	open_template_act = newAction("opentemplate", tr("Open template..."), this, SLOT(openTemplateClicked()), nullptr, QString{}, "templates_menu.html");
 	reopen_template_act = newAction("reopentemplate", tr("Reopen template..."), this, SLOT(reopenTemplateClicked()), nullptr, QString{}, "templates_menu.html");
 	
-	tags_window_act = newCheckAction("tagswindow", tr("Tag editor"), this, SLOT(showTagsWindow(bool)), "window-new", tr("Show/Hide the tag editor window"), "tag_editor.html");
+	tags_window_act = newCheckAction("tagswindow", tr("Tag Editor"), this, SLOT(showTagsWindow(bool)), "window-new", tr("Show/Hide the tag editor window"), "tag_editor.html");
```
dg0yt: Actually, capitalization was correct here. It is not a headline. It is consistent with all other entries.

src/gui/map/map_editor.cpp
```diff
@@ -905,9 +905,9 @@ void MapEditorController::createActions()
...
 
-	tag_select_window_act = newCheckAction("tagselectwindow", tr("Tag Selection"), this, SLOT(showTagSelectWindow(bool)), "tag-selector", tr("Show/Hide the tag selection window"), "tag_selector.html");
+	tag_select_window_act = newCheckAction("tagselectwindow", tr("Tag Selector"), this, SLOT(showTagSelectWindow(bool)), "tag-selector", tr("Show/Hide the tag selector window"), "tag_selector.html");
```
dg0yt: This has the wrong capitalization, given its context.

Apart from capitalization, I'm reluctant to accept this change because a) we would need to update (source strings in) translations at the same time, b) I still think the "Tag selector" is better made an extra feature of the new (v0.8) "Find objects" dialog, and c) I have no clue what native speakers (English) consider a better match.

src/gui/map/map_editor.cpp
```diff
@@ -2068,7 +2068,7 @@ void MapEditorController::createTagSelector()
 	auto tag_widget = new TagSelectWidget(map, main_view, this);
 	tag_select_dock_widget = new EditorDockWidget(tr("Tag Selector"), tag_select_window_act, this, window);
 	tag_select_dock_widget->setWidget(tag_widget);
-	tag_select_dock_widget->setObjectName(QString::fromLatin1("Tag selection dock widget"));
+	tag_select_dock_widget->setObjectName(QString::fromLatin1("Tag selector dock widget"));
```
dg0yt: This is not relevant for the user interface, only needed for properly storing/restoring window state.

src/gui/map/map_editor.cpp
Symbian9: Are there similar tools in JOSM, Merkaartor, QGIS? Maybe, this feature in those GIS already has some better & more correct name.

Symbian9: > I have no clue what native speakers (English) consider a better match.

Please, ask it on "StackExchange/GIS"
* https://gis.stackexchange.com/

dg0yt: Again, my proposal is to remove this menu item and the dock widget, and make it a feature of the "Find objects" dialog. It would call it query editor there. It is almost finished, because I wanted to do this in v0.8.0:

query-editor

Symbian9: > my proposal is to remove this menu item and the dock widget, and make it a feature of the "Find objects" dialog. It would call it query editor there.

OK, I agree then.

But, maybe, also rename "Tag Editor" to "Tag window".

P.S.: Or much better use "widget" instead of "window": "Colors widget", "Symbols widget", "Tags widget", etc.

lpechacek:
> Are there similar tools in JOSM, Merkaartor, QGIS?

FWIW, JOSM search by tag value is just "Search".
image

It looks to me like dg0yt's approach is in line with the OSM way.

Issue #1005, opened 2017-11-08
Android: Definable "mapping scale" for field work


mlerjen:
### Background
-> The possibility to zoom in and out is one of the advantages of mapping with a smartphone/tablet. Details on the templates can be zoomed in if necessary. At the same time this leads to a often changing "mapping scale" during field work. A least using the pinch move to zoom in and out it is not possible to return to the "constant mapping scale" anymore. So by the way a fixed mapping scale is one of the big advantages of the traditional mapping with pencil and Mylar. The mapper has the possibility to develop a feeling for distance and object density on its field notes with a constant mapping scale.

### Feature request
-> The possibility to define a "mapping scale" in the maps setting.
-> As zooming in still shall be possible, an easy way to return to this zoom level shall be offered.
-> Being on or off mapping scale shall be indicated to the mapper visually.

ollesmaps:
That would be helpful. Could be combined with zoom in on gps position #949.

JHuberman:
I believe a user definable zoom level would be very useful as mentioned above.  "The Zoom+" button long press offers a 1x and 2x choice.  Adding a user configurable choice would be useful to mappers who prefer a different working scale.  Or as an alternative all three choices would be user configurable.

lpechacek:
I'm of two minds here. On one hand, Mapper is supposed to be beginner-friendly. That IMHO mandates a simple user interface, straightforward use cases, and minimal configuration. On the other hand, we've got professional mappers who use the application. There, small details matter and need to be adjustable for the sake of efficiency. To name a few, the possibility to quickly show and hide templates, selection of frequently used symbols, and return to the mapping scale on a button press. These details have multiplicative effects on the total time spent in the field and in the end, on the economy of the project.

I'd personally welcome a configurable mapping scale. I originally introduced the 1x and 2x zoom levels but now I spend most of my time working at 4x and 5.66x. So, I'd immediately make use of this enhancement. But the additional configuration bit might feel overwhelming to newcomers. Any ideas on how to tackle the growth in capabilities and at the same time keep the learning curve reasonably flat?

dg0yt:
I tend to suggest press-and-hold to trigger configuration options. This even allows discoverability once the user learned the pattern (like learning shortcuts which are listed with menu items).

[...]

Issue #952, opened 2017-09-05
Android GUI - button clashes


ollesmaps:
Some buttons are too similar or on a similar place. And quite often I make a mistake when choosing a wrong button.  I assume these could be easy fixes before doing a bigger [overhaul](https://github.com/OpenOrienteering/mapper/issues/534).

**Close file x Pan map** - too similar buttons and on a similar place (both in corners)
**Solution** -  Put the Close file button to the upper left corner (next to Grid, not in the very corner). This way you will not accidentaly close file. Now it sometimes happens when handling tablet (it is on the edge where you usually hold tablet). Measurement button could be in the upper right corner instead (this button is touch safe).

**Delete x Close file** - too similar buttons close to each other
**Solution** - Differentiate the Delete object icon. Rubber eraser? Or just thinner cross.

**Duplicate x Templates** - too similar buttons close to each other
**Solution** - Differentiate the Template icon - adding colors to the 3 layers should do. Move the Template icon to upper left corner (it belongs more there than to editing).

**Connect paths x Convert to curves x Simplify path** - too similar buttons close to each other
**Solution** - Differentiate the  Connect paths icon, it does not have to be circular. Two lines with gab in between for example. Convert to curves - twisted snake for example. 

**Show whole map** - this button shall be rather in the down left corner, where zooming occurs.

 I can help with the icons design. Where can I volunteer for that?

---------

Is there an app that can trace touches, clicks, movements? I can try that and have some "scientific" data how OOMapper app is being used.

The order of buttons could be optional, so you can arrange buttons according to your needs. For example when there is an overflow in the upper right corner I would rather hide other buttons and other have visible. However it might be different user to user.

Symbian9:
> Is there an app that can trace touches, clicks, movements? 

@ollesmaps, look like Android has inbuilt functionality for that
* https://medium.theuxblog.com/enabling-show-touches-in-android-screen-recordings-for-user-research-cc968563fcb9

dg0yt:
The Android user interface is still very much the initial proof of concept, with some minor additions such as template and map part menus. There is quite some interest in, and need for, a overhaul. 

- IMO, we will need some different "modes" (viewing, editing, painting on template) with a limit number of icons. Maybe the Microsoft Office apps could be an inspiration.
- In the corners, we could place elements (menus) which need tap-and-hold interaction, so an accidental touch would do nothing. (How often do you close a map?)
[...]

Issue #949, opened 2017-09-04
New command request - zoom in on gps position


ollesmaps:
Please add into the app the possibility to zoom in on current gps position. Zoom around 8,04x (that might be optional). That will make coming back to work faster.

ollesmaps:
It could zoom into defined mapping scale as mentioned in #1005.
If location info is not available it could zoom into place where last editing occured.

lpechacek:
See [pull request 1028](https://github.com/OpenOrienteering/mapper/pull/1028). It does not change the zoom level bot moves the map view so that current position fix is in the centre. After spending about 10 hours in the field so far, testing the new feature, I agree it's a big time saver.

ollesmaps:
Great! I am looking forward to test it.

ollesmaps:
I included the zoom into mapping scale at the same time as fixing map view into current position (in feature request) because I think that is what I would do anyway.  After fixing map view into current position I would like to start immediately mapping and for that I need a zoom into a maping scale. I think that is a workflow - fixing map view into current position, zoom in - for majority of uses. So I combined it in the request. 
For the other cases there are Show whole map, Zoom out buttons.

ollesmaps:
I use this feature a lot. Thank you! 
I still think though that automatical zoom in at the same time would serve well as well :)

mlerjen:
The same here.  Thanks a lot.

Issue #534, opened 2015-11-02
New GUI for the android version


mlerjen:
I would like to propose some changes for GUI the Android version.
![2015-10-29-15-35-59](https://cloud.githubusercontent.com/assets/15611653/10883126/3d3d8ffe-8170-11e5-9ac8-742e0c2aab31.png)

Main issue is to reduce click-paths, number of interruptions and to get a better hierarchy into the different buttons.

1) The main bar consist in five big buttons: The selector, the symbol selector, the drawing tool, the scribble pen and the save button (as long as the autosave does not work properly). After the selection of a new symbol its defined default drawing tool is selected automatically.

2) The sidebar groups the second hierarchy of tools, used regularly.

3) On long click or corner click some buttons roll out a band of buttons of lower hierarchy/rare use, the content of these button bands is contextual (f.eg. no rectangle tool for point features)

4) According to the selection in the main bar, a sub-bar shows the according sub buttons.

5) Template selector and three definable templates which can be directly switched on/off from the main window.

6) Drag tool that does not interrupt the drawing process (returning to the ongoing action after one drag).

7) bottom bar live reporting object size/length and reference to its minimal size/area (green signal), also showing live scale and button to return to a defined mapping scale.

dg0yt:
Other comments for mobile GUI:
 - #491 Mobile UI Toolbar Buttons

dominikzalewski:
I need to mention that issue #545 is a bit correlated with that work/proposal.

mlerjen:
![image](https://user-images.githubusercontent.com/15611653/77778951-17f85000-7052-11ea-847e-49fdc71fc1c2.png)
Point 5 is still very relevant for me. I field-work often with about three templates resp. template combinations I regularly switch in between and about three more I eventually use. Thus switching is something I do all the time. I takes at least six and not seldom eight clicks to temporarily switch from one template to another 
e.g.
click template window
switch template 1 off
switch template 2 on
close template window.
>> Take field note <<  (e.g. localization of  a clearing)
click template window on
switch template 1 on; 
switch template 2 off; 
close template window

I recently followed another pro mapper still working with paper and pen to the field and noted that the time consumed by all this inefficient switching is probably the big minus for digital mapping...

My wish thus would be to have three template presets that can directly switched on with a single click on the main screen. That would somehow include the possibly to assign  a current template combination to a preset somewhere inside the template dialog.

ollesmaps:
[...]

PR #1028, opened 2018-01-02
Mobile UI enhancements

lpechacek:
Two practical enhancements recently tested in the field. I'm willing to update manual pages if they get merged.

For now, mapping (fieldwork) zoom level is fixed at 1:2. While this is an established practice for forest maps, it may not make sense for sprint maps.


lpechacek:
Thanks, Kai, for reviewing the patches and catching the run-away .orig file. I'll push updated patches soon.

Ad "GPS". While searching I could not find a common understandable term for position fix. The most precise would IMO be "satellite navigation position fix" but who would get the meaning. Next one is "(GNSS) position fix" but it sounds much like expert term. Following the Android trace and trying to figure out what terminology Android developers use I pushed the Google button and the first result took me to Android tutorial about [Making Your App Location-Aware](https://developer.android.com/training/location/change-location-settings.html). Sadly the first sentence refers to "GPS". [Next one](https://support.google.com/nexus/answer/6179507?hl=en) also talks about "GPS" in the "Device only" mode. IIRC, the same wording is in my Android 7 Location services setting dialog. Therefore I'm for keeping "GPS" on the UI side until we have a commonly understood term.

On the code side, there is already over 330 occurrences of "gps" string in origin/master and 10 files containing with "gps" in its name. While I'm in favor of adopting a more generic name, switching the naming convention only with this commit would be inconsistent. I'm willing to do the renames when there is consensus about the terms to use.

Ad Icon. I like the idea. The current one is a combination of the Pan tool icon (the four blue arrows) and Mapper's position fix symbol (red arrow in the middle). While the current icon is in line with Mapper's outlook, the location graphical idiom might be more understandable for newcomers.

Ad zoom levels. "2x zoom" and "1x zoom" is aligned with the zoom level info toast display. I'd like to keep the word "zoom" in the menu entry to make it larger and therefore easier to hit with stylus. Plain "2x" and "1x" might be too tiny for field work. I'll change the strings.

Regarding 2:1 magnification, it turned out to be useful in my practice as a overmapping prevention. While my device has about 290 PPI and can render 1:15 000 map with amazing details at first glance, 1x is incomparable with printed map and is not legible enough. 2x zoom level turned out to be useful for getting clearer view of the surveyed area. In my current work 2x represents a good compromise between "legible" and "showing the context". I'd be happy to hear other user's voices. If I'm not mistaken @mlerjen had an opinion about fieldwork map scale.

lpechacek:
related
* issue #949 - New command request - zoom in on gps position
* issue #1005 - Android: Definable "mapping scale" for field work

dg0yt:
I already used a more common location icon for the "Copy coordinates" pie menu action:

https://github.com/OpenOrienteering/mapper/blob/master/images/copy-coords.png

ollesmaps:
Ad "GPS" - just word "Position" might be suitable enough.

dg0yt:
zoom-action-changes


This proposed change is certainly very useful. But how to deal with regarding v0.8.0 (which still needs updates to [Mapper app manual page](http://www.openorienteering.org/mapper-manual/pages/android-app.html) anyway)? I would like to see/do a few modifications.

Major:
- The actions "Record temporary GPS trace", "Record temporary GPS position", and "Clear temporary GPS markers" need to remain grouped together.

Minor:
- Indeed, I prefer the general term "position".
- I would like to change the position-/track-related icons (standard position marker instead of triangle).
- The zoom-out popup menu could include the "whole map" action from the top bar. When removing the "whole map" button from the top bar, the grid button could move to the position above the map parts button.

Anyway, position/sensor related actions remain scattered over three corners of the screen (top left, bottom left, bottom right).

ollesmaps:
I never use actions "Record temporary GPS trace", "Record temporary GPS position", and "Clear temporary GPS markers". This might be because I use a lot the scribbling into bitmap tool.  I wonder if anybody is a heavy user of this?

ollesmaps:
This relates to #952 where "Show whole map - this button shall be rather in the down left corner, where zooming occurs" was also mentioned.

dg0yt:
"Reversed drop" was the word I was looking for :-)

I can't say much about "Record temporary GPS trace", "Record temporary GPS position", and "Clear temporary GPS markers". As I already said at the of my last post, the position features need consolidation, but its out of scope for v0.8.0 / #1028.

Important is to
 - not make the mobile UI even worse.
 - get the UI strings to be translated right, early.

Given that I have this branch checked out the moment I can add a commit with my adjustments for further review.

dg0yt:
Actually, Android now uses the term "location" in its [API](https://developer.android.com/reference/android/location/Location.html). But there is no text anyway in the Android button:

screenshot

mlerjen:
Sorry the question, but where are versions to test?

dg0yt:
There are no downloads which include this change. I'm waiting for feedback from lpechacek. As soon as I can merge the change, I will update the documentation page and start building the v0.8.0 release (candidate).

ollesmaps:
I would argue that making incremantal changes is better than a sudden overhall of UI. Moving buttons within a region is a very small change. It can be tested only in unstable versions. This way we get feedback.

dg0yt:
If this is a minor change, it should be fine to release it (soon) with the next stable build which is v08.0. If it is considered a major change, it won't be in v0.8.0. There is [litte reason](https://github.com/OpenOrienteering/mapper/compare/v0.7.93...master) to provide another preview before v0.8.0.

lpechacek:
Sorry for delayed response - was out sick and kept away from computers in general.

@dg0yt: Good point about icon grouping. I didn't notice it as I made some more changes in my devel branch which shadowed the discrepancy. I also like the term "my location" and the proposal of moving over to "reversed drop" graphical idiom. With that I'm all fine with commit 19454b6.

Regarding the mobile UI as whole, I agree with @ollesmaps that it needs a serious overhaul. While I have made [some more changes](https://github.com/lpechacek/mapper/commits/mobile-ui-gadgets) which I'm happily using in the field, I'm not offering them for merge yet. IMO a big-bang change in direction of mlerjen's proposal in #534 is a better course of action than incremental rework.

dg0yt:
@lpechacek Glad to hear you are back. I will do the merge and also update the documentation page. (I plan to remove the numbers from the screenshot, now that there are icons.)

It may still take a few more days until I can make a release which includes Android and macOS.

Issue #1030, opened 2018-01-05
Mapper prints "runtime error: member access within misaligned address..." upon exit


lpechacek:
On openSUSE Tumbleweed which has Poppler 0.61.1, GDAL 2.2.3 and GCC 7.2.1 Mapper debug builds report misaligned member access upon process termination. This applies to the main Mapper process and system tests.

### Steps to reproduce
1. Build Debug version of Mapper in Qt Creator
2. Run Mapper or undo_manager_t

### Actual behaviour
The program prints the below message upon exit. Automated tests fail.
/usr/include/c++/7/bits/stl_tree.h:772:53: runtime error: member access within misaligned address 0x000000000002 for type 'struct _Rb_tree_node_base', which requires 8 byte alignment
0x000000000002: note: pointer points here
<memory cannot be printed>

Program received signal SIGSEGV, Segmentation fault.

### Expected behaviour
Successful program completion with zero exit status.

### Configuration
Mapper Version: 3a606a3b92af
Operating System: openSUSE Tumbleweed 20180103

### Analysis
This is a result of an ODR violation. Both Mapper and Poppler define `Object` class and its destructor. Poppler has a [global variable aggregating several `Object` instances](https://github.com/freedesktop/poppler/blob/4c7814342806b61fedbb2d45ce74462f9dbc20bc/poppler/StructElement.cc#L250-L266). This global variable gets finalized upon program termination. The finalization code is calling Mapper's `Object::~Object()` for Poppler's class instances resulting in the crash.

Additional data
```
********* Start testing of UndoManagerTest *********
Config: Using QtTest library 5.10.0, Qt 5.10.0 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 7.2.1 20171020 [gcc-7-branch revision 253932])
PASS   : UndoManagerTest::initTestCase()
PASS   : UndoManagerTest::testUndoRedo()
PASS   : UndoManagerTest::cleanupTestCase()
Totals: 3 passed, 0 failed, 0 skipped, 0 blacklisted, 0ms
********* Finished testing of UndoManagerTest *********
/usr/include/c++/7/bits/stl_tree.h:772:53: runtime error: member access within misaligned address 0x000000000002 for type 'struct _Rb_tree_node_base', which requires 8 byte alignment
0x000000000002: note: pointer points here
<memory cannot be printed>

Program received signal SIGSEGV, Segmentation fault.
[...]

PR #1031, opened 2018-01-05
Object: Put class into namespace OpenOrienteering

dg0yt:
The identifier `Object` is very generic. Other software may use the same
identifier for different structures, leading to ODR violations which may
result in any form of undefined behaviour. In particular, poppler (used
by gdal) is known to use this identifier (#1030).

### DIFF
--- a/src/core/autosave.cpp
+++ b/src/core/autosave.cpp
@@ -28,6 +28,9 @@
 
 #include "settings.h"
 
+
+namespace OpenOrienteering {
+
 AutosavePrivate::AutosavePrivate(Autosave& document)
 : document(document)
 , autosave_needed(false)
@@ -135,3 +138,6 @@ bool Autosave::autosaveNeeded() const
 {
 	return autosave_controller->autosaveNeeded();
 }
+
+
+}  // namespace OpenOrienteering
--- a/src/core/map.h
+++ b/src/core/map.h
@@ -51,6 +51,8 @@ class QTranslator;
 class QWidget;
 // IWYU pragma: no_forward_declare QRectF
 
+namespace OpenOrienteering {
+
 class CombinedSymbol;
 class FileFormat;
 class Georeferencing;
@@ -89,7 +91,6 @@ Q_OBJECT
 friend class MapTest;
 friend class MapRenderables;
 friend class OCAD8FileImport;
-friend class XMLFileImport;
 friend class NativeFileImport;
 friend class NativeFileExport;
 friend class XMLFileImporter;
@@ -1560,12 +1561,9 @@ private:
 	static CombinedSymbol* covering_combined_line;
 };
 
-Q_DECLARE_METATYPE(const Map*)
 
 
 // ### Map inline code ###
-
-Q_DECLARE_OPERATORS_FOR_FLAGS(Map::ImportMode)
 
 inline
 int Map::getNumColors() const
@@ -1885,5 +1883,14 @@ TextSymbol* Map::getUndefinedText()
 {
 	return undefined_text;
 }
+
+
+}  // namespace OpenOrienteering
+
+
+Q_DECLARE_METATYPE(const OpenOrienteering::Map*)
+
+Q_DECLARE_OPERATORS_FOR_FLAGS(OpenOrienteering::Map::ImportMode)
+
 
 #endif
--- a/src/core/map_color.cpp
+++ b/src/core/map_color.cpp
@@ -30,8 +30,10 @@
 #include <QLatin1String>
 
 
+namespace OpenOrienteering {
+
 MapColor::MapColor()
-: name(QCoreApplication::translate("Map", "New color")),
+: name(QCoreApplication::translate("OpenOrienteering::Map", "New color")),
   priority(Undefined),
   opacity(1.0f),
   q_color(Qt::black),
@@ -45,7 +47,7 @@ MapColor::MapColor()
 }
 
 MapColor::MapColor(int priority)
-: name(QCoreApplication::translate("Map", "New color")),
+: name(QCoreApplication::translate("OpenOrienteering::Map", "New color")),
   priority(priority),
   opacity(1.0f),
   q_color(Qt::black),
@@ -74,7 +76,7 @@ MapColor::MapColor(int priority)
 			break;
 		case Registration:
 			Q_ASSERT(isBlack());
-			name = QCoreApplication::translate("MapColor", "Registration black (all printed colors)");
+			name = QCoreApplication::translate("OpenOrienteering::MapColor", "Registration black (all printed colors)");
 			break;
 		default:
 			; // no change
@@ -359,3 +361,6 @@ MapColorRgb MapColor::rgbFromSpotColors() const
 	}
 	return rgb;
 }
+
+
+}  // namespace OpenOrienteering
--- a/src/core/map_printer.cpp
+++ b/src/core/map_printer.cpp
@@ -90,6 +90,8 @@ namespace literal
 
 
 
+namespace OpenOrienteering {
+
 // #### MapPrinterPageFormat ###
 
 MapPrinterPageFormat::MapPrinterPageFormat(QSizeF page_rect_size, qreal margin, qreal overlap) 
@@ -341,6 +343,7 @@ void MapPrinterConfig::save(QXmlStreamWriter& xml, const QLatin1String& element_
 #ifdef QT_PRINTSUPPORT_LIB
 
 // ### MapPrinter ###
+
 const QPrinterInfo* MapPrinter::pdfTarget()
 {
 	static QPrinterInfo pdf_target; // TODO: set name and features?
@@ -484,7 +487,7 @@ std::unique_ptr<QPrinter> MapPrinter::makePrinter() const
 		PlatformPrinterProperties::restore(printer.get(), native_data);
 	}
 	
-	printer->setDocName(tr("- Map -"));
+	printer->setDocName(::OpenOrienteering::MapPrinter::tr("- Map -"));
 	printer->setFullPage(true);
 	if (page_format.paper_size == QPrinter::Custom)
 	{
@@ -1235,8 +1238,8 @@ bool MapPrinter::printMap(QPrinter* printer)
 	int step = 0;
 	auto num_steps = v_page_pos.size() * h_page_pos.size();
 	const QString message_template( (options.mode == MapPrinterOptions::Separations) ?
-	  tr("Processing separations of page %1...") :
-	  tr("Processing page %1...") );
+	  ::OpenOrienteering::MapPrinter::tr("Processing separations of page %1...") :
+	  ::OpenOrienteering::MapPrinter::tr("Processing page %1...") );
 	auto message = message_template.arg(1);
 	emit printProgress(0, message);
 	
@@ -1286,16 +1289,16 @@ bool MapPrinter::printMap(QPrinter* printer)
 	
 	if (cancel_print_map)
 	{
-		emit printProgress(100, tr("Canceled"));
+		emit printProgress(100, ::OpenOrienteering::MapPrinter::tr("Canceled"));
 	}
 	else if (!painter.isActive())
 	{
-		emit printProgress(100, tr("Error"));
+		emit printProgress(100, ::OpenOrienteering::MapPrinter::tr("Error"));
 		return false;
 	}
 	else
 	{
-		emit printProgress(100, tr("Finished"));
+		emit printProgress(100, ::OpenOrienteering::MapPrinter::tr("Finished"));
 	}
 	return true;
 }
@@ -1304,5 +1307,8 @@ void MapPrinter::cancelPrintMap()
 {
 	cancel_print_map = true;
 }
+
+
+}  // namespace OpenOrienteering
 
 #endif // QT_PRINTSUPPORT_LIB
--- a/src/core/objects/object.cpp
+++ b/src/core/objects/object.cpp
@@ -64,6 +64,8 @@ namespace literal
 
 
 
+namespace OpenOrienteering {
+
 // ### Object implementation ###
 
 Object::Object(Object::Type type, const Symbol* symbol)
@@ -404,7 +406,7 @@ Object* Object::load(QXmlStreamReader& xml, Map* map, const SymbolDictionary& sy
 	Object::Type object_type = object_element.attribute<Object::Type>(literal::type);
 	Object* object = Object::getObjectForType(object_type);
 	if (!object)
-		throw FileFormatException(ImportExport::tr("Error while loading an object of type %1.").arg(object_type));
+		throw FileFormatException(::OpenOrienteering::ImportExport::tr("Error while loading an object of type %1.").arg(object_type));
 	
 	object->map = map;
 	
@@ -436,7 +438,7 @@ Object* Object::load(QXmlStreamReader& xml, Map* map, const SymbolDictionary& sy
 				break;
 			default:
 				throw FileFormatException(
-				  ImportExport::tr("Unable to find symbol for object at %1:%2.").
+				  ::OpenOrienteering::ImportExport::tr("Unable to find symbol for object at %1:%2.").
 				  arg(xml.lineNumber()).arg(xml.columnNumber()) );
 		}
 	}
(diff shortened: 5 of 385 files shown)


dg0yt:
- Do we want to apply this only to class Object at the moment, or also to related classes?
- Do we need to reconsider our formatting of namespaces?
- How to properly forward declare and use the namespaced classes?

dg0yt:
I integrated the patch in the master and unstable packages on OBS for now. As tested locally, all system tests now pass now also on Fedora Rawhide and Debian Testing.

dg0yt:
Now this is a change to put everything in namespace OpenOrienteering. See the commit comment for details. Is this the road to go?

lpechacek:
Sorry for delayed reply.

>    Do we want to apply this only to class Object at the moment, or also to related classes?

IMO all Mapper code should be enclosed in namespaces to prevent future and not-yet-identified ODR violations.

On second thought, the namespace name should be `OOMapper`, or something else containing 'Mapper', to prevent collisions with other OpenOrienteering tools. It also makes code navigation easier while debugging as it's clearer where the code or variable in question belongs.

```
lpechacek@fmn:~/build-oomapper-Desktop-Debug/test> gdb undo_manager_t
...
(gdb) info functions 
All defined functions:

File /home/lpechacek/build-oomapper-Desktop-Debug/3rd-party/clipper/Clipper-prefix/src/Clipper/cpp/clipper.hpp:
void ClipperLib::Clipper::ReverseSolution(bool);
void ClipperLib::Clipper::StrictlySimple(bool);
void ClipperLib::Clipper::~Clipper();
void ClipperLib::DoublePoint::DoublePoint(double, double);
...
void OpenOrienteering::PointObject::setPosition(int, int);
void OpenOrienteering::PointObject::setRotation(OpenOrienteering::MapCoordF);
void OpenOrienteering::PointObject::setRotation(float);
static bool OpenOrienteering::(anonymous namespace)::isPointOnSegment(OpenOrienteering::MapCoordF const&, OpenOrienteering::MapCoordF const&, OpenOrienteering::MapCoordF const&);

File /home/lpechacek/oomapper/src/core/objects/object.h:
const QRectF &OpenOrienteering::Object::getExtent() const;
OpenOrienteering::Map *OpenOrienteering::Object::getMap() const;
const OpenOrienteering::MapCoordVector &OpenOrienteering::Object::getRawCoordinateVector() const;
const OpenOrienteering::Symbol *OpenOrienteering::Object::getSymbol() const;
OpenOrienteering::Object::Type OpenOrienteering::Object::getType() const;
bool OpenOrienteering::Object::isOutputDirty() const;
...
```

>    Do we need to reconsider our formatting of namespaces?

Sorry, I don't get the question. What formatting do you mean?

>    How to properly forward declare and use the namespaced classes?

Inspired by [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Forward_Declarations), I'm in favor of eliminating forward declarations which serve merely as compile time savers. Those remaining ones should be enclosed in namespace block.

lpechacek:
> Now this is a change to put everything in namespace OpenOrienteering. See the commit comment for details. Is this the road to go?

This has been my plan too. :)

sikmir:
Agree with @lpechacek that namespace name `OOMapper` looks better than `OpenOrienteering` or `OpenOrienteeringMapper`.

src/gui/main_window.h
```diff
@@ -58,7 +60,7 @@ Q_OBJECT
 	/**
 	 * Creates a new main window.
 	 */
-	explicit MainWindow(QWidget* parent = nullptr, Qt::WindowFlags flags = Qt::WindowFlags());
+	explicit MainWindow(QWidget* parent = nullptr, Qt::WindowFlags flags = 0);
```
lpechacek: Unrelated to namespaces.

src/gui/map/map_editor.cpp
```diff
@@ -3944,15 +3948,17 @@ bool MapEditorController::importMapFile(const QString& filename, bool show_error
 			QFile crt_file{ crt_filename };
 			if (!crt_file.open(QFile::ReadOnly))
 			{
-				QMessageBox::warning(window, Map::tr("Error"),
-				                     Map::tr("Cannot open file:\n%1\nfor reading.").arg(crt_filename));
+				QMessageBox::warning(window,
+				                     ::OpenOrienteering::Map::tr("Error"),
+				                     ::OpenOrienteering::Map::tr("Cannot open file:\n%1\nfor reading.").arg(crt_filename));
```
lpechacek: Why the extra qualification on `tr()`?

src/gui/select_crs_dialog.cpp
```diff
@@ -36,15 +36,20 @@
...
 
+namespace OpenOrienteering {
+
 namespace {
 
 enum SpecialCRS {
 	SameAsMap  = 1,
 	Local      = 2,
-	Geographic = 3
+	Geographic = 3
```
lpechacek: Spurious white space.

src/gui/util_gui.cpp
```diff
@@ -75,8 +77,8 @@ QValidator::State DoubleValidator::validate(QString& input, int& pos) const
 
 
 
-namespace Util
-{
+namespace Util {
+
```
lpechacek: Stray white space. Even GH displays it. Pro tip: `git show --color` shows stray white space in red.

src/tools/cut_hole_tool.h
```diff
@@ -35,11 +35,14 @@ class QMouseEvent;
...
 
+namespace OpenOrienteering {
+
+class DrawLineAndAreaTool;
 class MapCoordF;
 class MapEditorController;
 class MapWidget;
 class PathObject;
-class DrawLineAndAreaTool;
+
```
lpechacek: This is another instance of the phenomenon... Why has the empty line been added?

test/locale_t.cpp
```diff
@@ -69,7 +71,7 @@ void LocaleTest::testEsperantoTranslationUtil()
 	TranslationUtil::setBaseName(test_basename);
 	
 	auto test_filename = QString { QLatin1String("some_dir/") + test_basename + QLatin1String("_eo.qm") };
-	QCOMPARE(TranslationUtil::languageFromFilename(test_filename), TranslationUtil::languageFromCode(eo));
+	QVERIFY(TranslationUtil::languageFromFilename(test_filename) == TranslationUtil::languageFromCode(eo));
```
lpechacek: Unrelated change.

lpechacek:
After renaming `OpenOrienteering` namespace to `OOMapper`, the size increase is 64KiB on my system. With `OpenOrienteering` it's 12KiB.

```
-rwxr-xr-x 1 lpechacek users 4108840 Jan 10 11:04 /tmp/Mapper.bin-no-namespaces-3a606a3b92af-stripped
-rwxr-xr-x 1 lpechacek users 4231720 Jan 10 11:04 /tmp/Mapper.bin-w-namespaces-c92de8118912-stripped
-rwxr-xr-x 1 lpechacek users 4174376 Jan 10 11:24 /tmp/Mapper.bin-w-namespaces-OOMapper-stripped
```

rename command:
```
find -type f \( -name \*.h -o -name \*.cpp \) \
 | xargs sed -i 's/namespace OpenOrienteering/namespace OOMapper/;s/OpenOrienteering::/OOMapper::/g'
```

dg0yt:
Many thanks for the feedback and for taking the time to look at it thoroughly. This was quite some work, and some rough edges from the middle of the battles slipped through. I will cleanup, maybe split and re-push.

However, I'm not willing to accept `namespace OOMapper`.
- Outside the orienteering community, OO* is more likely to be understood as OpenOffice*.
- OOMapper is not even an established name in our project. (Well, I always pay attention to who I name the project.)
- We have all rights to assume that `namespace OpenOrienteering` is controlled only by us, so we don't have to fear surprising collisions.
- Not all code in this repository is specific to Mapper: Some code is generic. In fact, we will pobably benefit from using the same namespace across all possible subprojects.
- "OpenOrienteering" is fairly in line with the mentioned C++ guidelines from Goolge, which suggest to use "the name of the project or team" and argue "against abbreviated names". So the only alternativ would be "Mapper", but this too general again IMO.
- Coincidentally, the Mapper translation files are already named OpenOrienteering_*. (However, I'm not too happy about this particalar choice.)

BTW, we use org.openorienteering for Java (Android), and openorienteering.org for web and for macOS preferences.

src/core/autosave.cpp
```diff
@@ -38,10 +41,9 @@ AutosavePrivate::AutosavePrivate(Autosave& document)
 	settingsChanged();
 }
 
-AutosavePrivate::~AutosavePrivate()
-{
-	// nothing, not inlined
-}
+AutosavePrivate::~AutosavePrivate() = default;
```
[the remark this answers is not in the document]
dg0yt: Usually I include such changes when they are not dangerous and when they catch my eye during nearby changes. Since nobody else takes the time to cleanup/beautify/modernize, these changes might never happen otherwise. But I'm okay with separating it out here.

src/core/map.h
```diff
@@ -85,15 +87,16 @@ extern QPointer<QTranslator> map_symbol_translator;
...
+	Q_OBJECT
+	
+	friend class MapRenderables;
+	friend class MapTest;
+	friend class OCAD8FileImport;
+	friend class NativeFileImport;
+	friend class NativeFileExport;
+	friend class XMLFileImporter;
+	friend class XMLFileExporter;
+
```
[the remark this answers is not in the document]
dg0yt: For the record, touching the friends happened in an intermediate step when only core symbols were moved to namespace OpenOrienteering. At this time friends did need full qualification and  forward declaration...

src/core/map_color.cpp
```diff
@@ -130,7 +132,7 @@ bool operator==(const SpotColorComponents& lhs, const SpotColorComponents& rhs)
 	return lhs.size() == rhs.size()
 	       && std::is_permutation(begin(lhs), end(lhs), begin(rhs), [](const auto& left, const auto& right) {
 		return *left.spot_color == *right.spot_color
-		       && qAbs(left.factor - right.factor) < 1e-03;
+		       && qAbs(left.factor - right.factor) < 1e-03f;
```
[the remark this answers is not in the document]
dg0yt: FP is unrelated, indeed. I'm just trying to get rid of yellow markers when it seems appropriate, so that the real issues become more visible.

src/core/map_printer.cpp
```diff
@@ -1235,8 +1237,8 @@ bool MapPrinter::printMap(QPrinter* printer)
 	int step = 0;
 	auto num_steps = v_page_pos.size() * h_page_pos.size();
 	const QString message_template( (options.mode == MapPrinterOptions::Separations) ?
-	  tr("Processing separations of page %1...") :
-	  tr("Processing page %1...") );
+	  ::OpenOrienteering::MapPrinter::tr("Processing separations of page %1...") :
+	  ::OpenOrienteering::MapPrinter::tr("Processing page %1...") );
```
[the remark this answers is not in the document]
dg0yt: That's part of "exposes Mapper to all namespace-related issues of lupdate." Without this, `lupdate` would put the translation in the wrong context. At the moment I'm not much inclined to fix issues in `lupdate`...

src/core/objects/object.cpp
```diff
@@ -64,6 +64,27 @@ namespace literal
...
 
 
+namespace OpenOrienteering {
+
+namespace {
+	
+/** 
+ * Checks if a point is on the segment defined by the given start and end coordinates.
+ */
+bool isPointOnSegment(const MapCoordF& seg_start, const MapCoordF& seg_end, const MapCoordF& point)
```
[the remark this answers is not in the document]
dg0yt: All unnamed namespaces are used intentionally, to avoid external linkage when it is not  neccessary. In particular, this function was moved away from a location were it was not used, and the anonymous namespace helped to reconfirm that there is no other usage.

src/gui/map/map_editor.cpp
dg0yt: As above: `lupdate`

src/gui/util_gui.cpp
dg0yt: I used `gitk` in this case. I normally use `git`, too, but this change was too long, and GH wasn't available before the push...

test/locale_t.cpp
dg0yt: Actually it is related: `QCOMPARE` was no longer able to find a comparison function. Just another case of the issues with operator lookup which consumed a lot of time when preparing this change. There might be a better fix, though.

dg0yt:
> > Do we need to reconsider our formatting of namespaces?

> Sorry, I don't get the question. What formatting do you mean?

Actually I meant something like this: 
https://google.github.io/styleguide/cppguide.html#Namespace_Formatting


> > How to properly forward declare and use the namespaced classes?

> Inspired by Google Style Guide, I'm in favor of eliminating forward declarations which serve merely as compile time savers. Those remaining ones should be enclosed in namespace block.

Well, I would pay the price in terms of compile time. However, the Google state it a little bit different: "Try to avoid forward declarations of entities defined in another project." i.e. within your own project, you may use forward declarations (because you control changes to the types).

In addition, many Mapper classes include implementation details (such as `QLineEdit*` members) which you would hide in a library designed for external use. That's why Mapper's headers use many types, and including everything becomes much more expensive than forward declarations.

test/locale_t.cpp
dg0yt: Solved by moving `operator==` (at the top of the file) to `namespace OpenOrienteering`.

src/core/map_printer.cpp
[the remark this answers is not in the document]
lpechacek: Thanks for the explanation. Then the `lupdate` limitation should be documented either in the commit message or a comment so that future Mapper developers can decide when to remove the workaround.

src/core/objects/object.cpp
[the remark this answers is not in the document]
lpechacek: Understood. My point was that I'd like to see a brief note about the move in commit message so that it's clear that the move was indeed intentional. Again, this serves as a hint for future contributors.

src/gui/util_gui.cpp
lpechacek: Np. This remark is merely a result of my text-terminal-centric usage pattern and influence of the Linux kernel coding convention.

lpechacek:
>>> Do we need to reconsider our formatting of namespaces?

>> Sorry, I don't get the question. What formatting do you mean?

>  Actually I meant something like this: https://google.github.io/styleguide/cppguide.html#Namespace_Formatting 

Yes, I like the style and I can see it was applied in the proposed patch already. Let's keep it.

>>>  How to properly forward declare and use the namespaced classes?   

>> Inspired by Google Style Guide, I'm in favor of eliminating forward  declarations which serve merely as compile time savers. Those remaining  ones should be enclosed in namespace block. 

> Well, I would pay the price in terms of compile time. However, the  Google state it a little bit different: "Try to avoid forward  declarations of entities defined in another project." i.e. within your  own project, you may use forward declarations (because you control  changes to the types).

Reading the section again, I can see the language is a bit contradictory in its language. While it start with imperative, the decision section uses softer tone. Obviously I chose a wrong reference. Sorry.

BTW it was the compilation unit dependency handling point which caught my attention in GSG. ON second thought I don't see how the dependency handling could break with sane use of forward declarations.

> In addition, many Mapper classes include implementation details (such as QLineEdit*  members) which you would hide in a library designed for external use.  That's why Mapper's headers use many types, and #including everything  becomes much more expensive than forward declarations.

Back to your original question. For the moment I have just one point in mind:
* class forward declarations will be enclosed in namespace block
```
namespace OpenOrienteering {
class B;
class C;
...
} // OpenOrienteering
```

Another thing which creeps into my mind is "class forward declarations are allowed only in header files". I don't see a reason for forward declaration in code files except for cyclic dependencies. HTH

lpechacek:
> However, I'm not willing to accept `namespace OOMapper`.
> ...

I more or less agree with all your points. Especially the OpenOffice association came to my mind when writing the proposal. Yet, `OOMapper` is a lot easier to read in compiled code (think of debugging outside of IDE) than the generic identifier.

dg0yt:
Source code changes trimmed, translation file updates moved to separate commit. Commit messages updated.

I'm not happy with putting the QFlags operators in the global namespace (cf. https://stackoverflow.com/questions/10755058/qflags-enum-type-conversion-fails-all-of-a-sudden). Fixing that requires either:
- Say `using Qt::operator|` in the files that need it, or
- Move our QFlags and their operators from class scope to namespace scope (where they are found by ADL), maybe in `::OpenOrienteering::Mapper`.

But this is stuff for another change and not urgent.

dg0yt:
(And of course, I will refresh the translations commit with the latest updates from Weblate before merging.)

test/locale_t.cpp
```diff
@@ -70,6 +78,7 @@ void LocaleTest::testEsperantoTranslationUtil()
 	
 	auto test_filename = QString { QLatin1String("some_dir/") + test_basename + QLatin1String("_eo.qm") };
 	QCOMPARE(TranslationUtil::languageFromFilename(test_filename), TranslationUtil::languageFromCode(eo));
+	//QVERIFY(TranslationUtil::languageFromFilename(test_filename) == TranslationUtil::languageFromCode(eo));
```
lpechacek: Comment added intentionally?

PR #1034, opened 2018-01-15
OCD/Mapper coordinate reference system mapping

lpechacek:
In Mapper, the map spatial reference system of choice can be expressed in multiple ways. Either as plain UTM zone description, Gauss–Krüger zone, EPSG code or proj.4 description string.  The other popular map drawing software uses its own way of reference system identification. Combinations of proprietary combination of grid and zone id's point to EPSG codes.

This patch series introduces a subsystem for bi-directional translation between these two systems. On both sides there are ambiguities. Shall OCD UTM zone 33N be translated to Mapper UTM 33N or EPSG 32633? Shall Mapper EPSG 5514 (S-JTSK, covering former Czechoslovakia) become OCD Czech / S-JTSK or Slovak / S-JTSK? For now these ambiguities get resolved by taking the first match, keeping original code behavior.

First two patches in the series extract CRS translation code into standalone unit and provide unit test for the code piece.

Third patch moves OCD string parsing back to its original place and changes the interface. Fourth patch makes the translation code more extensible by establishing a central translation table for EPSG codes and handling UTM and G-K in code. Internally is the translation working with `<CRS_id, CRS_parameter>` pairs. Rewrite into the `Georeferencing` class notation happens in `applyGridAndZone()`.

The fifth patch add reverse translation from Mapper to OCD codes, which will be needed for completing pull request #1032. The last patch in the series adds synthetic test cases to complete test coverage.


lpechacek:
Commit 30ab9da is the final piece that completes my idea of georeferencing translation between the two programs. There are some rough edges to peel but the core is working.

Note to myself - failed translation must be handled gracefully. Currently unitialized values leak out into the file. The behavior of `r` parameter also deserves exploration. We should perhaps use `r==0` for all Mapper "local" projections.

src/fileformats/ocd_georef.cpp
```diff
@@ -0,0 +1,409 @@
...
+		}
+		break;
+
+	case OCD_GRID_GERMANY:
+		// DHDN, Gauss-Krüger, Zone 2-5
+		if (ocd_zone_id >= 2 && ocd_zone_id <= 5)
+		{
+			retval.push_back({ MapperCrs::GAUSS_KRUEGER, ocd_zone_id });
+		}
+		[[ fallthrough ]]; // get the EPSG code
```
dg0yt: No `[[ fallthrough ]]` with older GCC (Android). I didn't even dare to use Q_FALLTHROUGH, but this is something which we could add to our backports.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+	                             const int combined_ocd_grid_zone,
+	                             const std::function<void(const QString&)>& warning_handler);
+
+	/**
+	 * @brief Plain translation from OCD grid/zone combination
+	 *        into possible Mapper schemes.
+	 *
+	 * @see translateMapperOcd
+	 */
+	static auto translateOcdMapper(const int ocd_grid_id, const int ocd_zone_id);
```
dg0yt: `auto` and  `translateOcdMapper` don't make it clear to me what this function is doing. (Expecting "translate...to...")

Unless it is meant to be public, consider removing it from the header. Unless it is meant to be called from the separate test, the function may even live in an anonymous namespace in the cpp file.
(Actually this may apply to other members of OcdGeoref as well.)

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * @brief translateOcdMapper counterpart. Provide possible OCD grid/zone
+	 *        combinations for a Mapper georeferencing.
+	 *
+	 * @see translateOcdMapper
+	 */
+	static auto translateMapperOcd(const MapperCrs crs_unique_id, const int crs_param);
```
dg0yt: Similar to `translateOcdMapper`.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with OpenOrienteering.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef OCD_GEOREF_H
+#define OCD_GEOREF_H
```
dg0yt: OPENORIENTEERING_OCD_GEOREF_H

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+	};
+
+	/// Imports string 1039 field i.
+	static void applyGridAndZone(Georeferencing& georef,
+	                             const int combined_ocd_grid_zone,
+	                             const std::function<void(const QString&)>& warning_handler);
+
+	/**
+	 * @brief Plain translation from OCD grid/zone combination
+	 *        into possible Mapper schemes.
```
dg0yt: BTW we don't need `@brief` because doxygen runs with autobrief activated.

lpechacek:
Thanks for the review comments, Kai. I'm going to incorporate your feedback into the patch set, and **rebase** it atop current dev branch. If you disagree with the rebase, please speak up.

dg0yt:
I'm fine with rebasing. It just starts to get complicated when you merge my WIP ocd export branch. Stop with this PR before the merge, and create a separate PR against my WIP branch for integration. Then it will be easier to review this code again, and to do the actual integration once I cleaned up my WIP branch, including unpublished changes.

lpechacek:
Sounds good.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+		// corresponds to the code and its value is the field value.
+		double a;
+		int m, x, y, i;
+
+		OcdStringFields() :
+		    a(std::nan("OCD")),
+		    m(-1),
+		    x(std::numeric_limits<int>::min()),
+		    y(std::numeric_limits<int>::min()),
+		    i(-1)
```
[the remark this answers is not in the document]
lpechacek: ACK

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+
+class OcdGeoref
+{
+	Q_DECLARE_TR_FUNCTIONS(OpenOrienteering::OcdGeoref)
+
+public:
+	struct OcdStringFields
+	{
+		// Named after OCD string fields. Structure element name
+		// corresponds to the code and its value is the field value.
```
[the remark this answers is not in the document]
lpechacek: Done.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * @brief Fill in provided georef with data from type 1039 string.
+	 * @param georef Reference to a Georeferencing object.
+	 * @param param_struct Data extracted from OCD type 1039 string.
+	 * @param warning_handler Function to handle import warnings.
+	 */
+	static void georefFromFields(Georeferencing& georef,
+	                             const OcdStringFields& param_struct,
+	                             const std::function<void (const QString&)>& warning_handler);
```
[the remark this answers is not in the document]
lpechacek: I recall that we discussed the traits of `Georeferencing` earlier but cannot find the reference. Anyway, my original proposal was passing `Georeferencing` by value but it is wrong because it inherits from `QObject`. We agreed that `Georeferencing` should not serve two purposes and should be freed from its `QObject` duties.

While writing this, I realized that for now the function can be called `setupGeorefFromFields()`. We can return to the original intent of passing by value after `Georeferencing` is fixed.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * @brief fieldsFromGeoref
+	 * @param georef Source Georeferencing.
+	 * @param warning_handler Function to handle conversion warnings.
+	 * @return Data for OCD type 1039 string compilation.
+	 */
+	static OcdStringFields fieldsFromGeoref(const Georeferencing& georef,
+	                                           const std::function<void (const QString&)>& warning_handler);
```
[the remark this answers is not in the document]
lpechacek: The spelling would be `OcdGeoref::fieldsFromGeoref()`. The pair `OcdGeoref::setupGeorefFromFields()` & `OcdGeoref::fromGeoref()` feels weird to me. Alternatives are:
* `OcdGeoref::georefFromFields()` & `OcdGeoref::fieldsFromGeoref()` (status quo)
* `OcdGeoref::setupGeorefFromFields()` & `OcdGeoref::fieldsFromGeoref()`

Only these two (complementary) functions are now visible to the outside.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+
+protected:
+	enum struct MapperCrs;
+	static const std::map<const MapperCrs, const QString> crs_string_map;
+	static const std::map<const QString, const MapperCrs> crs_id_map;
```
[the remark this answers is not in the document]
lpechacek: This part is a result of an attempt to stop `Georeferencing` interface design anti-pattern leaking into `OcdGeoref`. That is why I focused on simplicity rather than efficiency here.
But as `Georeferencing` rework is not on the roadmap, I'll rework this piece of code towards greater efficiency.

[side note - Early `OcdGeoref` implementations worked with strings in the translation table. Seeing the disaster I introduced scoped enums for CRS identification inside `OcdGeoref` and isolated string handling into `fieldsFromGeoref()` and `applyGridAndZone()` with the outlook that `Georeferencing` will provide a more programmer-friendly interface one day.]

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+#include <cmath>
+#include <limits>
+
+#include <QString>
+
+#include "core/georeferencing.h"
+
+namespace OpenOrienteering {
+
+class OcdGeoref
```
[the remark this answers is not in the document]
lpechacek: Absolutely. Class designation was a remnant of the early rewrite attempts. Given that `OcdGeoref` is stateless, everything is now packed in name spaces.

src/fileformats/ocd_georef.cpp
lpechacek: Good point. Converted `[[ fallthrough ]];` to `// fallthrough` so that it can be converted back later on.

src/fileformats/ocd_georef.h
lpechacek: Good catch, this was a blind spot. Functions have been renamed and return types are properly identified.

src/fileformats/ocd_georef.h
lpechacek: Fixed.

src/fileformats/ocd_georef.h
lpechacek: Fixed.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,115 @@
...
+	                                           const std::function<void (const QString&)>& warning_handler);
+
+protected:
+	enum struct MapperCrs;
+	static const std::map<const MapperCrs, const QString> crs_string_map;
+	static const std::map<const QString, const MapperCrs> crs_id_map;
+
+	struct MapperCrsRecord
+	{
+		const MapperCrs crs_id; // CRS id, same strings as CRSTemplateRegistry::find() consumes
```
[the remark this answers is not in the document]
lpechacek: Fixed. API documentation looks good now.

lpechacek:
Major changes in this revision - functions are enclosed in name spaces, enums are scoped, `std::map` is gone. Other changes requested in the above comments are included too.

lpechacek:
And two micro-optimizations. Capture by reference in `std::find_if` lambdas and stop searching for OCD on first match. I'm done for the moment.

test/georef_ocd_mapping_t.cpp
```diff
@@ -0,0 +1,318 @@
...
+Q_DECLARE_METATYPE(OcdGeoref::OcdStringFields)
+
+bool operator==(const OcdGeoref::OcdStringFields& lhs, const OcdGeoref::OcdStringFields& rhs)
+{
+	return lhs.i == rhs.i
+	        && lhs.m == rhs.m
+	        && lhs.x == rhs.x
+	        && lhs.y == rhs.y
+	        && ((qIsNaN(lhs.a) && qIsNaN(rhs.a)) || qAbs(lhs.a - rhs.a) < 5e-9); // 8-digit precision or both NaN's
+}
```
dg0yt: How about moving this function to ocd_georef.h/.cpp? It contains assumptions about OcdGeoref...Fields (precision) which belong there. And it would stop someone else from adding it there.

test/georef_ocd_mapping_t.h
```diff
@@ -0,0 +1,39 @@
...
+
+	void initTestCase();
+
+	void testOcdToMapper_data();
+	void testOcdToMapper();
+	void testMapperToOcd_data();
+	void testMapperToOcd();
+};
+
+#endif // OPENORIENTEERING_GEOREF_OCD_MAPPING_TEST_H
```
dg0yt: In the newer tests, I sticked with just the .cpp file, as there is little benefit from the header. I have no strong guideline for this, though.

src/fileformats/ocd_file_import.cpp
```diff
@@ -284,28 +288,38 @@ void OcdFileImport::importGeoreferencing(const QString& param_string, int /*ocd_
...
+				int_param_ptr = &fields.x;
+			// fallthrough
+		case 'y':
+			if (!int_param_ptr)
+				int_param_ptr = &fields.y;
+			// fallthrough
+		case 'i':
+			if (!int_param_ptr)
+				int_param_ptr = &fields.i;
+			// common string parsing code
```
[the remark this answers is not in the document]
lpechacek: OK. Does `tryLoad()` indeed needs to be a `OcdFileImport` _method_? Static function or anonymous namespace look like a more appropriate choice for this tiny code bit.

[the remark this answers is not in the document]
dg0yt: I had an anonymous namespace in my comment but removed it for brevity before posting ;-)

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,72 @@
...
+ * handling. Structure element name corresponds to field code and its value
+ * is the field value.
+ */
+struct OcdStringFields
+{
+	double a {NAN};
+	int m {-1},
+		x {std::numeric_limits<int>::min()},
+		y {std::numeric_limits<int>::min()},
+		i {std::numeric_limits<int>::min()};
```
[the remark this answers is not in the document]
lpechacek: Run away formatting... will fix and add comments.

Given than `Georeferencing`'s `projected_ref_point` is initialized to (0, 0), setting x and y to zero here is likely neutral in terms of program behavior. We are, however, giving up on distinguishing between uninitialized value and zero from OCD file. IMHO when there is no input (no x, y in file), there should be no action (no `setProjectedRefPoint()` call in `setupGeorefFromFields()`). Anyway, I'll carry out the change.

src/fileformats/ocd_georef.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Translate from Mapper CRS representation into OCD one.
+ * @param georef Source Georeferencing.
+ * @param warning_handler Function to handle conversion warnings.
+ * @return Data for OCD type 1039 string compilation.
+ */
+OcdStringFields fieldsFromGeoref(const Georeferencing& georef,
+				 const std::function<void (const QString&)>& warning_handler);
+
+}  // namespace OcdGeoref
```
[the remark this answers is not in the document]
lpechacek: Hmmm?! That's a logical next step in the evolution from embedded Type 1039 handling towards independent translation unit and I like the idea. I'll implement this as a new patch on top of the existing series.

test/georef_ocd_mapping_t.cpp
```diff
@@ -0,0 +1,318 @@
...
+					fields.a = double_param;
+			}
+			break;
+		default:
+			; // ignore unknown and unhandled fields
+		}
+	}
+
+	return fields;
+}
```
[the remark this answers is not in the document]
lpechacek: The test was created very early in the process and reflects the old string based interface and I never returned back to rethinking the test structure.

Now, with the `OcdStringFields`-based in place, I'd rather migrate the test to using the new interface and leave raw strings behind. Unless you see a serious flaw in the plan, I'll carry it out.

test/georef_ocd_mapping_t.cpp
lpechacek: Makes a lot of sense to me. The operator is generic to the type.

test/georef_ocd_mapping_t.h
lpechacek: Sounds good. I'm happy to adopt more compact form for the test.

lpechacek:
Changes in this revision:
- Removal of `OcdGeoref` namespace
- `setupGeorefFromFields` and `fieldsFromGeoref` are now members of `OcdGeorefFields` and their names have been shortened
- Move of non-member `operator==` into `OcdGeorefFields`
- Changed string parsing in `src/fileformats/ocd_file_import.cpp`
- Test case compacted and uses pre-parsed data instead of raw strings
- Trivial fixes in UTM zone number checks
- `OcdGeorefFields` data member documentation
- (formatting) Excess brace removal and whitespace cleanup

src/fileformats/ocd_georef_fields.h
```diff
@@ -52,30 +50,26 @@ struct OcdStringFields
...
-/**
- * Fill in provided georef with data extracted from type 1039 string.
- * @param georef Reference to a Georeferencing object.
- * @param param_struct Data extracted from OCD type 1039 string.
- * @param warning_handler Function to handle import warnings.
- */
-void setupGeorefFromFields(Georeferencing& georef,
-                           const OcdStringFields& param_struct,
-                           const std::function<void (const QString&)>& warning_handler);
+	bool operator==(const OcdGeorefFields& rhs) const;
```
dg0yt: Please make `operator==` a non-member function. Cf. [CppCoreGuidelines C.68](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c86-make--symmetric-with-respect-to-operand-types-and-noexcept). 
While at it, just add `operator!=` (in terms `operator==`).

This the final remark. Thanks again.

lpechacek: Done.

lpechacek:
Changes in this revision:
* Undone move of non-member `operator==` into `OcdGeorefFields`
* Added `operator!=(const OcdGeorefFields&, const OcdGeorefFields&)`

lpechacek:
Changes in this revision:
*  Undone move of non-member `operator==` into `OcdGeorefFields`
*  Added `operator!=(const OcdGeorefFields&, const OcdGeorefFields&)`

src/fileformats/ocd_georef_fields.h
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * OCD type 1039 string field values packed in a struct for convenient
+ * handling. Structure element name corresponds to field code and its value
+ * is the field value.
+ */
+struct OcdGeorefFields
+{
+	static constexpr int UNINIT_INT { std::numeric_limits<int>::min() }; ///< Signifies unitialized field value
+	static constexpr double UNINIT_DOUBLE { NAN }; ///< Signifies unitialized field value
```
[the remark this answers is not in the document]
lpechacek: The choice of NaN and INT_MIN is exactly for the purpose of identifying uninitialized values. Setting "reasonable defaults" may, however, simplify the code. I'll look into it.

src/fileformats/ocd_georef_fields.h
```diff
@@ -0,0 +1,85 @@
...
+bool operator==(const OcdGeorefFields& lhs, const OcdGeorefFields& rhs);
+
+/**
+ * Trivial non-equal-to operator.
+ * @see operator==(const OcdGeorefFields&, const OcdGeorefFields&)
+ */
+inline
+bool operator!=(const OcdGeorefFields& lhs, const OcdGeorefFields& rhs)
+{
+    return !operator==(lhs, rhs);
```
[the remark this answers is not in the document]
lpechacek: Hell!  I thought I fixed all the whitespace errors. :)

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,454 @@
...
+#include "fileformats/ocd_file_import.h"
+#include "fileformats/ocd_file_export.h"
+
+namespace OpenOrienteering {
+
+namespace {
+
+enum struct MapperCrs
+{
+	LOCAL, UTM, GAUSS_KRUEGER, EPSG,
```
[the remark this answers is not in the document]
lpechacek: ok

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,454 @@
...
+		switch (crs_id)
+		{
+		case MapperCrs::UTM:
+			if (param_string.endsWith(QLatin1String(" S")))
+			{
+				param_string.remove(QLatin1String(" S"));
+				param_string.prepend(QLatin1String("-"));
+			}
+			else
+				param_string.remove(QLatin1String(" N"));
```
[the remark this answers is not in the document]
lpechacek: ok

lpechacek:
Changes in this revision:
* Whitespace cleaned up, style change requests addressed
* `OcdGeorefFields` is initialized with safe defaults
* Field `r` is now handled
* Rebased on top of current `origin/dev`

src/fileformats/ocd_georef_fields.h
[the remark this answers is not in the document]
lpechacek: I dropped the "poison" values in favor of `r` field handling. Paper coordinates vs local coordinates case may need some fine tuning but as far as I can tell that is not critical at the moment. The good thing is that real world coordinate systems are handled well.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,461 @@
...
+	if (result[0].crs_id == MapperCrs::LOCAL)
+		return;
+
+	// look up CRS template by its unique name
+	auto entry_it = std::find_if(std::begin(crs_string_map),
+	                             std::end(crs_string_map),
+	                             [&result](auto e) { return e.key == result[0].crs_id; });
+	if (entry_it == std::end(crs_string_map))
+	{
+		    Q_ASSERT(false);
```
dg0yt: `Q_ASSERT(false)` looks like abuse of an assert statement, and in fact it is.
I switched to using `qDebug("Useful message")` instead. It does not abort where the release build wouldn't, it prints a useful message, and it is eliminated in release builds when `QT_NO_DEBUG_OUTPUT` is defined.
(We might even use categorized logging since Qt 5.3, but there are no established logging categories im Mapper.)

For this particular function, an alternative might be a [function-try-block](http://en.cppreference.com/w/cpp/language/function-try-block): Throw an error whenever there is one, and call the warning handler in the `catch` block.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,461 @@
...
+};
+
+/**
+ * Plain translation from OCD grid/zone combination into possible Mapper
+ * schemes.
+ *
+ * @see translateMapperToOcd
+ */
+std::vector<MapperCrsRecord> translateOcdToMapper(const int ocd_grid_id,
+                                                  const int ocd_zone_id)
```
dg0yt: The API is prepared to handle different ways of expressing the CRS in Mapper. But are we really going to use this in foreseeable time? And if yes, couldn't there be a separate function? My impression is that just returning `MapperCrsRecord` would be a fair simplification, here and in `applyGridAndZone`.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,461 @@
...
+	auto crs_id_string = georef.getProjectedCRSId();
+	auto crs_param_strings = georef.getProjectedCRSParameters();
+	auto report_warning = [&]() {
+		QStringList params;
+		std::for_each(begin(crs_param_strings), end(crs_param_strings),
+		              [&](const QString& s) { params.append(s); });
+		warning_handler(::OpenOrienteering::OcdFileExport::tr(
+		                    "Could not translate coordinate reference system '%1:%2'.")
+		                .arg(crs_id_string)
+		                .arg(params.join(QLatin1Char(','))));
```
dg0yt: clazy suggests that `arg(crs_id_string, params.join(QLatin1Char(',')))` would be more efficient than calling `args` twice for arguments of type `QString`.

dg0yt:
PS: I intended my review comments to be ordered by location (like seen when reviewing) rather by than creation time (like seen now, after finishing the review).

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+constexpr struct {
+	const char* key;
+	MapperCrs value;
+} crs_id_map[] {
+    { "Local", MapperCrs::LOCAL },
+    { "UTM", MapperCrs::UTM },
+    { "Gauss-Krueger, datum: Potsdam", MapperCrs::GAUSS_KRUEGER },
+    { "EPSG", MapperCrs::EPSG },
+};
```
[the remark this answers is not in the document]
lpechacek: Good point. It evolved from two `std::map`s but now one table suffices.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+enum struct OcdGrid {
```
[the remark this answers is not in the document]
lpechacek: Mind adding the no-all-caps rule to Mapper's Coding Style, along with examples? That will help me follow the rules more easily.

This is going against my habits. All caps is in my world an idiom for a constant. Doing that otherwise in OpenOrienteering requires a bit of effort from me.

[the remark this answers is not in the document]
dg0yt: I see. But it is coding style in Qt, and it is even in the C++ Core guidelines. 

Note that you are labeled 'member', so you can add missing pieces to the Mapper wiki ;-).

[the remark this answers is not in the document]
lpechacek: Just had a look at Qt style enums and can see it. I'll use that style.

src/fileformats/ocd_georef_fields.cpp
lpechacek: One of 508 clazy warnings issued for this branch build. I'd say the benefits of optimizing an error path are shadowed by the effort required for the patch set rework. On the positive side, I tried out clazy and learned about existence of multi-arg `QString::arg()`.

src/fileformats/ocd_georef_fields.cpp
lpechacek: Hmmm, `Q_ASSERT()` does not do what I thought it does. Unlike `assert(3)` it does not terminate the program. I'll turn this into `qCritical()` as these two occurences of `Q_ASSERT()` are triggered in case of internal software inconsistency.

dg0yt: Even C `assert` won't do anything if NDEBUG is defined (what we do in release builds):

>     If  the  macro  NDEBUG  was  defined  at the moment <assert.h> was last
>     included, the macro assert() generates no code, and hence does  nothing
>     at all.  Otherwise, the macro assert() prints an error message to stan‐
>     ...

src/fileformats/ocd_georef_fields.cpp
dg0yt: I run with `CLAZY_CHECKS` set to `level0,level1,no-rule-of-two-soft,no-const-signal-or-slot` in my Qt Creator build environment. I didn't try hard enough to put it in `code-check-wrapper.sh` or cmake settings.

lpechacek:
Changes in this revision:

* ALL_CAPS removed
* extra mapping table removed
* `Q_ASSERT()`s turned into warnings
* dropped `std::vector` usage in internal interfaces
* use of multi-arg `QString::arg()`
* includes cleaned up

lpechacek:
The patch series does not bring any new builds failures on Android. The existing ones have been reported. I'm done for the moment.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,457 @@
...
+};
+
+/**
+ * Plain translation from OCD grid/zone combination into a Mapper
+ * scheme.
+ *
+ * @see translateMapperToOcd
+ */
+MapperCrsRecord translateOcdToMapper(const int ocd_grid_id,
+                                     const int ocd_zone_id)
```
dg0yt: Instead of a free function, I would prefer a "named constructor" i.e.  `MapperCrsRecord::fromOcdGridAndZone`. This is more expressive, for example with auto style:
~~~
auto rec = MapperCrsRecord::fromOcdGridAndZone(grid_id, zone_id);
~~~
...GridAndZone helps to get the arguments in the right order. However if it would take a single combined grid and zone  parameter, it might be just `MapperCrsRecord::fromOcd`. And...

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,457 @@
...
+{
+	// while OCD codes can be negative, we are keeping grid_id always positive
+	// to match the enums; signum is carried in zone_id
+	const int ocd_grid_id = qAbs(combined_ocd_grid_zone / 1000),
+	          ocd_zone_id = qAbs(combined_ocd_grid_zone % 1000)
+	                        * (combined_ocd_grid_zone < 0 ? -1 : 1);
+
+	// Mapper has multiple ways of expressing some of the CRS's
+	// for now we are working with the first match
+	auto result = translateOcdToMapper(ocd_grid_id, ocd_zone_id);
```
dg0yt: ... this is the only time the split values are used.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,457 @@
...
+	        && lhs.x == rhs.x
+	        && lhs.y == rhs.y
+	        && ((qIsNaN(lhs.a) && qIsNaN(rhs.a)) || qAbs(lhs.a - rhs.a) < 5e-9) // 8-digit precision or both NaN's
+	        && lhs.r == rhs.r;
+}
+
+void OcdGeorefFields::setupGeoref(Georeferencing& georef,
+                                  const std::function<void (const QString&)>& warning_handler) const
+{
+	if (m >= 0)
```
dg0yt: `m >= 1`

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,457 @@
...
+		applyGridAndZone(georef, i, warning_handler);
+
+	QPointF proj_ref_point(x, y);
+	georef.setProjectedRefPoint(proj_ref_point, false);
+}
+
+OcdGeorefFields OcdGeorefFields::fromGeoref(const Georeferencing& georef,
+                                            const std::function<void (const QString&)>& warning_handler)
+{
+	OcdGeorefFields retval;
```
dg0yt: a, m, x, y need to be set here: They can be returned and are useful even if everything else fails.

src/fileformats/ocd_georef_fields.cpp
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	// similar to Mapper, OCD has multiple codes for one CRS
+	// for now we work with the first match
+	retval.a = georef.getGrivation();
+	retval.i = static_cast<int>(result.ocd_grid_id) * 1000 + qAbs(result.ocd_zone_id);
+	// propagate signum which we carry in ocad_zone_id
+	retval.i *= result.ocd_zone_id < 0 ? -1 : 1;
+	retval.m = georef.getScaleDenominator();
+	const QPointF offset(georef.getProjectedRefPoint());
```
dg0yt: You need the projected coordinates at 0,0. The existing code uses `georef.toProjectedCoords(MapCoord{})`.

test/georef_ocd_mapping_t.cpp
```diff
@@ -0,0 +1,261 @@
...
+		QTest::newRow("Mapper CRS Local")
+		        << OcdGeorefFields { 21.32000000, 15000, 0, 0, 1000, 1 }
+		        << GeorefCoreData { 15000, 21.32, 21.32, { 58.972, -34.412 }, QStringLiteral("Local"), { } };
+		QTest::newRow("Mapper UTM 41N expressed as EPSG code")
+		        << OcdGeorefFields { 2.30000000, 10000, 285016, 5542944, 2041, 1 }
+		        << GeorefCoreData { 10000, 2.3, 2.3, { 50, 60 }, QStringLiteral("EPSG"), { QStringLiteral("32641") } };
+		QTest::newRow("Mapper UTM 21S expressed as EPSG code")
+		        << OcdGeorefFields { -2.70000000, 15000, 285016, 4457056, -2021, 1 }
+		        << GeorefCoreData { 15000, -2.70, -2.70, { -50, -60 }, QStringLiteral("EPSG"), { QStringLiteral("32721") } };
+		// exercise error paths
```
dg0yt: Even in case of error, a, m, x, y need to be properly set. In addition, the test needs to verify that a warning was added (probably another test data field).

For verifying x and y, you must test with projected coordinates instead of geograpic coordinates. I think projected coordinates are more appropriate anyway: This is what is saved in the OCD file. And the test would no longer rely on proj transformations.

test/georef_ocd_mapping_t.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+#include <algorithm>
+#include <vector>
+
+#include <QByteArray>
+#include <QLatin1Char>
+#include <QList>
+#include <QMetaType>
+#include <QObject>
+#include <QStaticStringData>
```
dg0yt: Private implementation detail of QString. Do not include. (May need an update of `iwyu-qt.imp`.)

test/georef_ocd_mapping_t.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include <QString>
+#include <QStringList>
+#include <QStringLiteral>
```
dg0yt: Private implementation detail of QString. Do not include. (May need an update of `iwyu-qt.imp`.)

(It doesn't even exist on my computer.)

src/fileformats/ocd_georef_fields.cpp
lpechacek: Renaming `MapperCrsRecord::translateOcdToMapper()` to `MapperCrsRecord::fromOcd()` would break symmetry. `translateOcdToMapper()` and `translateMapperToOcd()` form a pair. Either we rename both (to `fromOcd()` and `toOcd()`) or none.

src/fileformats/ocd_georef_fields.cpp
lpechacek: So be it. A sensible compiler will inline `translateOcdToMapper()` into `applyGridAndZone()` so there are no gains nor losses from the signature change. I'll also drop `OcdCrsRecord`, which is the same case.

src/fileformats/ocd_georef_fields.cpp
lpechacek: Fun to watch you fix your own code. This line came from commit 630987577 through a few reviews and rewrites. Only in "OCD: Make OcdGeoref even more isolated" `scale` was renamed to `m`. Will apply the change.

src/fileformats/ocd_georef_fields.cpp
lpechacek: Understood. Thanks for the hint!

test/georef_ocd_mapping_t.cpp
lpechacek: OK, will weed these out. Thanks!

test/georef_ocd_mapping_t.cpp
lpechacek: > Even in case of error, a, m, x, y need to be properly set.

They are. See above.

> In addition, the test needs to verify that a warning was added (probably another test data field).

I had that in previous iterations but dropped the extra field. IMO it just cluttered the code while adding little value.

> For verifying x and y, you must test with projected coordinates instead of geographic coordinates.

I don't see why I exactly *must* switch to projected coords but the point about transformation is persuasive. I'll look into it.

src/fileformats/ocd_georef_fields.cpp
lpechacek: They are already initialized via default member initialization to sensible defaults: https://github.com/lpechacek/mapper/blob/e12b60093dbb521c0e6063c1d50150026729016d/src/fileformats/ocd_georef_fields.h#L41-L46

src/fileformats/ocd_georef_fields.cpp
dg0yt: My argument (in the previous comment) was not a sensible compiler, but comprehensibility for humans. "Named constructor" is one point, and avoiding arguments of same type next to each other is another point. Cf.
- [C++ FAQ on the named constructor idiom](https://isocpp.org/wiki/faq/ctors#named-ctor-idiom)
- [C++ Core Guidelines:  Avoid adjacent unrelated parameters of the same type](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Ri-unrelated)
- Recent blog entry linked on isocpp.org on [certain constructors in the STL which got it wrong](https://isocpp.org/blog/2018/03/strings-competing-constructors-andrzej-krzemieski).

Note that the named constructors don't interfere with the Rule of Zero.

src/fileformats/ocd_georef_fields.cpp
dg0yt: It would be much less funny if I wouldn't notice the errors I introduced... Feel free to fix them faster than I can ;-)

src/fileformats/ocd_georef_fields.cpp
dg0yt: This is not about default values: It is about preserving as much input from OCD as possible when we hit an unknown grid. a, m, x, y will cause no trouble.

(Ideally, we would preserve the grid/zone parameter even if we don't understand the value. We might even record the adjustment to the CRS the user makes in Mapper so we can suggest this the next time we hit then unknown grid.)

test/georef_ocd_mapping_t.cpp
dg0yt: >> For verifying x and y, you must test with projected coordinates instead of geographic coordinates.
>
> I don't see why I exactly must switch to projected coords but the point about transformation is persuasive. I'll look into it.

With geographic coordinates, you bring PROJ.4 conversions into the code path under test. IMO this is conceptually wrong because it introduces complexity and more modes of error (PROJ.4 installation, additional floating point calculations) when we actually want to test a piece of code which does not deal (directly) with geographic coordinates. It is simply not this test's responsibility: Coordinate conversions with PRO.4 are to be covered by `GeoreferencingTest`.

The conversion to geographic coordinates also blocks the testing of the graceful (i.e. parameter-preserving) handling of unknown grids.

src/fileformats/ocd_georef_fields.cpp
lpechacek: Parameter preservation is a new concept here. I'll have a look.

lpechacek:
Changes in this revision:
* fixed obvious coding style violation
* reduced number of patches

test/georef_ocd_mapping_t.cpp
lpechacek: ack

src/fileformats/ocd_georef_fields.cpp
lpechacek: I'm aware of the recommendations and acknowledge that this parameter arrangement is against recommendations stated in Guidelines. I question whether fixing parameter count is worth the effort associated with rework of the patch set. Anyway, the other points warrant the respin so this one is included.

test/georef_ocd_mapping_t.cpp
lpechacek: > >    In addition, the test needs to verify that a warning was added (probably another test data field).
>
> I had that in previous iterations but dropped the extra field. IMO it just cluttered the code while adding little value.

So I gave the number_of_warnings test second chance and it's back. Updated patch set will follow shortly.

lpechacek:
Changes in this revision:

*  `georef_ocd_mapping_t` checks for warnings again

huema-ch:
**list of all ocad-grid-id's:**
for an own project i made a csv-file, which contains for all ocad-grid-id's ( found in ocad 12) the corresponding epsg-code. 
I think this list could be interesting for you, if you want to add more ocad-grid-id's to the mapper.

you find the list in github under [https://github.com/huema-ch/OCAD-Grid-ID_to_EPSG](url)

lpechacek:
@huema-ch, the table is a real gift! How did you collect information about the mappings, please?

huema-ch:
@lpechacek, actually quit simple. I went through all the selectable coordinate systems in ocad (under map -> coordinate-system and scale) and wrote the grid-id and the epsg-code to the list.

krticka:
Nice to have such extensive list of CRS. I have few notes.
Missing description for some CRS, which are deprecated. This should be shown, also with information that CRS is DEPRECATED.
http://epsg.io/2244
http://epsg.io/2245
http://epsg.io/3454

There are sometimes inconsistencies within existing CRS databases - EPSG, ESRI, and User-defined (called sr-org). For example 4685 in EPSG is deprecated CRS for the Democratic republic of the Congo, but in SR-OGR it is Gauss Boaga Fuso Ovest (Italy). 

Probably this list will need some revision in future.

lpechacek:
Agreed. Seeing the sheer number of codes, I'd wonder if there were no bugs in it. They can come both from OCAD and the rewrite done by huema-ch.

Anyway, decision about the last commit should be purely software engineering one. We have unreliable source of information - how much information do we take from it? Take the whole table with the risk of having to fix many entries or only part of it with the risk of throwing away good entries? Or take it as whole and drop those which turn out to be wrong over time? Or invent some method of entry verification to be applied on all of them.

Let's wait for dg0yt to state his preference though.

dg0yt:
We need to keep some things separate:
1. The mapping of OCD codes to EPSG codes (or explicit specifications)
2. The EPSG code definitions (or explicit specifications), and comments about their deprecation or similar.
3. Providing a more comfortable GUI for selecting a suitable CRS, e.g. based on continent and country. 

The key addition of this pull request is to deal with the first point, i.e. implementing a mapping from OCD to Mapper and back. I would agree that the reference table could go to a file, for more flexibility. Errors found in this mapping can be handled by us, and users might edit the file in the meantime.

Regarding the second: Mapper uses PROJ.4 internally. (Via GDAL, it might even do much more, e.g. WKT). EPSG definitions come with PROJ.4. The `epsg` file has comments which even provide deprecation remarks:

https://github.com/OSGeo/proj.4/blob/ee22447de347c78944af849b10e0c9fd10184c28/nad/epsg#L1198-L1199

We should use, not duplicate, this information. Errors in these definition need to be reported upstream.

Regarding the third: This is included in this pull request, with a bias on the OCD side. Maybe this needs some more work on generalization. I would prefer to reuse country codes from `QLocale` which gives us translated country names for free. 

I didn't want to ask for more changes before merging what we have, because each desired change came with other changes which I was more sceptical of. The conversation is already quite long.

krticka:
Agree with @dg0yt.
1. Mapping OCD codes to EPSG must be separate.
Regarding 2 and 3 we for sure will need more data than epsg string provides. 
QGIS 3 has nice interface for selection of CRS showing also area of use, but I am not sure how the information is stored, probably in some files of c:\OSGeo4W64\share\epsg_csv\
There is also option to gather information online from Epsg.io API https://github.com/klokantech/epsg.io
For example http://epsg.io/?q=32638&format=json will return enough information including bbox and description.

PR #1041, opened 2018-02-02
Remove qt_lv.ts

sikmir:
- Latvian Qt translation is available from upstream starting from Qt 5.6.0 ([qtbase_lv.ts](https://code.qt.io/cgit/qt/qttranslations.git/tree/translations/qtbase_lv.ts)).
- SuperBuild provides Qt 5.6.2 ([qt-5.6.2-2.cmake](https://github.com/OpenOrienteering/superbuild/blob/master/qt-5.6.2-2.cmake)).
- ~~In case of Latvian there is only `qtbase_lv.qm`, but no `qt_lv.qm`, so Mapper should try to load both files.~~


dg0yt:
1. I would propose to remove qt_lv.qm from `dev` not from `master`. Ubuntu 16.04 with Qt 5.5.1 is still a current LTS. Of course this means that qt_lv.qm will remain visible a little bit longer in Weblate.
2. [Qt documentation](https://doc.qt.io/qt-5/linguist-programmers.html#deploying-translations) recommends not to deploy (read: load) split translation files, but to simply concatenate the required files. (I don't know if this is the right guideline for Linux packages.)
3. Concatenation is implemented in `packaging/CMakeLists.txt:
https://github.com/OpenOrienteering/mapper/blob/5c9dbedff22ef2b271a2541ddabd166813709e10/packaging/CMakeLists.txt#L56-L59

However, looking at the translations in OpenOrienteering-Mapper-0.8.20180121-Windows-x64.zip, I suspect it is not working as expected.

dg0yt:
Qt translation deployment is now fixed in e33ef24.
Maybe we could indeed remove qt_lv.ts now, and I would re-add it as a patch where needed.

dg0yt:
Also Qt Linguist loads only qt_..., not qtbase_.... The absence of qt_lv.qm is a bug in Qt and/or Linux distribution packages. I'm not sure if we want to add a workaround in Mapper source code. I didn't spot another language which would be affected.

sikmir:
The absence of qt_lv.qm is a bug in Qt (see [git tree](https://code.qt.io/cgit/qt/qttranslations.git/tree/translations/)).

dg0yt:
`qt_lv.ts` [happens to be fixed in Qt 5 on 19 March](http://code.qt.io/cgit/qt/qttranslations.git/commit/?id=1c22aa4029920619404c3162656b5749189ea6f8), going to land in (qttranslations for) 5.6.4,  5.9.5 and 5.11.0.

Will take a while until we can remove our translate completely, but we could switch to the upstream version in the meantime. And we can take as confirmed that we don't want to load `qtbase_*.qm`.

sikmir:
Good news about Qt. Updated PR (removed loading `qtbase_*.qm`).

PR #1049, opened 2018-02-15
Android: Fix baseline view rendering

lpechacek:
Submitting this tiny fix so that it's not lost.

### DIFF
--- a/src/core/renderables/renderable.cpp
+++ b/src/core/renderables/renderable.cpp
@@ -715,7 +715,7 @@ bool PainterConfig::activate(QPainter* painter, const QPainterPath*& current_cli
 	if (mode == PainterConfig::PenOnly)
 	{
 #ifdef Q_OS_ANDROID
-		if (pen_width * config.scaling < 0.1)
+		if (pen_width > 0 && pen_width * config.scaling < 0.1)
 			return false;
 #endif
 		if (config.testFlag(RenderConfig::ForceMinSize) && pen_width * config.scaling <= 1.0)


dg0yt:
The commit message is fine. It is a pity it is somewhat hidden in the Github PR page.

dg0yt:
.. and now this change fixes #1172, at least the bug part of it.

dg0yt:
I took a different approach in 0230da0. I wanted to create a pull request for this and a few more changes, https://github.com/OpenOrienteering/mapper/compare/aef7460...23cdf35, but accidently pushed to `master`. I don't want to reset the protected master, and I will even stage new master builds.
Could you perhaps have a critical look at these changes anyway?

lpechacek:
I'm glad to see multiple issue reports being addressed by the series. AFAICT commit 0230da0e0f12 includes the functionality of my change, obsoleting this PR.

Overall, I don't completely understand the aim for CPU cycle savings on desktop but that's perhaps something to talk about when we meet in person. As expected, the code change alters map appearance when zooming out. For example, in my 1:10 000 test map light brown shades caused by downsized contours, and green shades from undergrowth, disappear at 0.15x zoom level.

In 23cdf355f7e2 ("Don't default to GDAL for GPX") it looks to me like the default is disabled for DXF and not GPX. Am I misreading the code or is it a bug?

Thanks!

dg0yt:
> I don't completely understand the aim for CPU cycle savings on desktop [...]

I'm not to keep on that, but I would very much like to replace the Q_OS_ANDROID hack by a better solution which works for desktop and mobile. As I said, some commits went into `master` too quickly. It will probably become clearer when I manage to place the next pull request.

dg0yt:
> In 23cdf35 ("Don't default to GDAL for GPX") it looks to me like the default is disabled for DXF and not GPX. Am I misreading the code or is it a bug?

This is a bug. This is why we need code reviews ;-)

PR #1053, opened 2018-02-16
Add missing Qt messages (printing dialogs)

sikmir:


dg0yt:
Which Qt version is this for?

I suppose most of it has no use on Windows. What about macOS?

sikmir:
1) It's for Qt >= 5.3.
2) At least CUPS-related stuff seems to be used only on Linux/FreeBSD. As far as I see on macOS another system dialog is used.

dg0yt:
My idea was to deal with Qt translations only for the platforms where we deploy Qt. This is Android, macOS, Windows. Otherwise it is just too many versions of Qt, and another package's responsibility anyway. What do you think?

sikmir:
Well, there is something to think over, so let's postpone a bit this PR. I'll be back in 3 weeks, right now I'm going away, so have no time.

Issue #783, opened 2016-10-26
Feature: Android: Generate scribble layer


mlerjen:
The generation and placement of a geo-referenced scribble layer for the android version is pretty much complicated at the moment. Especially for beginners. I would therefore like to propose you add a "add scribble layer" functionality to the android version.
The function could be triggered in the template manager under "+">"add scribble layer" giving two options: resolution (? necessary?) and extent (entire map/actual window).
The result would be a geo-referenced transparent .png with the chosen resolution and with a 3px(?) red border added as layer to the template stack.

dg0yt:
Basically this was on the list of ideas when Thomas published the Android prototype. Good to have it as an issue now.

(To be honest, I always used freehand drawing with symbols instead. But the scribble feature avoids changing the original file which mitigates our OCD limitations.)

ollesmaps:
I agree, this would be a welcomed feature. It could perhaps be also possible to make a new .omap file with the android app. 
I can imagine situations I am somewhere out and there is a sudden need of mapping. (For example a course planner encounters a new clearing).

ollesmaps:
Resolution is necessary, how many meters in pixel.
The rectangle could be defined by two points - so a rectangle not necessarily to window size can come up.

There are online image generators, as https://dummyimage.com/
So generating just .pgw might be an option.

dg0yt:
> There are online image generators, as https://dummyimage.com/
Well, generating the blank image is the least difficult part.

Especially on Android, there is hardly an alternative to doing it in Mapper. You may be offline, and Android security design could make it hard to put a download into a writable folder belonging to the Mapper app.

mlerjen:
A general comment (android, fieldwork): The size of the scribble layer has a crucial effect on the time span blocked by the autosave. While it is around 5 seconds for a 500KB raises to a painful 25-30 seconds for a 4.5MB. It is therefore advisable to split the scribble layer over an entire map project into multiple parts resp. work with multiple scribble layers. 
This also lowers other risks linked with the scribble layer as the container of your fieldwork/value. I suffered three losses so far, all related to the file size of a single whole-project scribble layer (2x incomplete cloud uploads, 1x incomplete saving before low battery shutdown).

ollesmaps:
Does the autosave time relate also to RAM, frequency?
I keep the scribble layer in maximum size something around 4000*4000 pixels. But that is the limit afterwards it takes too loong to save.
[...]

PR #1064, opened 2018-02-28
Add option to export a world file when exporting an image

### LINKED ISSUE #463: Export image with worldfile
**Somebody reported on Sourceforge [[tickets:#463](https://sourceforge.net/p/oorienteering/tickets/463)]:**

It would be very useful to be able to export raster images with georeferencing information. A checkbox in the image export-dialog such as 'include worldfile' would be a nice way to do it.

mpickering:
This patch adds a checkbox which can be ticked when exporting
an image to save a wld file in the same location as the image. This
allows users to load maps into other GIS programs such as QGIS and
retain the georeferencing.

Fixes #463

(This is the first time I have ever written in C++ so there might be some stupid things I'm doing)

### DIFF
--- a/src/core/map_printer.h
+++ b/src/core/map_printer.h
@@ -165,10 +165,6 @@ public:
 	 *  Only available in MapPrinterOptions::Raster mode.
 	 */
 	bool simulate_overprinting;
-
-	/** Controls whether to save a world file when outputting an image
-	 */
-	bool save_world_file;
 };
 
 
@@ -347,9 +343,6 @@ public slots:
 	
 	/** Controls whether to print in overprinting simulation mode. */
 	void setSimulateOverprinting(bool enabled);
-
-	/** Controls whether to save a world file */
-	void setSaveWorldFile(bool enabled);
 	
 	/** Controls the color mode. */
 	void setColorMode(MapPrinterOptions::ColorMode color_mode);
--- a/src/core/map_printer.h
+++ b/src/core/map_printer.h
@@ -165,6 +165,10 @@ public:
 	 *  Only available in MapPrinterOptions::Raster mode.
 	 */
 	bool simulate_overprinting;
+
+	/** Controls whether to save a world file when outputting an image
+	 */
+	bool save_world_file;
 };
 
 
@@ -343,6 +347,9 @@ public slots:
 	
 	/** Controls whether to print in overprinting simulation mode. */
 	void setSimulateOverprinting(bool enabled);
+
+	/** Controls whether to save a world file */
+	void setSaveWorldFile(bool enabled);
 	
 	/** Controls the color mode. */
 	void setColorMode(MapPrinterOptions::ColorMode color_mode);
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -1235,14 +1235,7 @@ void PrintWidget::exportWorldFile(const QString& path) const
 
 	QTransform final_wld(xscale, yskew, 0, xskew, yscale, 0, top_left.x(), top_left.y());
 	WorldFile wld(final_wld);
-
-	// Compute the path for the file which is the original file name with the suffix replaced with wld
-	QFileInfo file_info(path);
-	QString wld_path = file_info.canonicalPath()
-	                   + QLatin1Char('/')
-	                   + file_info.completeBaseName()
-	                   + QStringLiteral(".wld");
-	wld.save(wld_path);
+	wld.save(WorldFile::pathForImage(path));
 }
 
 void PrintWidget::exportToPdf()
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -333,7 +333,6 @@ PrintWidget::PrintWidget(Map* map, MainWindow* main_window, MapView* main_view, 
 	connect(show_templates_check, &QAbstractButton::clicked, this, &PrintWidget::showTemplatesClicked);
 	connect(show_grid_check, &QAbstractButton::clicked, this, &PrintWidget::showGridClicked);
 	connect(overprinting_check, &QAbstractButton::clicked, this, &PrintWidget::overprintingClicked);
-	connect(world_file_check, &QAbstractButton::clicked, this, &PrintWidget::worldFileClicked);
 	connect(color_mode_combo, &QComboBox::currentTextChanged, this, &PrintWidget::colorModeChanged);
 	
 	connect(preview_button, &QAbstractButton::clicked, this, &PrintWidget::previewClicked);
@@ -606,6 +605,9 @@ void PrintWidget::setTarget(const QPrinterInfo* target)
 		printer_properties_button->setEnabled(is_printer);
 
 	world_file_check->setVisible(!is_printer);
+	// If MapCoord (0,0) maps to projected (0,0), then there is probably
+	// no point in writing a world file.
+	world_file_check->setChecked(!map->getGeoreferencing().toProjectedCoords(MapCoordF{}).isNull());
 	
 	bool is_image_target = target == MapPrinter::imageTarget();
 	vector_mode_button->setEnabled(!is_image_target);
@@ -1092,11 +1094,6 @@ void PrintWidget::overprintingClicked(bool checked)
 	map_printer->setSimulateOverprinting(checked);
 }
 
-void PrintWidget::worldFileClicked(bool checked)
-{
-	map_printer->setSaveWorldFile(checked);
-}
-
 void PrintWidget::colorModeChanged()
 {
 	if (color_mode_combo->currentData().toBool())
@@ -1211,8 +1208,8 @@ void PrintWidget::exportToImage()
 	else
 	{
 		main_window->showStatusBarMessage(tr("Exported successfully to %1").arg(path), 4000);
-		if (map_printer->getOptions().save_world_file)
-			exportWorldFile(path);
+		if (world_file_check->isChecked())
+			exportWorldFile(path);  /// \todo Handle errors
 		emit finished(0);
 	}
 	return;
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -1218,7 +1218,7 @@ void PrintWidget::exportToImage()
 	return;
 }
 
-const void PrintWidget::exportWorldFile(QString path)
+void PrintWidget::exportWorldFile(const QString& path) const
 {
 	const auto& georef = map->getGeoreferencing();
 	const auto& ref_transform = georef.mapToProjected();
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -77,6 +77,7 @@
 
 #include <printer_properties.h>
 
+#include "core/georeferencing.h"
 #include "core/map.h"
 #include "core/map_printer.h"
 #include "core/map_view.h"
@@ -88,6 +89,7 @@
 #include "gui/map/map_editor.h"
 #include "gui/map/map_widget.h"
 #include "templates/template.h" // IWYU pragma: keep
+#include "templates/world_file.h"
 #include "util/backports.h"
 #include "util/scoped_signals_blocker.h"
 
@@ -264,6 +266,10 @@ PrintWidget::PrintWidget(Map* map, MainWindow* main_window, MapView* main_view, 
 	
 	overprinting_check = new QCheckBox(tr("Simulate overprinting"));
 	layout->addRow(overprinting_check);
+
+	world_file_check = new QCheckBox(tr("Save world file"));
+	layout->addRow(world_file_check);
+	world_file_check->hide();
 	
 	color_mode_combo = new QComboBox();
 	color_mode_combo->setEditable(false);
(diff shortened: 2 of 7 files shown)


src/templates/world_file.cpp
```diff
@@ -73,6 +80,31 @@ bool WorldFile::load(const QString& path)
...
+    parameters[5] = pixel_to_world.m32();
+
+    QFile file( path );
+    if ( file.open(QIODevice::WriteOnly))
+    {
+        QTextStream stream(&file);
+        for (auto& parameter : parameters)
+        {
+            stream << parameter << endl;;
+        }
```
dg0yt: Double `;`.

Instead of a loop, I would prefer to have six explicit statements which access `pixel_to_world.mXX()` directly. It would be less code in total (no need to setup `parameters`), and it would express more clearly what is written to the stream.

src/templates/world_file.cpp
```diff
@@ -73,6 +80,31 @@ bool WorldFile::load(const QString& path)
 	return true;
 }
 
+bool WorldFile::save(const QString &path)
+{
+    if (!loaded) { return false;}
```
dg0yt: Please add an empty line after an early `return` so that it becomes more visible.

src/core/map_printer.h
```diff
@@ -165,6 +165,10 @@ class MapPrinterOptions
 	 *  Only available in MapPrinterOptions::Raster mode.
 	 */
 	bool simulate_overprinting;
+
+    /** Controls whether to save a world file when outputting an image
+     */
+    bool save_world_file;
```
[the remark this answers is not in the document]
mpickering: Perhaps I should move `exportWorldFile` to `MapPrinter` like `drawImage` is in `MapPrinter`?

ollesmaps:
Maybe this way can be generated also the scribble layer. See  #783

src/gui/print_widget.cpp
```diff
@@ -1197,11 +1212,47 @@ void PrintWidget::exportToImage()
...
+    OpenOrienteering::Georeferencing geo = map->getGeoreferencing();
+
+    MapCoord top_left(map_printer->getPrintArea().topLeft());
+    MapCoord bot_right(map_printer->getPrintArea().bottomRight());
+    MapCoord extent (geo.toProjectedCoords(bot_right) - geo.toProjectedCoords(top_left));
+
+    // Calulate distance per pixel
+    MapPrinterPageFormat page_format = map_printer->getPageFormat();
+    int w = qCeil(page_format.paper_dimensions.width() * pixel_per_mm);
+    int h = qCeil(page_format.paper_dimensions.height() * pixel_per_mm);
```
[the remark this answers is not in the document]
mpickering: Does `qCeil` not return an `int`? There was also some similar code in `drawPage` which used `int`. What is a more appropriate type?

mpickering:
I handled all your comments and managed to simplify the implementation when thinking about how to deal with the rotation correctly. The file you suggested doesn't get aligned quite correctly but I think this is a precision issue as the generated image is quite small.

mpickering:
@ollesmaps I don't think the two features are related at all.

src/core/map_printer.h
lpechacek: `save_world_file` member should be initialized in class constructor. Otherwise UBSAN complains:
```
/home/lpechacek/oomapper/src/core/map_printer.cpp:720:14: runtime error: load of value 116, which is not a valid value for type 'bool'
```

mpickering: Fixed thanks.

dg0yt:
@lpechacek  What is your opinion about these points:

1. Do not add `save_world_file` to `MapPrinterOptions`. Rationale: It is not really a printing option. It is never evaluated by `MapPrinter`. There is no point in saving it with printer options.

2. Initialize the world file checkbox depending on georeferencing. Rationale: It is convenient to export a world file by default, unless there is no georeferencing. Even a local georeferencing is enough, so we would probably only need to look for a non-zero projected reference point.

3. Do we want to add this already in `master`? Rationale: It is a quite small change adding something really useful, and it won't break anything if doesn't work. `dev` may still take some time to be released because it contains changes which need much more testing before being ready for release.

src/gui/print_widget.cpp
```diff
@@ -1222,20 +1222,16 @@ void PrintWidget::exportToImage()
...
-	const auto xscale = extent.width() / w;
-	const auto yscale = extent.height() / h;
-
-	// Not sure that the rotation value is right here but it looks correct in my tests
-	QTransform final_wld(xscale, 0, 0, 0, yscale, 0, top_left.x(), top_left.y());
+
+	const auto xscale = ref_transform.m11() / pixel_per_mm;
+	const auto yscale = ref_transform.m22() / pixel_per_mm;
+	const auto xskew = ref_transform.m12() / pixel_per_mm ;
+	const auto yskew = ref_transform.m21() / pixel_per_mm ;
```
lpechacek: The export dialog allows setting _image DPI_ and _map scale_. Both parameters must be taken into account when calculating `pixels_per_mm`.

Quick fix:
```
diff --git a/src/gui/print_widget.cpp b/src/gui/print_widget.cpp
index e2df9fa99d3b..0bfb25d6b5c9 100644
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -1225,11 +1225,12 @@ const void PrintWidget::exportWorldFile(QString path, qreal pixel_per_mm)
        const auto& ref_transform = georef.mapToProjected();
 
        const auto top_left     = georef.toProjectedCoords(MapCoord{map_printer->getPrintArea().topLeft()});
+       const auto pixel_per_mm_adjusted = pixel_per_mm * map_printer->getScaleAdjustment();
 
-       const auto xscale = ref_transform.m11() / pixel_per_mm;
-       const auto yscale = ref_transform.m22() / pixel_per_mm;
-       const auto xskew = ref_transform.m12() / pixel_per_mm ;
-       const auto yskew = ref_transform.m21() / pixel_per_mm ;
+       const auto xscale = ref_transform.m11() / pixel_per_mm_adjusted;
+       const auto yscale = ref_transform.m22() / pixel_per_mm_adjusted;
+       const auto xskew = ref_transform.m12() / pixel_per_mm_adjusted;
+       const auto yskew = ref_transform.m21() / pixel_per_mm_adjusted;
 
        QTransform final_wld(xscale, yskew, 0, xskew, yscale, 0, top_left.x(), top_left.y());
        WorldFile wld(final_wld);
```

Given that `exportWorldFile()` method has access to `map_printer`, the adjusted `pixels_per_mm` should IMHO be calculated in situ and the method parameter should be dropped.

src/gui/print_widget.cpp
```diff
@@ -264,6 +267,10 @@ PrintWidget::PrintWidget(Map* map, MainWindow* main_window, MapView* main_view,
 	
 	overprinting_check = new QCheckBox(tr("Simulate overprinting"));
 	layout->addRow(overprinting_check);
+
+    world_file_check = new QCheckBox(tr("Save world file"));
+    layout->addRow(world_file_check);
+    world_file_check->hide();
```
lpechacek: In my opinion, World File generation should not be offered in case map georeferencing is not set. Not sure if `isLocal()` is the right method to check georeferencing validity as `Georeferencing`'s internal state handling looks somewhat conflated to me.

```
diff --git a/src/gui/print_widget.cpp b/src/gui/print_widget.cpp
index ebdec3c758c3..4755a1654337 100644
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -271,6 +271,8 @@ PrintWidget::PrintWidget(Map* map, MainWindow* main_window, MapView* main_view,
        world_file_check = new QCheckBox(tr("Save world file"));
        layout->addRow(world_file_check);
        world_file_check->hide();
+       if (map->getGeoreferencing().isLocal())
+               world_file_check->setEnabled(false);
        
        color_mode_combo = new QComboBox();
        color_mode_combo->setEditable(false);
```

... or shorter `setEnabled(!...isLocal());`.

src/gui/print_widget.cpp
```diff
@@ -1197,11 +1212,37 @@ void PrintWidget::exportToImage()
...
 	}
 	return;
 }
 
+const void PrintWidget::exportWorldFile(QString path, qreal pixel_per_mm)
+{
+	const auto& georef = map->getGeoreferencing();
+	const auto& ref_transform = georef.mapToProjected();
+
+	const auto top_left     = georef.toProjectedCoords(MapCoord{map_printer->getPrintArea().topLeft()});
```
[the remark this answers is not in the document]
lpechacek: Even with this change I see misalignment between original map and exported image.

image
Map is a georeferenced (EPSG:5514) 1:15000 work. Image was exported at 300 DPI.

lpechacek:
I'll have to look into code before answering point 1.

>  Initialize the world file checkbox depending on georeferencing. Rationale: It is convenient to export a world file by default, unless there is no georeferencing. Even a local georeferencing is enough, so we would probably only need to look for a non-zero projected reference point.

Sounds good. How about going one step further and exporting World File unconditionally? Rationale - the export dialog has already a lot of options.

>  Do we want to add this already in master? Rationale: It is a quite small change adding something really useful, and it won't break anything if doesn't work. dev may still take some time to be released because it contains changes which need much more testing before being ready for release.

No objection from me. The code change is safe for a maintenance release. The only downside is that v0.9 will have one "new feature" check mark less.

ollesmaps:
@mpickering I will decsribe how I get a scribble layer now. I need Ocad for that. From there I export a georeferenced blank white image of area I need - jpeg with jgw. I then switch to graphic program and convert jpeg to png, rewrite jgw into pgw. ---> voila there is a sribble layer I can use in Mapper.

If these steps can be done within Mapper, that would make things more straightforward.

mpickering:
@ollesmaps This would be trivial to implement in mapper itself and would not be made any easier by this patch.

src/gui/print_widget.cpp
```diff
@@ -1197,11 +1212,37 @@ void PrintWidget::exportToImage()
...
+
+	QTransform final_wld(xscale, yskew, 0, xskew, yscale, 0, top_left.x(), top_left.y());
+	WorldFile wld(final_wld);
+
+	// Compute the path for the file which is the original file name with the suffix replaced with wld
+	QFileInfo file_info(path);
+	QString wld_path = file_info.canonicalPath()
+	                   + QLatin1Char('/')
+	                   + file_info.completeBaseName()
+	                   + QStringLiteral(".wld");
```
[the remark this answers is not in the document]
mpickering: The world files exported by QGIS use the `.wld` extension which is why went with that. It wasn't clear which was best to me.

src/gui/print_widget.cpp
mpickering: It seemed redundant to calculate `pixels_per_mm` again as it is calculated already in the function where `exportWorldFile` is called. I didn't want to duplicate the code, especially as I didn't understand where the magic number 25.4 comes from.

lpechacek: Print resolution is in Dots-Per-Inch and an inch has 25.4 mm. :)  Obviously, the constant needs a name.

lpechacek:
> I'll have to look into code before answering point 1.

Crawling through the code is giving me no joy today... One of the best moments was the discovery of `MapPrinter::isPrinter()`. Seems some _printers_ are not printers. :(  Will need more time to familiarize with the code.

src/gui/print_widget.cpp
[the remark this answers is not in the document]
mpickering: I think this is right but it took me a while to think about why. It is really the same as `0.5 * 1/pixels_per_mm` as `1/pixels_per_mm` is the same as `mm_per_pixel`.

mpickering:
I applied the suggestions and now the test file lines up perfectly. Perhaps try again @lpechacek it was probably the precision change which fixed the issues to do with precision you were seeing.

mpickering:
@lpechacek I disagree with exporting the .wld file by default. Not even QGIS does this.

dg0yt:
> Seems some printers are not printers

From PrintWidget's point of view, yes. There are image export and PDF export.

dg0yt:
@ollesmaps  wrote:
> ... export a georeferenced blank white image of area I need - jpeg with jgw. I then switch to graphic program and convert jpeg to png, rewrite jgw into pgw.

With this change, will be able to do a png/pgw export directly and in Mapper. But this is a pull request  submitting code. So please let our comments be focused on reviewing this particular change.

dg0yt:
@lpechacek wrote:
> How about going one step further and exporting World File unconditionally?

@mpickering wrote:
> I disagree with exporting the .wld file by default. Not even QGIS does this.

Actually I tend towards lpechacek's proposal, for georeferenced maps. 

Pro:
- There is no harm from this small text file.
- KISS. QGIS isn't the right example for modern/our user interface design.

Con:
- We might overwrite an existing file (but that's true also for the image).
- There might already be another world file with a different file name extension, causing confusion.
- Some users might (unconciously) use this file without being aware of world file limitations (missing CRS spec).
- Some users might not understand why they don't get the file when the map is not georeferenced.

When removing the checkbox, `MapPrinterOptions::save_world_file` becomes obsolete.

lpechacek:
> ... it was probably the precision change which fixed the issues to do with precision you were seeing.

Works like charm now! Thanks, @mpickering!

dg0yt:
Merged to master. Builds triggered. Thank you very much.

I'm still interested in opinions whether to remove the check box. Shall we open an extra issue for "voting"?

Issue #1066, opened 2018-03-02
ESRI Shapefiles Export


mpickering:
The standard vector file format for QGIS is the ESRI Shapefile. It would be a useful feature to be able
to export a map in this format.

It is possible to write ESRI files with GDAL. 

http://www.gdal.org/drv_shapefile.html

There is also an example of how to create the shape file using the library.

http://www.gdal.org/ogr_apitut.html

Specifically, I think that it would be simplest to export one layer for each symbol type. 

The use cases for this are interesting visualisation examples. For example, taking an urban map, underlying a DEM and giving the buildings some height would create an interesting visualisation experience. 

I haven't investigated how exactly the map information is stored but this seems plausible to implement.

dg0yt:
Effectively, this is general geospatial vector data export via GDAL/OGR. It is the same interface for Shapefile, DXF etc.

Issue #601, opened 2016-02-19
Feature discussion - export to GPX/KML


dominikzalewski:
I would like to know if people are interested in the following feature. Please comment.

Use case 1: Execute an orienteering training session with zero or minimal effort.
Use case 2: Setting up control points before a competition.

Extended description of 1.
I have to prepare trainings for myself. It's really hard to do when I want to execute more advanced training session like
- http://o-training.net/w/No-map_compass_training
- http://o-training.net/w/Map_memory
- or the one where you have only a fragment of map around a control point and the rest is blank
Without setting up control points in the woods trainings like these are simply impossible (at least for me).

Extended description of 2.
When preparing a competition and setting up control points before the race, there's usually time pressure. If you're in hurry, you're more likely to make a mistake. Sometimes people less experienced with setting up control points would like to help, but you don't want them to setup the kite in the wrong spot.

Can we have technology to help us in both of these use cases?

I came up with the following idea. The description below assumes that you have access to 'Suunto Ambit 3' (I own such device). But I have been also experimenting with 'GPS Essentials' application on Android that doesn't work exactly like Suuunto's navigation, but is close enough to be considered a valid replacement.

So one simple solution to use case 2 would be to setup control points carying a tablet that use use in OO Mapper for drawing. This would work, but usually you have other equipment in your hands (kites, stands) that it's impractical to carry a tablet. And this wouldn't work for use case 1.

What one needs is an audible feedback (as opposed to visual feedback) to signal that you are at the place were a control point is (or should be). So here's the prototype I have built and tested for few months now.

**Step A - draw a simple course in OO Mapper**
![screen shot 2016-02-18 at 05 31 39](https://cloud.githubusercontent.com/assets/13519572/13170994/92a1a644-d6f0-11e5-8267-86578304d73f.png)

**Step B - export the course as KML**
I have coded a bit in OO Mapper and built a custom version that has the option shown on the screenshot. The code is around 50 lines and is only 'experimental' and of low quality (has no unit tests and does not handle exceptional cases correctly - like if you don't have any simple course, or you have two of them). It's just to check whether a prototype would work.
![screen shot 2016-02-18 at 05 29 20](https://cloud.githubusercontent.com/assets/13519572/13171035/c04ea34e-d6f0-11e5-91ef-44cd532e3cd3.png)

[...]

Issue #109, opened 2012-10-27
GPX export


dg0yt:
**puzzlepaint reported on Sourceforge [[tickets:#109](https://sourceforge.net/p/oorienteering/tickets/109)]:**

Add gpx export functionality (for georeferenced maps). Should include the possibility to limit the export to selected objects.

PR #1068, opened 2018-03-03
GPX/KML/SHP export via GDAL/OGR

xc-racer99:
Ok, here is an implementation to solve  #1066 and #109  Doesn't really deal with #601 as that is more specific to courses from what I understand of it.

A few notes/comments/questions:
1) My C++ is fairly limited, so please let me know of any stupid mistakes I've made.
2) GPX driver has several quirks, most notably that area symbols don't exist.  I've tried to account for all of them.
3) Should this code be refactored into its own file?  If so, placed where?
4) I've used the KML GDAL driver for KML files as opposed to the more fully featured LIBKML driver (see http://www.gdal.org/ogr_formats.html for all drivers).  The advantages to the KML driver are that it will auto-convert coordinates to WGS 1984 (necessary for KML spec) and that it is built in by default.  The drawbacks are that it doesn't support interleaved data (ie adding a point to a layer, then adding a polyline, then adding another point is not possible - but the GPX driver doesn't support this either), plus the polygons appear to be only wire-frames and not filled in.
5) I've created the name field as the symbol name.  I haven't exported any of the fields that an object may have.
6) No styles have been exported.
7) Should there be an option to export entire map as opposed to just the selected symbols?
8) For bezier curves, I've dropped all of the control points.  This results in some distortion to the output, but I'm not sure how else to do it

Thanks!


dg0yt:
Thank you very much for your PR. At the moment I would classify it as a proof-of-concept. I will do a detailed review for improvement. But answering question 3 first:

`MapEditorController` is much too fat already.

GDAL-related code is in `src/gdal`. **There is a file `src/gdal/ogr_file_format.h`** which has the implementation of class `OgrFileImport`. This class can serve as a guideline for the (future) class `OgrFileExport` which can go to the same file. The export shall be an implementation of class `Exporter`. This provides a way of collecting errors and warnings without using the GUI directly.

We cannot use the C++ API of GDAL/OGR because it is prone to source incompatible changes, and we deploy with a great variety of GDAL versions on Linux, starting at 1.11.3. If you need to check for GDAL 2, you can use `#ifdef GDAL_DMD_EXTENSIONS`... Maybe the whole GDAL/OGR export feature can be limited to GDAL >= 2 if it gives you more export features (such as curves with some drivers).

Building with GDAL is still optional. But given that we manage to build it reliably even for Android for a while now, I wouldn't mind making at mandatory in `dev` by now.

`src/gdal/ogr_file_format.cpp` has a set of resource management utilites named `unique_...` which add modern C++ style memory management (`unique_ptr`) for GDAL C structs which need special cleanup. `OgrFileImport` shows how to use them.

`GdalManager` is another important utility class for dealing with options in a ways which makes them available in the settings GUI. It also takes care of doing initialization only once.

dg0yt:
Of course, `OgrFileExport` would be a tool for exporting the whole map in the first place. For operating on the selection, you can look at `MapEditorController::copy()` (for copying to clipboard): It creates a copy of the map with the selected objects only, and the exports them. Copying might not be the most efficient way, but it shouldn't be to expensive when exporting just a single path.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
 
+void MapEditorController::exportOgrClicked()
+{
+	const auto& georef = map->getGeoreferencing();
+
+	// Check if map is georeferenced
+	if (georef.getState() == Georeferencing::Local)
+	{
+		QMessageBox::warning(window, tr("No Georeferencing"), tr("The map is not georeferenced.  Cannot export."));
+		return;
```
dg0yt: You can export to some formats (e.g. DXF) even with a local georeferencing.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+	}
+
+	// Make sure GDAL can work with the georeferencing info
+	OGRSpatialReference o_spat_ref;
```
dg0yt: We cannot use C++ API.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+	}
+
+	QSettings settings;
+	QString export_directory = settings.value(QString::fromLatin1("exportFileDirectory"), QDir::homePath()).toString();
+
+	QString filename = QFileDialog::getSaveFileName(
+						   window,
+						   tr("Export GPX, KML, or SHP file"),
+						   export_directory,
+						   tr("GPS Exchange Format (*.gpx);;Keyhole Markup Language (*.kml);;Shapefile (*.shp)"));
```
dg0yt: This will probaly not be need with OgrFileExport. Anyway, both strings are not well suited for translation because they are going to change frequently: with every new format we use from GDAL.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+	{
+		psz_driver_name = "GPX";
+		gpx_quirks = true;
+	}
+	else if (filename.endsWith(QLatin1String(".kml"), Qt::CaseInsensitive))
+	{
+		psz_driver_name = "KML";
+	}
+	else
+		psz_driver_name = "ESRI Shapefile";
```
dg0yt: Style: Use braces again for consistency.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+
+	// GPX driver can't automatically overwrite files
+	if (gpx_quirks)
+	{
+		QFile file(filename);
+		if (file.exists())
+			file.remove();
```
dg0yt: It is better to not remove (replace) the file until export was successful.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+	}
+
+	GDALAllRegister();
```
dg0yt: GDAL 2.0. We don't want to call this every time. This is done in `GdalManager` on first use.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+	std::vector<const Object*> area_objects;
+	for (const auto& object: map->selectedObjects())
+	{
+		if (object->getSymbol()->getType() == Symbol::Point)
+			point_objects.push_back(object);
+		else if (object->getSymbol()->getType() == Symbol::Line)
+			line_objects.push_back(object);
+		else if (object->getSymbol()->getType() == Symbol::Area)
+			area_objects.push_back(object);
+	}
```
dg0yt: No text objects. 
Alternative to copying, you might iterate over the same selection multiple times.

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+	o_name_field.SetWidth(32);
+
+	if (point_objects.size() > 0)
+	{
+		auto po_point_layer = po_ds->CreateLayer("point_symbols", &o_spat_ref, wkbPoint, nullptr);
+		if (po_point_layer == nullptr)
+		{
+			QMessageBox::warning(window, tr("Error Exporting"), tr("Failed to create layer"));
+			GDALClose(po_ds);
+			CSLDestroy(papsz_options);
```
dg0yt: This cleanup can be left to `unique_...`

src/gui/map/map_editor.cpp
```diff
@@ -3901,6 +3911,318 @@ void MapEditorController::createSymbolWidget(QWidget* parent)
...
+			OGRFeature::DestroyFeature(po_feature);
+		}
+	}
+
+	// GPX doesn't support area symbols
+	if (gpx_quirks && area_objects.size() > 0)
+	{
+		QMessageBox::information(window,
+								 tr("GPX Export"),
+								 tr("The GPX format doesn't support area symbols.  These will not be exported."));
```
dg0yt: Tabs uses wrong here.

dg0yt:
One more remark: A stumbling stone with the `Exporter` interface is the fact that we cannot use `QIODevice` directly with GDAL but need to pass filenames. I worked around it in `OgrFileImport::doImport` but it is not so easy with export. Maybe you can simply start this class with different constructors and entry points.

xc-racer99:
Great, thanks for the review and suggestions.  I was unaware that there was a significant API difference between the various GDAL versions.  I'll slowly work along and try and figure things out.

dg0yt:
> I'll slowly work along and try and figure things out.

That's fine. It will really take much more work to get in shape for `Exporter`.

For clarity, there are three different types of 'export':
- Normal file saving (.omap, .ocd). This is what `Exporter` does.
- Export of a rectangular section of the map. This is what *File > Export as... > Image, PDF* does.
- Export of the current selection. This is what your PR does. There is a similarity in the implementation of copying to clipboad, but of course that code doesn't need a file dialog.

So while I propose to implement this as an `Exporter`, I'm aware of that this is a different thing.
If you want I can try to provide the skeleton, including how to deal with some stumbling stones I mentioned in review.

mpickering:
Thanks @xc-racer99 . Were you already working on this? 

I am going to try out this code and report back on my experience.

Some initial points I noticed.

1. There shouldn't be a message box after a successful export (there isn't when exporting an image).
2. When I import the layer back into mapper - it works fine but when I import it into QGIS, it seems that it is misaligned slightly. Any ideas why this could be?

mpickering:
image

The green is the correct position for the buildings. As you can see, the polygon is slightly offset.

xc-racer99:
Thanks @dg0yt I'll have a go at implemented it as an Exporter and ask for help if I get stuck.

@mpickering 
Yes, I had been working on this a few days before submitting the PR.  I wasn't sure if I wanted to do the necessary refactoring to get it upstream or not but decided to try.

For your image, it looks as if everything is rotated wrong.  Does it do that with the KML export as well?  

> I would also prefer an interface which exported one shapefile for each symbol type rather than having to select each symbol individually.

I'm not sure if I understand what you mean.  Are you wanting all roads to be exported to the same shapefile geometry (multiLineString as opposed to lineString)?  Or one layer for each symbol type as opposed to the point/line/area layers I have now?  Or just a better way to select objects?

mpickering:
> Yes, I had been working on this a few days before submitting the PR. 

It's a good coincidence we were both looking at doing this at the same time then. Perhaps attackpoint has something to do with bringing it to our consciousness? 

> Or one layer for each symbol type as opposed to the point/line/area layers I have now?

Yes precisely. It is a bit tedious to separate them out again filtering by name as you may have a lot of symbols. On the other hand, it's easy to combine together lots of vector layers if you want to identify  different symbols, footpaths and roads for example. 

The same thing happens with the `kml` export. My map is rotated in mapper so this does appear to be a case where the rotation is not corrected appropriately? I'm sure Kai will be identify quite quickly where the problem is. 

image

mpickering:
OTOH, here is the raster map image, exported with a world file (from #1064).

image

xc-racer99:
Alright, here's v2 using the Exporter interface and GDAL C API.  It also exports Text and Combined objects and isn't limited to whitelisted OGR drivers for GDAL >= 2.  I briefly looked at exporting curves, but GDAL only supports arcs, not bezier curves (bezier curves are limited to read-only for supported formats, and even then they are translated into line approximations).

A couple of new issues have come up:
1) The drivers enabled are the ones that can import OGR formats.  Unfortunately, you can't just add another FileFormat to deal with these as the extensions are the same - this then causes errors when trying to export because FileFormats.findFormatForFilename() returns the first FileFormat - which is the one for importing.  So we would need to:
-a) make FileFormats.findFormatForFilename() import/export aware. 
-b) in Map::exportTo(), go through all of the formats and find one that supports the given extension and supports export.
-c) it would be possible to create three separate FileFormats - one for drivers supporting only import, another for export, and a third that supports both.  However, this would be clunky from a users point of view.
2) I'm concerned that if this is left in the "Save As" dialog as-is, then when the user saves as an OGR format, and then makes changes to templates/symbols/etc and then closes the program, all of those changes have essentially "disappeared" as the file format doesn't support them.  An option would be to only have lossy export file types show up in a new "Save A Copy As" option (and not in the Save As), but this would also also include OCAD files.

dg0yt:
I will look into the (expected) problems around `FileFormat`. One idea is to review the notion of "lossy" exports. Ocd export, as supported in "File > Save" shouldn't be marked lossy. Instead a "preflight check" could test for actual problems before saving a file. The regular omap driver could benefit as well, e.g. when saving in an older version of the format.

"Save a copy as" is an option indeed, also for manual snapshots. (I think it is offered in Notepad++.) This could be done independently in a separate pull request.

In the first place, I would like to extent the Export as... image/PDF feature a) to export also map file formats, and b) to offer exporting just the current selection.

dg0yt:
I needed some more changes to make this compile with GDAL 1.11 (on Ubuntu 16.04). I will add some review comments. But I want to get the OCD export changes into `dev` first.

src/gdal/ogr_file_format_p.h
```diff
@@ -272,6 +297,50 @@ MapCoord OgrFileImport::toMapCoord(double x, double y) const
...
+{
+	Q_DECLARE_TR_FUNCTIONS(OpenOrienteering::OgrFileExport)
+
+public:
+	OgrFileExport(QIODevice* stream, Map *map, MapView *view);
+	~OgrFileExport() override {}
+
+	void doExport() override;
+
+	bool noIoStream() { return true; }
```
dg0yt: Just for completeness, it needs to be marked `override` here. And the implementation should go to the cpp file. (It is a virtual function, and not on a critical path.)

However I don't think it will remain like this. The name has no analogy in Mapper, and we already have flags for describing capabilities in `FileFormat`.

src/core/map.cpp
```diff
@@ -682,19 +682,37 @@ bool Map::exportTo(const QString& path, MapView* view, const FileFormat* format)
 	bool success = false;
 	if (file.open(QIODevice::WriteOnly))
 	{
-		try
+		if (exporter->noIoStream())
```
dg0yt: I don't fee comfortable with the two similar blocks (in `if`and `else`) following this line. However I can't recommend a particular change on the moment. I would like to come to this later.

(For GDAL 2.0, we might keep using QSaveFile if we can leverage [`/vsistdout/`](http://gdal.org/gdal_virtual_file_systems.html#gdal_virtual_file_systems_vsistdout) with `VSIStdoutSetRedirection()`. Alternatively, we might explicitly support file formats which operate directly on paths. Maybe we need the latter anyway to leverage the full potential of GDAL for import and export. As a third, long-term option, one could contribute to GDAL another `/vsi.../` which uses Qt.)

src/gdal/ogr_file_format.cpp
```diff
@@ -69,23 +70,22 @@
 
 // IWYU pragma: no_forward_declare QFile
 
-
 namespace OpenOrienteering {
 
 namespace ogr {
-	
-	class OGRDataSourceHDeleter
```
dg0yt: I wondered how you could delete this class without breaking existing code. The answer is that there is a second definition in ogr_file_format_p.h. Both definitions were identical, fortunately.
The lesson from this observation is that we should keep all these utilities together, in ogr_file_format_p.h.

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+}
+
+void OgrFileExport::doExport()
+{
+	// Choose driver and setup format-specific features
+	QString file_extn = QFileInfo(output_file).suffix();
+	GDALDriverH po_driver;
+
+#ifdef GDAL_DMD_EXTENSIONS
+	// GDAL >= 2.0
```
dg0yt: I didn't test/review the GDAL 2.0 code.

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+				if (qstrcmp(cap_open, "YES") != 0)
+					continue;
+				po_driver = driver_data;
+				break;
+			}
+		}
+	}
+#else
+	const char *psz_driver_name;
+	if (file_extn.compare(QString::fromLatin1("gpx"), Qt::CaseInsensitive))
```
dg0yt: `compare()` returns 0 on equality, which is `false` when converted to `bool`.  So you must write:
~~~
if (file_extn.compare(QString::fromLatin1("gpx"), Qt::CaseInsensitive) == 0)

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+		psz_driver_name = "GPX";
+	else if (file_extn.compare(QString::fromLatin1("kml"), Qt::CaseInsensitive))
+		psz_driver_name = "KML";
+	else
```
dg0yt: Don't assume it is Shape file here. Compare explicitly,  and make `else` lead to an error message.

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+		psz_driver_name = "ESRI Shapefile";
+
+	po_driver = GDALGetDriverByName(psz_driver_name);
```
dg0yt: This is for GDAL 1.x where OGR is fairly separate. So we need `OGRGetDriverByName()` instead.

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+		}
+	}
+
+	/**
+	 * Only certain drivers work without georeferencing info
+	 * Based on http://www.gdal.org/ogr_formats.html as of March 5, 2018
+	 * */
+	static const std::vector<QByteArray> local_drivers = {
+	    "ARCGEN", "BNA", "DWG", "DXF",
+	    "DXF", "CSV", "Geomedia", "INGRES",
```
dg0yt: Duplicate `"DXF"`

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+	if (line_objects.size() > 0)
+		createLineLayer();
+
+	if (area_objects.size() > 0)
+		createAreaLayer();
+}
+
+void OgrFileExport::createPointLayer()
+{
+	auto po_point_layer = GDALDatasetCreateLayer(po_ds.get(), "point_symbols", map_srs.get(), wkbPoint, nullptr);
```
dg0yt: In GDAL/OGR 1.x, we don't have `GDALDatasetCreateLayer()` but must use `OGR_DS_CreateLayer()`. There are multiple occurences of `GDALDatasetCreateLayer()`.

OGR 1.11 API documentation is here: http://gdal.org/1.11/ogr/index.html

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1416,354 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+	}
+
+	// Create output dataset
+	po_ds = ogr::unique_datasource(OGR_Dr_CreateDataSource(po_driver, output_file.toLatin1(), nullptr));
+	if (!po_ds)
+		throw FileFormatException(tr("Failed to create dataset.  Error: %1").arg(QString::fromLatin1(CPLGetLastErrorMsg())));
+
+	/**
+	 * Create collections of point, line, and areas that are selected
+	 * These have to be separate as some drivers required them to be added in a specific order
```
dg0yt: Creating collection means allocation memory. Even if it is not much, I would prefer to avoid it, and rather loop over the all objects multiple times, filtering as needed. I would build upon `Map::applyOnMatchingObjects()` (which takes two arguments, an "operation" function and a "condition" function). But I understand that this approach maybe a little bit hard when you have little experience with function objects and/or lambdas.

Maybe be it is also a disadvantage to always create layers in this way. For shape files, this means individual sets of files, so we would actually need to select a *directory* for exporting shape files, not a regular *file*.

src/gdal/ogr_file_format_p.h
```diff
@@ -25,6 +25,7 @@
 #include <QByteArray>
 #include <QCoreApplication>
 #include <QHash>
+#include <QTemporaryDir>
```
dg0yt: Unused.

src/gdal/ogr_file_format.cpp
dg0yt: Need to correct myself: There was no duplication. You moved it to `ogr_file_format_p.h`. So everything is okay here.

xc-racer99:
Ok, thanks for the review.  I'll go through the comments and work on the issues and spin up an Ubuntu 16.04 VM to test on as well.

src/gdal/ogr_file_format.cpp
xc-racer99: Ok, sounds good with regards to `Map::applyOnMatchingObjects()`  I guess I'll learn about lambdas and function objects :)

Most drivers don't support multiple geometry types in the same layer; adding the points, then the lines, then the area symbols was compatible with all drivers that I tested.  Do you have a preference on how symbols should be arranged into layers?  ie all of the same geometries in a singular layer like I have now, one layer per symbol, or something different?

With regards to shapefiles, should the layers be named <basename>-line, <basename>-point, etc so that we hopefully don't get conflicting names for files?

xc-racer99:
Ok, here's a new patch set.  Major differences are:
*Refactored to use `map->applyOnMatchingObjects`
*GDAL's vsimem is used as temporary storage.  It works with all GDAL versions mapper supports and enables the use of the QIODevice.  However, it does increase the memory consumption.  I tried /vsistdout/ and for the most part it worked, but unfortunately not for shapefiles or any other format that creates multiple files.  There is also no way to determine this at runtime either as they still show support for VSI even if it's not fully supported.
*DXF export now works (GDAL>=2)
*OgrFileFormat split into import and export formats - some formats only support one or the other.  This relies on #1081 to work properly - for testing, you can simply comment out the registration of OgrFileImportFormat in global.cpp
*Basic styling has been added.  Uses `symbol->guessDominantColor()`.  Note that I've used a style table, but written the style attributes directly as some drivers (KML) won't look them up in the style table.
*LIBKML driver is preferred to the KML driver when present.  This is totally optional, but the generic KML driver doesn't support exporting areas with fill, only wireframes.

I've tested compiling this with both GDAL 2.1 (Debian Stretch) and GDAL 1.11 (Ubuntu 16.04).  I haven't tested a Windows build.  Please check all of my pointers and make sure I haven't created any memory leaks or made any other rookie C++ mistakes.  I've made a couple of inline comments about things I'm not sure on.

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1383,593 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+						throw FileFormatException(tr("Failed to read from temporary file in memory"));
+					io_device->write(reinterpret_cast<const char*>(pointer.get()), length);
+				}
+				else
+				{
+					// Handle any "extra" files, eg dbf, shx for shapefiles
+					// VSIRename() is only for copying within same folder and can't be used here
+					QFile out_file(output_file_info.absolutePath()
+					               .append(QString::fromLatin1("/")).append(QString::fromLatin1(names.get()[i])));
+					if (!out_file.open(QIODevice::WriteOnly))
```
xc-racer99: Should an exception or warning be added if the attempted output file already exists?

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1383,593 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+
+		auto line_string = ogr::unique_geometry(OGR_G_CreateGeometry(wkbLineString));
+		auto num_coords = path->getCoordinateCount();
+		for (MapCoordVector::size_type i = 0; i < num_coords; i++)
+		{
+			// Linear approx of bezier curve
+			if (path->getCoordinate(i).isCurveStart())
+			{
+				for (double j = 0; j <= 10.0; j++)
+				{
```
xc-racer99: I've made each bezier curve into 10 sections - should this be dynamic based on the distance or a smaller/larger number?

src/gdal/ogr_file_format.cpp
```diff
@@ -1408,5 +1383,593 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+			throw FileFormatException(tr("Error reading info about temporary file in memory"));
+		}
+	}
+}
+
+void OgrFileExport::addPointsToLayer(OGRLayerH layer, const std::function<bool (const Object*)>& condition)
+{
+	const auto& georef = map->getGeoreferencing();
+
+	auto add_feature = [&](Object* object) {
```
xc-racer99: Is an implicit capture OK or should I be explicitly capturing all variables used below?

dg0yt:
Great. I saw your latest commits already this morning. I will need some more time for substantial feedback. 

Regarding the approximation of curves: Actually we already have a "flattened" structure, `PathCoordVector`, which exists as member `path_coords` for every `PathPart` in `PathPartVector PathObject::parts()`. This flattening is dynamic. However, I have no idea if this isn't too much in some cases.

xc-racer99:
Ok, thanks for pointing me to the flattened structure.  It is much better than what I was doing and actually results in a smaller file size than my method for a contour-heavy map.
This latest revision uses it and also adds a couple of missing const qualifiers.

However, I have now found a new bug - I'm assuming all combined symbols are area symbols while some can be just line symbols (eg Railway)

dg0yt:
> However, I have now found a new bug - I'm assuming all combined symbols are area symbols while some can be just line symbols (eg Railway)

You may even mix areas and lines, e.g. for "body of water with bank line".  For the new OCD export, I do some preprocessing (eliminationg empty elements, counting, classifying, sorting) before deciding how to export. Cf. [`OcdFileExport::exportCombinedSymbol()`](https://github.com/OpenOrienteering/mapper/blob/e97169236d38ac8bf33415f0816dadb80ef28d25/src/fileformats/ocd_file_export.cpp#L2056)

xc-racer99:
Ok, I think I think it's ready for review to find the next round of bugs/improvements :)  This version solves the combined symbol bug and a few minor coding fixes.  Thanks for pointing me to OcdFileExport, it meant I also picked up some missing nullptr checks 
@mpickering this also adds the option to export one layer per symbol.

dg0yt:
I'm about to merge a change to dev which allows importers and exporters to use file path strings directly (instead of forcing use of QIODevice). Cf. #1120, d828f2e, fd54ff3. This includes some more substantial changes the Exporter API.

xc-racer99:
Thanks for the heads up.  I've rebased the patches for the new exporter interface and removed the use of GDAL's vsimem.  Tested on Debian Stretch (GDAL 2.1.2) and Ubuntu 16.04 (GDAL 1.11.3).

src/gdal/ogr_file_format.cpp
```diff
@@ -1405,5 +1379,573 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+					if (!layer)
+					{
+						addWarning(QString::fromLatin1("Failed to export symbol %1, %2")
+						           .arg(symbol->getPlainTextName())
+						           .arg(QString::fromLatin1(CPLGetLastErrorMsg())));
+						continue;
+					}
+					addPointsToLayer(layer, [&symbol](const Object* object) {
+						const auto& sym = object->getSymbol();
+						return sym && sym->equals(symbol);
```
dg0yt: Can't we simply use `sym == symbol`? 
If you really want to merge equal symbols into a single layer, you have to make sure that you don't add objects twice when looping over all symbols.
(Multiple occurrences.)

src/gdal/ogr_file_format.cpp
```diff
@@ -1405,5 +1379,573 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+		if (OGR_L_CreateFeature(layer, po_feature.get()) != OGRERR_NONE)
+			throw FileFormatException(tr("Failed to create feature in layer"));
+	};
+
+	map->applyOnMatchingObjects(add_feature, condition);
+}
+
+OGRLayerH OgrFileExport::createLayer(const char* layer_name, OGRwkbGeometryType type)
+{
+#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(2,0,0)
```
dg0yt: This is great. It should replace the `#ifdef GDAL_DMD_EXTENSIONS` in the existing code.

dg0yt:
And please take time to review all source strings for translation (inside `tr(...)`) if they are clean and final. They are more difficult to change once we got translations.

src/gdal/ogr_file_format.cpp
```diff
@@ -1405,5 +1379,573 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+			{
+			case Symbol::Text:
+				{
+					QString new_style = QString::fromLatin1("LABEL(f:\"")
+					                    .append(symbol->asText()->getFontFamily())
+					                    .append(QString::fromLatin1("\", s:"))
+					                    .append(QString::number(12))
+					                    .append(QString::fromLatin1(", t:{Name})"));
+					OGR_SM_AddStyle(manager.get(),
+					                QString::number(map->findSymbolIndex(symbol)).toLatin1(), new_style.toLatin1());
```
dg0yt: Number and new_style could be done with `QByteArray` directly. For efficiency, reserve() a reasonable size before appending. And I guess you could directly use `i` instead of `map->findSymbolIndex(symbol)`.
(Similar for the other symbol types below.)

dg0yt:
DXF seems to be a good choice for testing style export and import. (It is not yet on the GDAL 1 white list.) A first test with the "complete map" example indicates that
- we don't handle holes correctly.
- the order of layers/objects might need to reflect the order of colors (lowest first).

src/gdal/ogr_file_format.cpp
```diff
@@ -1405,5 +1379,573 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+			return QString::fromLatin1("PEN(c:#00000000)");
+
+		QString rgb = QColor(main_color->getRgb()).name();
+		return QString::fromLatin1("PEN(c:%1)").arg(rgb);
+	};
+
+	auto get_brush_style = [](const Symbol* symbol) {
+		auto main_color = symbol->guessDominantColor();
+		if (!main_color)
+			return QString::fromLatin1("BRUSH(fc:#00000000)");
```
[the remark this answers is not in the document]
xc-racer99: Depends on if it should default to black (#000000) or invisible (#00000000), but black probably does make more sense.

src/gdal/ogr_file_format.cpp
xc-racer99: Yes, you're correct.  I misinterpreted equals() and was thinking it was similar to QString::compare

src/gdal/ogr_file_format.cpp
```diff
@@ -1405,5 +1379,573 @@ LatLon OgrFileImport::calcAverageLatLon(OGRDataSourceH data_source)
...
+		addAreasToLayer(layer, is_area_object);
+	}
+	else if (option(QString::fromLatin1("Per Symbol Layers")).toBool())
+	{
+		// Determine symbols in use
+		std::vector<bool> symbols_in_use;
+		map->determineSymbolsInUse(symbols_in_use);
+
+		// Add points, lines, areas in this order for driver compatability (esp GPX)
+		for (auto i = map->getNumSymbols() - 1; i >= 0; --i)
```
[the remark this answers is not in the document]
xc-racer99: No, there's no particular reason to go backwards, it's just ever so slightly more efficient only calculating getNumSymbols() once.  Nothing to do with order of layers.

xc-racer99:
> DXF seems to be a good choice for testing style export and import. (It is not yet on the GDAL 1 white list.) A first test with the "complete map" example indicates that
> 
>     we don't handle holes correctly.
>     the order of layers/objects might need to reflect the order of colors (lowest first).
Thanks for the review, I've tried to accomodate your comments.  Additionally, this patch set changes to using the LIBKML driver when available for GDAL < 2.0, plus creates an enum for OGR driver quirks.  toUtf8() appears to be the correct QString function based on https://trac.osgeo.org/gdal/wiki/rfc30_utf8_filenames

Holes don't export properly with the DXF driver but they do with the LIBKML driver.  I'm assuming it's just a limitation of the GDAL DXF driver.

I've yet to handle order of colors, it might require another OGR driver quirk depending on whether on top should be first or last.  I'll try to investigate this over the coming days.

dg0yt:
> Holes don't export properly with the DXF driver but they do with the LIBKML driver. I'm assuming it's just a limitation of the GDAL DXF driver.

I was wrong about the holes: The complete map example has the north lines pattern covering the actual map. I think we can improve the style part later.

xc-racer99:
> I was wrong about the holes: The complete map example has the north lines pattern covering the actual map. I think we can improve the style part later.

Ok.

Also, I've left  commit "map: Add const version of applyOnMatchingObjects" in the PR as I don't see it in dev yet.  I'll happily drop it when I see it there.

dg0yt:
> Also,  I've left  commit "map: Add const version of applyOnMatchingObjects" in  the PR as I don't see it in dev yet.  I'll happily drop it when I see  it there.

https://github.com/OpenOrienteering/mapper/commit/95d53993a1698dfa586b95bfbc0c91838cfecf29 

Did you fetch the latest git dev?

xc-racer99:
> 95d5399
> 
> Did you fetch the latest git dev?

Ah, my bad, it is there.  I hadn't rebased onto the latest dev (was based off of #1120 previously) and had actually implemented the commit independently of yours, despite the very similar comments, and so was expecting to see it as the latest commit in dev.

dg0yt:
Merged and modified. Thank you very much.

PR #1071, opened 2018-03-08
Fix compile time warnings

lpechacek:
Reported by GCC 7.3.1.

### DIFF
--- a/src/core/symbols/symbol.h
+++ b/src/core/symbols/symbol.h
@@ -387,12 +387,12 @@ public:
 	/**
 	 * Returns the i-th component of the symbol number as int.
 	 */
-	int getNumberComponent(int i) const { Q_ASSERT(i >= 0 && i < number_components); return number[i]; }
+	int getNumberComponent(std::size_t i) const { Q_ASSERT(i < number_components); return number[i]; }
 	
 	/**
 	 * Sets the i-th component of the symbol number.
 	 */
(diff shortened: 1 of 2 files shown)


src/core/symbols/symbol.h
```diff
@@ -387,12 +387,12 @@ class Symbol
 	/**
 	 * Returns the i-th component of the symbol number as int.
 	 */
-	int getNumberComponent(int i) const { Q_ASSERT(i >= 0 && i < number_components); return number[i]; }
+	int getNumberComponent(std::size_t i) const { Q_ASSERT(i < number_components); return number[i]; }
```
dg0yt: This warning needs to be worked on, indeed. However, this modification changes the signature, moving the problem to the places where the function is used. We need to take the time to review these spots, too.

.. Or even better, review the implementation of number: Can it be exposed more directly? Does it deserve a type on its own? (Yes, it has invariants.)

dg0yt:
I cherry-pick the second commit (path_object_t), and replace the second one (symbol.h) by having explict type conversions in the body of the function.

src/core/symbols/symbol.h
lpechacek: Good point. Reminds me how I last tried to get signedness right in OO Mapper. Signature changes triggered domino effect and I didn't get very far with fixing. Type-cast approach is much less problematic.

PR #1078, opened 2018-03-15
LineSymbol: Implement alternative mid symbol placement

dg0yt:
So far, mid symbol were placed in the middle of the dash which
actually meant the middle of the dash group. This change turns the
mid symbol into a more flexible "style symbol" (without changing
the name) which alternatively can be placed in the middle of every
dash in dash groups, or at the gap between dash groups.


dg0yt:
This change will not only add flexibility to line symbols but also help with OCD import/export.
I consider to complement this by changing "mid symbol" to "style symbol" at least in the UI.

@petergolde This would add a new attribute to the file format. While not yet really documented, you see where I would add the documentation/changelog entry.

lpechacek:
Will review.

lpechacek:
Before I dive into the code details, I'd welcome an explanation why is the change needed. At the moment, I don't recall any ISOM symbol which would make use of this feature.

src/core/symbols/line_symbol.h
```diff
@@ -109,6 +109,17 @@ friend class OCAD8FileImport;
 		RoundJoin = 2
 	};
 	
+	/**
+	 * Mid symbol placement on dashed lines.
+	 */
+	enum MidSymbolPlacement
```
lpechacek: Early comments... If `MidSymbolPlacement` were a scoped enum, the below identifiers could go without the leading `CenterOf`. It's natural that MidSymbol is in the centre.

src/core/symbols/line_symbol.cpp
```diff
@@ -1127,6 +1131,31 @@ SplitPathCoord LineSymbol::createDashGroups(
...
+				{
+					if (position <= start.clen)
+						continue;
+					split = SplitPathCoord::at(position, split);
+					if (mid_symbol_rotatable)
+						orientation = split.tangentVector().angle();
+					mid_symbol->createRenderablesScaled(split.pos, orientation, output);
+				}
+			}
+
```
lpechacek: This line, and a few others, introduces whitespace error. While neither the coding style nor GitHub tools mandate trailing-whitespace free code, `git show --check origin/linesymbol-extended` shows them prominently.

dg0yt: Apart from how it is judged by `git show --check`, I fail to see this as an error, or even as *trailing* whitespace. Actually it is *leading* whitespace, consistent with the lines before and after. Git can match this only with blank lines having the same amount of whitespace, not with any blank line. I feel like that is what I want.

But I do not have a strong opinion about this. Qt seems to use strict blank lines, but I don't find an explicit rule about this (unless you interprete "blank line" in a very strict sense).

src/core/symbols/line_symbol.h
dg0yt: If "mid symbol" becomes "style symbol", we might need the `CenterOf.

lpechacek:
> This change will not only add flexibility to line symbols but also help with OCD import/export.
> I consider to complement this by changing "mid symbol" to "style symbol" at least in the UI.

I still don't understand motivation for this change. Can you be more verbose about how are the three different mid-symbol placement modes going to be used?

Another important point, from map export perspective, is how the placement modes map into the other map editor concepts. I'd be happy to learn how it is going to help in that area.

dg0yt:
> I still don't understand motivation for this change. Can you be more verbose about how are the three different mid-symbol placement modes going to be used?

This is an attempt to improve OCD import without massive changes in LineSymbol.  
CenterOfDash is what we have at the moment. In OCD files we have symbols at primary gaps and secondary gaps (where gaps may have zero width). CenterOfGap (matching the primary gap) and CenterOfDashgroup (matching the secondary gap) are going to be used like this.
https://github.com/OpenOrienteering/mapper/blob/19623e3674e9ce619a9e77cdbe2e376a630f6fe7/src/fileformats/ocd_file_import.cpp#L1310-L1329

> Another important point, from map export perspective, is how the placement modes map into the other map editor concepts.

I don't really understand what is meant here, "map editor concepts" "from map export perspective".

> > I consider to complement this by changing "mid symbol" to "style symbol" at least in the UI.

Note that mid symbols are already used with lines which are not dashed. "Mid symbol" has no meaning in this context. These are just repeated symbols dominating a line symbol's style. Hence the proposal to call them "style symbols".

src/core/symbols/line_symbol.cpp
lpechacek: Let's quit the holy war before it breaks out. I've instructed Git not to highlight trailing white space. And a blank line on my side of the wire matches regexp `^$`. :)

src/core/symbols/line_symbol.h
lpechacek: OK, no prob. The words `Dash` and `Gap` will anyway have to be dealt with as mid/style symbols apply on solid lines too.

src/fileformats/xml_file_format.h
```diff
@@ -91,6 +91,11 @@ class XMLFileFormat : public FileFormat
 
 \section changes Changes
 
+\subsection version-8 Version 8
+
+- 2018-03-10 Added `mid_symbol_placement` attribut to `line_symbol` element.
```
lpechacek: Spelling of word `attribute`.

dg0yt:
> However, the feature must be covered by a test before merging.

There is no example of such test, apart from what I did recently in `symbol_t`. This work was done partly in preparation of this change. (This change does not break the `line-symbol-mid-symbol-variants` test.) I probably lost another test file which covers the new features and will be a different commit anyway.

lpechacek:
> This is an attempt to improve OCD import without massive changes in LineSymbol.
CenterOfDash is what we have at the moment. In OCD files we have symbols at primary gaps and secondary gaps (where gaps may have zero width). CenterOfGap (matching the primary gap) and CenterOfDashgroup (matching the secondary gap) are going to be used like this.

Thanks! Now is the motivation clearer to me. However, the other program defines independent symbols for the two positions, while Mapper has only one symbol with three possible placements. To be sure about the plan, is the goal full compatibility or something else?

> I don't really understand what is meant here, "map editor concepts" "from map export perspective".

With map editor concepts I meant how the other program [defines line appearance and symbol placement](https://www.ocad.com/wiki/ocad12/en/index.php?title=Create_a_New_Line_Symbol). With map export I meant the process of translating Mapper map into the other representation. It is now clear that the various `CenterOf`* options correspond to different gap size and symbol placement combinations.

Attaching example file I used in my testing: [twosymbols.zip](https://github.com/OpenOrienteering/mapper/files/1823146/twosymbols.zip)
twosymbols

lpechacek:
> There is no example of such test, apart from what I did recently in `symbol_t`. This work was done partly in preparation of this change. (This change does not break the `line-symbol-mid-symbol-variants` test.) I probably lost another test file which covers the new features and will be a different commit anyway.

I had exactly `symbol_t` in mind. It should be extended with the new symbol placement combinations.

dg0yt:
Citing myself: "This is an attempt to improve OCD import without massive changes in LineSymbol." It is not about full compatibility. And it is not about 2020. It is about improving situation for common orienteering map symbol sets, now. And they don't seem to use both primary and secondary symbols.

(If I were to design a line symbol from scratch, it would be very different from Mapper's current line symbol as well as from OCD format. But there is no point in doing so unless we implement course setting.)

lpechacek:
> Citing myself: "This is an attempt to improve OCD import without massive changes in LineSymbol." It is not about full compatibility. And it is not about 2020. It is about improving situation for common orienteering map symbol sets, now. And they don't seem to use both primary and secondary symbols.

Understood. If you have an example of a symbol set that uses secondary symbol, please share it with me. I'd like to have a closer look. Thanks!

dg0yt:
Updated.

dg0yt:
Merge to branch ocd-export, #1032.

PR #1081, opened 2018-03-23
Respect user-selected file format when saving

dg0yt:
This PR adds missing pieces, cleans up the API, and applies to ocd-export.

### DIFF
--- a/src/gui/main_window.cpp
+++ b/src/gui/main_window.cpp
@@ -985,7 +985,7 @@ bool MainWindow::save()
 	return savePath(currentPath());
 }
 
-bool MainWindow::savePath(const QString &path)
+bool MainWindow::savePath(const QString &path, const FileFormat* format)
 {
 	if (!controller)
 		return false;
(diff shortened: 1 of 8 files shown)


src/gui/main_window.cpp
```diff
@@ -1132,7 +1143,7 @@ bool MainWindow::showSaveAsDialog()
 	// Fails when using different formats for import and export:
 	//	Q_ASSERT(FileFormats.findFormatForFilename(path) == format);
 	
-	return savePath(path);
+	return savePath(path, format);
```
dg0yt: Hint: This is the root of the changes: We have got the format from the file dialog's *filter* and must pass it down.

src/fileformats/ocd_file_format.cpp
```diff
@@ -45,6 +72,16 @@ OcdFileFormat::OcdFileFormat()
 	// Nothing
 }
 
+OcdFileFormat::OcdFileFormat(quint16 version)
+: FileFormat { MapFile, idForVersion(version),
```
lpechacek: Both these entries and "OCAD" from default constructor (line 69) are presented in the Save as... dialog as possible file types. Only code inspection reveals that "OCAD" means "save with v9 or the version from which the map was loaded". A more descriptive text is needed.

src/fileformats/ocd_file_format.cpp
```diff
@@ -45,6 +72,16 @@ OcdFileFormat::OcdFileFormat()
...
+               version == 1 ? ::OpenOrienteering::ImportExport::tr("OCAD legacy export")
```
lpechacek: Legacy export should mention the version number. E.g. "OCAD legacy exporter (v8)". Not every user knows...

src/fileformats/ocd_file_format.cpp
dg0yt: Proposal: "OCAD (auto-selected version)"

Details might go to a manual page on file formats (or OCD format support).

src/fileformats/ocd_file_format.cpp
dg0yt: Proposal: "OCAD version 8 (old exporter)". 

Anyway, we should remove the legacy exporter as soon as possible. On Android, we don't have this dialog at all.

dg0yt:
I will merge all but the OCD-related commit directly to `dev` so that it becomes available for testing and for PR #1068.

lpechacek:
The proposals look good to me. Go ahead!

dg0yt:
OCD-related commit will go to ocd-export branch.

PR #1089, opened 2018-04-16
OCD file import checks

lpechacek:
Basic sanity checks for new OCD file import.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -352,8 +352,11 @@ void OcdFileImport::importImplementation(bool load_symbols_only)
 void OcdFileImport::importGeoreferencing(const OcdFile<Ocd::FormatV8>& file)
 {
 	const Ocd::FileHeaderV8* header = file.header();
-	const Ocd::SetupV8* setup = reinterpret_cast< const Ocd::SetupV8* >(file.byteArray().data() + header->setup_pos);
+	const Ocd::SetupV8* setup = Ocd::getPointerChecked< Ocd::SetupV8 >(file.byteArray(), header->setup_pos);
 	
+	if (Q_UNLIKELY(!setup))
+		return;
+
 	Georeferencing georef;
 	georef.setScaleDenominator(qRound(setup->map_scale));
 	georef.setProjectedRefPoint(QPointF(setup->real_offset_x, setup->real_offset_y));
@@ -844,6 +847,9 @@ void OcdFileImport::importSymbols(const OcdFile< F >& file)
 {
 	for (auto ocd_symbol_entry : file.symbols())
 	{
+		if (Q_UNLIKELY(!ocd_symbol_entry.entity))
+			continue; // skip invalid entries
+
 		auto& ocd_symbol = *ocd_symbol_entry.entity;
 		
 		// When extra symbols are created, we want to insert the main symbol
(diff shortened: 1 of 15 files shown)


src/fileformats/ocd_file_import.cpp
```diff
@@ -844,6 +847,9 @@ void OcdFileImport::importSymbols(const OcdFile< F >& file)
 {
 	for (auto ocd_symbol_entry : file.symbols())
 	{
+		if (Q_UNLIKELY(!ocd_symbol_entry.entity))
+			continue; // skip invalid entries
```
dg0yt: Cf. (1)

src/fileformats/ocd_file_import.cpp
```diff
@@ -933,7 +939,8 @@ void OcdFileImport::importObjects(const OcdFile< F >& file)
 	{
 		if ( ocd_object.entry->symbol
 		     && ocd_object.entry->status != Ocd::ObjectDeleted
-		     && ocd_object.entry->status != Ocd::ObjectDeletedForUndo )
+		     && ocd_object.entry->status != Ocd::ObjectDeletedForUndo
+		     && ocd_object.entity )
```
dg0yt: Cf. (1)

src/fileformats/ocd_types.h
```diff
@@ -683,7 +704,8 @@ OcdEntityIndexIterator<V> OcdEntityIndexIterator<V>::operator++(int)
 template< class V >
 typename OcdEntityIndexIterator<V>::value_type OcdEntityIndexIterator<V>::operator*() const
 {
-	return { &block->entries[index], reinterpret_cast<const typename value_type::EntityType*>(byte_array->data()+block->entries[index].pos) };
+	auto entry_ptr = Ocd::getPointerChecked< const typename value_type::EntityType >(*byte_array, block->entries[index].pos);
+	return { &block->entries[index], entry_ptr };
```
dg0yt: (1) This validation is performed each time the iterator is dereferenced, and it requires checking the result in calling code. However I intended to make non-dereferencable entries unreachable via iterators, i.e. ATM you cannot reach entries with `pos` zero.

https://github.com/lpechacek/mapper/blob/ebcc8837f596d5b50e7e06bc52e4b048030e0b57/src/fileformats/ocd_types.h#L668-L694

So I wonder if it would be possible to move the pointer validation to `OcdEntityIndexIterator` constructor and `operator++`, thus making entries with invalid pointer unreachable. However I can't tell now if this would work for strings, symbols and objects.

src/fileformats/ocd_types.h
```diff
@@ -308,6 +308,27 @@ namespace Ocd
...
+		if (Q_UNLIKELY(offset + sizeof(StructureType) > byte_array.size()))
+		{
+			qWarning("Attempt to access misplaced file structure (%s@%d)",
+			         typeid(StructureType).name(), offset);
+			return nullptr;
+		}
+
+		return reinterpret_cast< StructureType* >
+		        (const_cast< char* >(byte_array.data()) + offset);
+	}
```
dg0yt: This is almost identical with what is already implemented by `Ocd::getBlockChecked`.

(`Ocd::getBlockChecked` returns `nullptr` also when `offset` is zero.)

dg0yt:
Conceptually transfered to #1032 ef4d93fcf1bc7e28253f8a65defa58fc085bc2be, 20a3652d15413d759a771c3945af86a6804527c9, 67cc1a468aeabcf3d2224ba6cd44f0d55392b33b.

PR #1108, opened 2018-06-26
images: Add image for 'missing template'

lpechacek:
First shot at the 'missing template' issue:
image

Enhancements welcome.


dg0yt:
Funny, but probably to complex for the intented use. It needs five times the space or resolution of a single cross.

lpechacek:
Yes, this was primarily intended to make a bit of fun. Of course, the most practical course of action is to resurrect the red cross image from GIT history.

PR #1118, opened 2018-07-18
transform_t: Avoid hitting a Qt bug

lpechacek:
This is the minimal change to skip tests failing with Qt 5.11.1. Use of QEXPECT_FAIL would require modification at four places or definition of a new inline function. Given the importance of this issue I opted for minimal code footprint.

### DIFF
--- a/test/transform_t.cpp
+++ b/test/transform_t.cpp
@@ -365,10 +365,19 @@ void TransformTest::testEstimateSimilarityTransformation()
 		QCOMPARE(t2.template_scale_x, 1.0);
 		QCOMPARE(t2.template_scale_y, 1.0);
 		
+// Qt 5.11.1 has a regression in QPointF::operator==
+// References:
+// https://bugreports.qt.io/browse/QTBUG-69368
+// https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903237
+// https://github.com/OpenOrienteering/mapper/issues/1116
+#if QT_VERSION != 0x050B01
 		QCOMPARE(QPointF(passpoints[0].calculated_coords), QPointF(passpoints[0].dest_coords));
 		QVERIFY(passpoints[0].error < min_distance);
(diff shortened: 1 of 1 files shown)


dg0yt:
Without digging for the commit message, it won't be possible to see which QTBUG motivated these lines. How about defining and using a more telling macro, e.g.

~~~
#if QT_VERSION == 0x050B01
// https://bugreports.qt.io/browse/QTBUG-69368
#  define QTBUG_69368_QUIRK
#endif
...
#ifndef QTBUG_69368_QUIRK
QCOMPARE...
#endif
~~~

This would also make it easy to identify which liness can be removed once Qt 5.11.1 is ancient history.

lpechacek:
No objection per se. I'd, however, like to avoid spending too much time and LOCs on a patch that will be reverted in a few years latest. This **is a temporary code change** and very likely the buggy Qt release will not end up in any of Mapper supported OS-es.

That said, I offer:

1) Rewording the warning to include the bug reference: "Parts of test skipped to avoid hitting QTBUG-69368"
2) Rewording warning and adding an explanatory comment including all references before the first `#if`.
3) Implementing the proposed `#define`ry.
4) Going with `QEXPECT_FAIL` and `#define`s and create `QCOMPARE_QTBUG_69368`
```
#if QT_VERSION == 0x050B01
// https://bugreports.qt.io/browse/QTBUG-69368
#define QCOMPARE_QTBUG_69368(actual, expected) \
do {\
    QEXPECT_FAIL("", "QTBUG-69368", Continue); \
    QCOMPARE(actual, expected); \
} while (false)
#else
#define QCOMPARE_QTBUG_69368 QCOMPARE
#endif
```

In retrospect, 1) should have been done before submission.

dg0yt:
To be honest, I would rather avoid the complex macro.

lpechacek:
Me too.

PR #1120, opened 2018-08-02
Review of file import/export

dg0yt:
- Export is now operating on const Map and MapView.
- The FileFormat of edited files is identified early, passed on and remembered for saving. This should help with pairing generic importers with specialized exporters, e.g. for OCD versions.
- Importers and Exporters may now work on file paths directly, or continue to use QIODevice. This should help with GDAL.

### DIFF
--- a/src/fileformats/file_format.h
+++ b/src/fileformats/file_format.h
@@ -31,8 +31,6 @@
 #include <QString>
 #include <QStringList>
 
-class QIODevice;
-
 namespace OpenOrienteering {
 
 class Exporter;
@@ -243,7 +241,7 @@ public:
 	 * If the Importer can not be created, a FileFormatException shall be
 	 * thrown. The default implementation does just that.
 	 */
-	virtual std::unique_ptr<Importer> makeImporter(QIODevice* stream, Map *map, MapView *view) const;
+	virtual std::unique_ptr<Importer> makeImporter(const QString& path, Map *map, MapView *view) const;
 	
 	/** 
 	 * Creates an Exporter that will save a map to the given path.
--- a/src/fileformats/file_format.h
+++ b/src/fileformats/file_format.h
@@ -251,7 +251,7 @@ public:
 	 * If the Exporter can not be created, a FileFormatException shall be
 	 * thrown. The default implementation does just that.
 	 */
-	virtual std::unique_ptr<Exporter> makeExporter(QIODevice* stream, Map *map, MapView *view) const;
+	virtual std::unique_ptr<Exporter> makeExporter(QIODevice* stream, const Map* map, const MapView* view) const;
 	
 private:
 	FileType file_type;
--- a/src/fileformats/file_format.h
+++ b/src/fileformats/file_format.h
@@ -246,12 +246,12 @@ public:
 	virtual std::unique_ptr<Importer> makeImporter(QIODevice* stream, Map *map, MapView *view) const;
 	
 	/** 
-	 * Creates an Exporter that will save A map into the given stream.
+	 * Creates an Exporter that will save a map to the given path.
 	 *
 	 * If the Exporter can not be created, a FileFormatException shall be
 	 * thrown. The default implementation does just that.
 	 */
-	virtual std::unique_ptr<Exporter> makeExporter(QIODevice* stream, const Map* map, const MapView* view) const;
+	virtual std::unique_ptr<Exporter> makeExporter(const QString& path, const Map* map, const MapView* view) const;
 	
 private:
 	FileType file_type;
--- a/src/fileformats/file_format_registry.cpp
+++ b/src/fileformats/file_format_registry.cpp
@@ -69,6 +69,31 @@ std::unique_ptr<FileFormat> FileFormatRegistry::unregisterFormat(const FileForma
 }
 
 
+std::unique_ptr<Importer> FileFormatRegistry::makeImporter(const QString& path, Map& map, MapView* view)
+{
+	auto extension = QFileInfo(path).suffix();
+	auto format = findFormat([extension](auto format) {
+		return format->supportsImport()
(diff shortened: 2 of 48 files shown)


dg0yt:
This is the set of changes which precede the latest update to the ocd-export PR #1032.

dg0yt:
(No thorough review requested.)

lpechacek:
My analytical capacity is currently completely consumed by job tasks. Even though I'll attempt to complete the review until end of week.

dg0yt:
Thanks for letting me know. Yesterday I already merged this, the OCD export, and OCD georef PR to dev, but it still makes sense to have a look at this separately. (I tried hard to not throw everything in a single big commit.) Following up, I managed to remove the last GUI (widgets) code from the `Map` class, and the contributed OGR export PR was updated to use the file path directly instead of complicated dealing with QIODevice.

src/gdal/ogr_template.cpp
```diff
@@ -70,7 +70,8 @@ namespace {
 		tmp_map.setGeoreferencing(initial_georef);
 		OgrFileImport importer{ &file, &tmp_map, nullptr, OgrFileImport::UnitOnGround};
 		importer.setGeoreferencingImportEnabled(true);
-		importer.doImport(true);
+		importer.setLoadSymbolsOnly(true);
+		importer.doImport();
```
lpechacek: It is unclear to me why is the symbol import functionality controlled by an object property. I wonder if separate `doSymbolImport()` did a better job.

src/fileformats/file_format.h
```diff
@@ -243,7 +241,7 @@ class FileFormat
 	 * If the Importer can not be created, a FileFormatException shall be
 	 * thrown. The default implementation does just that.
 	 */
-	virtual std::unique_ptr<Importer> makeImporter(QIODevice* stream, Map *map, MapView *view) const;
+	virtual std::unique_ptr<Importer> makeImporter(const QString& path, Map *map, MapView *view) const;
```
lpechacek: What is the semantics for both `map` and `view` being `nullptr`?

src/fileformats/file_format_registry.cpp
```diff
@@ -97,5 +98,35 @@ const FileFormat *FileFormatRegistry::findFormatForFilename(const QString& filen
 	});
 }
 
+const FileFormat* FileFormatRegistry::findFormatForData(const QString& path, FileFormat::FileTypes types) const
+{
+	QFile file(path);
+	if (!file.open(QIODevice::ReadOnly))
+		return nullptr;
+	
+	char buffer[256];
```
lpechacek: Suggest initializing the buffer to zeros (`char buffer[256] = {}`). Number of bytes to be read from the input stream is unknown as well as behavior of underlying `FileFormat::understands()` implementations.

src/fileformats/file_format_registry.cpp
```diff
@@ -69,6 +69,31 @@ std::unique_ptr<FileFormat> FileFormatRegistry::unregisterFormat(const FileForma
 }
 
 
+std::unique_ptr<Importer> FileFormatRegistry::makeImporter(const QString& path, Map& map, MapView* view)
```
lpechacek: `makeImporter()` signature differs from `FileFormat::makeImporter()` as well as neighboring `makeExporter()`. Suggest either explaining the difference or aligning interfaces.

src/fileformats/ocd_file_import.cpp
```diff
@@ -2087,9 +2087,9 @@ void OcdFileImport::import(bool load_symbols_only)
 	Q_ASSERT(buffer.isEmpty());
 	
 	buffer.clear();
-	buffer.append(stream->readAll());
+	buffer.append(device()->readAll());
```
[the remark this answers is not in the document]
dg0yt: I have no experience with mmap. It may work here and now. But for mmap, code will be more complex: We will need to make sure to operate on QFileDevice (for using its map() function), and we will need to make sure to unmap() even when exceptions are thrown. OTOH I wonder if could make sense to retain the whole file in memory, for reuse when writing (to preserve original data). The map files are fairly small in contrast to most raster templates, so this is not too expensive.

However, even while `buffer.append' will avoid unneccessary copying for an empty QByteArea, a simple assignment should be used here.

src/fileformats/file_format.h
dg0yt: None. I didn't invent it, I just started with the API of the Importer/Exporter classes. 
FileFormat and friends will need substantial review until they really can handle more then Map.

src/fileformats/file_format_registry.cpp
dg0yt: But the number of bytes read is known after the next line, and it is passed to `FileFormat::understands()`:
~~~
auto total_read = int(file.read(...));
....
switch (format->understands(buffer, total_read))
...
~~~

src/fileformats/file_import_export.h
```diff
@@ -64,6 +65,31 @@ class ImportExport
...
+	 * might wish to use other ways of accessing the output/input path.
+	 * In these cases, the default implementation should be overwritten to
+	 * return `false`.
+	 */
+	virtual bool supportsQIODevice() const noexcept;
+	
+	/**
+	 * Returns the input device if it has been set or created.
+	 */
+	QIODevice* device() const noexcept { return device_; }
```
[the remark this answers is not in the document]
dg0yt: IIUC C.131 is not about a single setter/getter pair but about the design of a type as a whole.
IMO `Importer` does not classify as "a `struct` – that is, a behaviorless bunch of variables, all public data and no member functions" (C.131). It is polymorphic. Its key feature is the virtual member function `doImport()`, and so the destructor is virtual, too. You cannot use default assignments or similar. And I don't want to couple inheriting classes to a particular member variable.
(I'm only touching `device` in this change. `map` and `view` are a different story.)

PR #1128, opened 2018-08-21
WIP: Something is wrong about the codes and descriptions

comradekingu:


comradekingu:
Symbol 106 is "Ruined earth wall", not "Small earth wall".
According to https://orienteering.org/wp-content/uploads/2017/04/ISOM2017_20May2017.pdf#page=17

This throws the numbering off, as it is listed as 108 some places, and I don't really know the difference between 2000, 10000, etc. It seems to be resolution, and http://openorienteering.org/apps/mapper/xml/v2 is dead.

The description is also not correct:

"A small or partly ruined earth wall shall be shown with a dashed line. Minimum height is 0.5 m." 
↓
"A ruined or less distinct earth wall. Minimum height: 0.5 m."

Spotted by yevhenmazur @ Weblate

https://hosted.weblate.org/translate/openorienteering/map-symbols/uk/?checksum=aae04da025f83787

https://hosted.weblate.org/translate/openorienteering/map-symbols/uk/?checksum=ef9eece4fad48661

Additionally the desc for 304 is wrong

https://orienteering.org/wp-content/uploads/2017/04/ISOM2017_20May2017.pdf#page=23

"A crossable watercourse, minimum 2 m wide. The width of watercourses over 5 m wide should be shown to scale."
↓
"Should be at least 2 m wide."

https://hosted.weblate.org/translate/openorienteering/map-symbols/uk/?checksum=c9529c0346e3fb4

dg0yt:
"108 Small earth wall" is correct for ISOM 2000. It seems it was copied without updating ISOM 2017 symbol 106.

I will pick this up. This needs to be 
1. changed in `symbol sets/src/ISOM2000_15000.xmap`, 
2. propagated to the scaled symbol sets and translations with `symbol_set_t`, 
3. propagated to `examples/src/*.omap` if they use ISOM 2017 at all.

dg0yt:
Examples don't use ISOM2000 (at least in `master`).
For future pull requests: just edit files in `symbol sets/src/`. There's a number of more map symbol comments on weblate which need to be handled. What we cannot use at the moment is sizes in mm, because the symbol sets can be used with different scales.

Issue #866, opened 2017-05-06
Editable symbol icons


zlenhart:
Automaticaly generated symbol icons bring several problems (see Issue #848), look as being same for similar symbols (even for very different symbols, e.g. 101.0 contour – 106.2 earth bank tag). Some users may need and create tiny working lines and point signs for survey what gives almost unvisible icons. In combination with absence of textual specification in mobile version it is very inconvenient. Simple icon editor would help to modify some of icons to avoid confusion.

### Actual behaviour
Same or very similar icons for different symbols, difficult to understand what is what in mobile version.

### Expected behaviour
Simple icon editor (desktop version only) for additional changes of icons. Editor in Ocad may serve as inspiration. If this is to complex, addition of some colour dots and/or stripes may  be sufficient. The creation of icons may remain as it is now.

### Configuration
Mapper Version: 0.6.6.
Operating System: Android 6.0.1

ollesmaps:
I quite like the simplicity of creating icons. You can differentiate icons by their positioning in Symbol bar.

lpechacek:
A quick comment on generated icons. Mapper works on devices with various [PPI](https://en.wikipedia.org/wiki/Pixel_density) (pixels per inch) values. Typical desktop computer display has about 100 PPI, laptop displays are anywhere from 100 PPI to 300 PPI. Handheld devices range from 200 PPI up to 600 PPI. Mapper runs on all of these and as the symbol icon size is defined in millimeters, as opposed to pixels, the generated icon is ideally fit to display resolution.

When Mapper adopts fixed size icon, depending on the chosen size, they will either look coarse on high resolution displays, or fail to downscale on low resolution ones.

Mapbeek:
There are two issues discussed under this topic heading, the method of creating and editing icons, and identifying icons on different devices when using Mapper.

As a user I would like to be able to 1 ) easily create and edit icons, as can be done in OCAD using the same drawing tools used for general map drafting, and 2) easily identify icons when mapping in the field.

At present both are challenging in Mapper.  Editing icons is too hard, and identifying the tiny icons on a mobile device is practically impossible for a lot of similar symbols,  with the result that the wrong symbols get selected, leading to extra work and possibly errors when mistakes are not found soon enough and are not corrected.

[...]

PR #1129, opened 2018-08-23
Custom icon support

dg0yt:


dg0yt:
OCD icons (22 pixels, 256 colors, scaled up) versus generated icons (optimal quality):  
ocd-versus-generated

lpechacek:
First thoughts when looking at the patch set. `Symbol: Add custom icon functions` adds custom icon which replaces the generated icon. I have given a thought to the topic recently and came up with the idea of "overlay" icon, leveraging the image alpha channel. That way either the whole generated icon could be replaced, or just a part of it (my understanding of issue #866). However, any user edited bitmap image will fail to match the dynamically  generated symbol icons. Icon overlaying/composition still may make sense. Thinking on...

dg0yt:
icon-settings

lpechacek:
The combination of custom icon for OCD and OMAP export is troublesome. _...bla bla bla deleted..._ Anyway, let's go with 22px true color icons and forget about overlaying. Although this is not "perfect", it will be understandable for end users.

BTW, I'll surely be one of the early adopters. I frequently have difficulty distinguishing between contour slope line vs form line slope line and shallow body of water border line vs minor water channel.

dg0yt:
Done. 
(Should be a good starting point for adding the overlay behaviour.)

src/fileformats/ocd_file_import.cpp
```diff
@@ -1771,6 +1833,188 @@ void OcdFileImport::setupPointSymbolPattern(PointSymbol* symbol, std::size_t dat
...
+	for (int y = icon_size - 1; y >= 0; --y)
+	{
+		for (int x = 0; x < icon_size; x += 2)
+		{
+			image.setPixelColor(x, y, palette[(*icon_bits) >> 4]);
+			image.setPixelColor(x+1, y, palette[*(icon_bits++) & 0xf]);
+		}
+		icon_bits++;
+	}
+	return image;
```
lpechacek: Just dreaming....
```
QImage image(icon.bits, icon_size, icon_size, QImage::Format_Indexed8);
image.setColorTable(palette);
return mirrored(true, false).convertToFormat(QImage::Format_ARGB32_Premultiplied);
```
would be my starting point for implementation of the above. It of course assumes that palette is a `QVector`.

src/fileformats/ocd_file_import.h
```diff
@@ -46,8 +46,15 @@
 #include "fileformats/ocd_types_v8.h" // IWYU pragma: keep
 
 class QChar;
+class QImage;
 class QTextCodec;
 
+namespace Ocd {
+
+struct IconV9;
```
lpechacek: How come that only `IconV9` has to be forward-declarated?

src/fileformats/ocd_file_import.cpp
```diff
@@ -1771,6 +1833,188 @@ void OcdFileImport::setupPointSymbolPattern(PointSymbol* symbol, std::size_t dat
...
+	};
+	
+	auto icon_bits = icon.bits;
+	constexpr int icon_size = 22;
+	QImage image(icon_size, icon_size, QImage::Format_ARGB32_Premultiplied);
+	for (int y = icon_size - 1; y >= 0; --y)
+	{
+		for (int x = 0; x < icon_size; ++x)
+		{
+			image.setPixelColor(x, y, palette[*(icon_bits++)]);
```
lpechacek: `icon_bits` is `quint8` and palette size is 125. Suggest range checking/clipping here.

src/fileformats/ocd_file_import.cpp
```diff
@@ -1076,6 +1085,59 @@ void OcdFileImport::setupBaseSymbol(Symbol* symbol, const OcdBaseSymbol& ocd_bas
...
+	}
+	else
+	{
+		symbol->setCustomIcon(importSymbolIcon(ocd_base_symbol.icon));
+	}
+}
+
+
+template<class OcdBaseSymbol>
+void OcdFileImport::checkSymbolIcon(Symbol* symbol, const OcdBaseSymbol& ocd_base_symbol)
```
lpechacek: `checkSymbolIcon()` does not express what the function is doing. `tryIconReplace()` might be a better fit.

src/core/symbols/symbol.h
```diff
@@ -333,9 +333,11 @@ class Symbol
 	/**
 	 * Sets a custom symbol icon.
 	 * 
-	 * The custom icon takes precedence over the generated one. Like the
-	 * generated icon, it is not part of the symbol state which is compared by
-	 * the `equals` functions. However, it is copied when duplicating an icon.
+	 * The custom icon takes precedence over the generated one when displaying
+	 * custom icons is enabled.
```
lpechacek: Language. "... generated one when _custom icon display is enabled._"

src/fileformats/ocd_file_export.cpp
```diff
@@ -23,6 +23,7 @@
 #include "ocd_file_export.h"
```
lpechacek: If there were a way to comment on the patch as whole...

This patch should be squashed into the previous one. See also additional comments below.

src/fileformats/ocd_file_export.cpp
```diff
@@ -2384,7 +2210,7 @@ QByteArray OcdFileExport::exportCombinedLineSymbol(
 void OcdFileExport::exportSymbolIcon(const Map* map, const Symbol* symbol, Ocd::IconV8& icon)
 {
 	// Icon: 22x22 with 4 bit palette color, origin at bottom left
-	constexpr int icon_size = 22;
+	constexpr int icon_size = Ocd::IconV8::size();
```
lpechacek: Naming. When I see `size()` method on an image-like object, I expect it to return `QSize`. `width()` would tell me better that a single value is being returned.

src/fileformats/ocd_file_import.cpp
```diff
@@ -1836,28 +1846,11 @@ void OcdFileImport::setupPointSymbolPattern(PointSymbol* symbol, std::size_t dat
...
-	    {   0, 255, 255 },
-	    { 255, 255, 255 }
-	};
+	static const auto palette = Ocd::IconV8::palette<QColor>();
+	constexpr int icon_size = Ocd::IconV8::size();
 	
 	auto icon_bits = icon.bits;
-	constexpr int icon_size = 22;
-	QImage image(icon_size, icon_size, QImage::Format_ARGB32_Premultiplied);
+	auto image = QImage{icon_size, icon_size, QImage::Format_ARGB32_Premultiplied};
```
lpechacek: `QImage image {icon_size, icon_size, QImage::Format_ARGB32_Premultiplied};` is the way to go (ES.23).

src/gui/symbols/area_symbol_settings.cpp
```diff
@@ -267,6 +267,8 @@ AreaSymbolSettings::AreaSymbolSettings(AreaSymbol* symbol, SymbolSettingDialog*
 		emit propertiesModified();
```
lpechacek: I don't see how this patch fits into the Custom Icon series. Should it live in a different branch?

src/gui/symbols/icon_properties_widget.cpp
```diff
@@ -0,0 +1,212 @@
...
+	auto layout = new QFormLayout();
+	setLayout(layout);
+	
+	layout->addRow(Util::Headline::create(tr("Default icon")));
+	
+	default_icon_display = new QLabel();
+	layout->addRow(default_icon_display);
+	
+	default_icon_size_edit = Util::SpinBox::create(22, 256, tr("px"));
+	layout->addRow(tr("Width:"), default_icon_size_edit);
```
lpechacek: What is the purpose of the icon size change spin box?

src/gui/symbols/icon_properties_widget.cpp
```diff
@@ -0,0 +1,212 @@
...
+	custom_icon_size_edit->setDisabled(true);
+	layout->addRow(tr("Width:"), custom_icon_size_edit);
+	
+	layout->addItem(Util::SpacerItem::create(this));
+	
+	
+	auto buttons_layer = new QHBoxLayout();
+	export_button = new QPushButton(tr("Export..."));
+	buttons_layer->addWidget(export_button);
+	import_button = new QPushButton(tr("Import..."));
```
lpechacek: Suggest using traditional Save/Load instead of Export/Import.

src/core/symbols/symbol.cpp
```diff
@@ -261,6 +264,26 @@ void Symbol::save(QXmlStreamWriter& xml, const Map& map) const
 	if (!description.isEmpty())
 		xml.writeTextElement(QLatin1String("description"), description);
 	saveImpl(xml, map);
+	if (!custom_icon.isNull())
+	{
+		QBuffer buffer;
+		QImageWriter writer{&buffer, QByteArrayLiteral("PNG")};
+		if (writer.write(custom_icon))
```
lpechacek: AFAICT, `QImage::save(...)` will do. There are no special `QImageWiter` settings being applied.

src/core/symbols/symbol.cpp
```diff
@@ -307,7 +330,38 @@ std::unique_ptr<Symbol> Symbol::load(QXmlStreamReader& xml, const Map& map, Symb
...
+					data = QByteArray::fromBase64(data);
+			}
+			else
+			{
+				data.clear();  // Ignore unknown data, warning is generated later.
+			}
+			
+			QBuffer buffer{&data};
+			QImageReader reader{&buffer, QByteArrayLiteral("PNG")};
+			auto icon = reader.read();
```
lpechacek: Propose `QImage::load(...)`. E.g. 
```
QImage icon;
icon.load(buffer, QByteArrayLiteral("PNG"));
```

src/fileformats/ocd_file_export.cpp
```diff
@@ -47,6 +47,7 @@
 #include <QPointF>
```
lpechacek: Patch comment should mention that v8 compressed icons are not yet implemented.

src/fileformats/ocd_file_import.cpp
lpechacek: Hmm, this of course applies to v9 icon format. There is nothing like `QImage::Format_Indexed4` which could be leveraged here.

src/core/symbols/symbol.cpp
dg0yt: `QImage::save/load` offer no way to determine what kind of error occured in the process. 
(In the given function I could only throw an error at the moment. I chose to print a debug message instead.)

src/fileformats/ocd_file_export.cpp
dg0yt: Probably you mean the commit which introduces the icon types?

src/fileformats/ocd_file_import.cpp
dg0yt: How about `dropRedundantIcon()`?

src/fileformats/ocd_file_import.cpp
dg0yt: Well spotted.

src/fileformats/ocd_file_import.h
dg0yt: Because the `ocd_types_v8.h` header is already included, line 46.

src/gui/symbols/area_symbol_settings.cpp
dg0yt: I could apply this commit separately, but at the moment it is needed only for the following commit which adds another tab.

src/gui/symbols/icon_properties_widget.cpp
dg0yt: For now, that's the only way to change the size of the default icon when you want to save it for customization. We can use a [QInputDialog](http://doc.qt.io/qt-5/qinputdialog.html#details) instead. This change will create sequence of two dialogs  for saving the default icon (1. file dialog, 2. size dialog).

PS: Should there be an extra Save button for the default icon?

src/fileformats/ocd_file_export.cpp
dg0yt: Okay, 'width' is also what I use in the GUI. 
(I want to avoid adding more Qt types in the Ocd namespace. Most types can be easily replaced with C++11 ones.)

src/fileformats/ocd_file_import.cpp
lpechacek: `dropRedundantIcon()` sounds good to me.

src/fileformats/ocd_file_export.h
```diff
@@ -249,11 +249,13 @@ class OcdFileExport : public Exporter
 	        const LineSymbol* double_line );
 	
 	
-	void exportSymbolIcon(const Symbol* symbol, Ocd::IconV8& icon);
+public:
+	static void exportSymbolIcon(const Map* map, const Symbol* symbol, Ocd::IconV8& icon);
 	
-	void exportSymbolIcon(const Symbol* symbol, Ocd::IconV9& icon);
+	static void exportSymbolIcon(const Map* map, const Symbol* symbol, Ocd::IconV9& icon);
```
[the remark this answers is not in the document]
lpechacek: Think so too.

src/fileformats/ocd_file_export.cpp
lpechacek: `a4a96070002e5 Ocd: Move icon palette and size to format headers`

I commented on individual patches. Switching to "Commits" and looking at the patches one by one will put the comments into context. For the next time I'll try to make my commentary compatible with the "merged code change" view.

src/gui/symbols/area_symbol_settings.cpp
lpechacek: I see. Please consider mentioninig in patch comment that this is pre-requesite for the next change.

src/gui/symbols/icon_properties_widget.cpp
lpechacek: The spinbox is fine from my POV. It is the intended workflow which is unclear to me at the moment.

Regarding buttons, my first thought was having "Populate custom icon from generated one". That way all operations can be restricted to the custom icon. In the end I also think that size of both icons can be locked and controlled by only one spinbox.

Summary of my proposal:
1) display both generated and custom icon in the dialog
2) provide possibility to load and save custom icon for external modifications
3) provide possibility to copy generated icon into custom one (for later editing)
4) control icon size by only one spinbox, rescale custom icon upon size changes
5) future - provide integrated icon editor

Remark 1 - Custom icons get (down)scaled and color-reduced when exported into OCD.
Remark 2 - Vector based custom icons might be worth a shot. Graphic primitives can be limited to line and Bézier curve + line width and color settings.

src/fileformats/ocd_file_export.cpp
lpechacek: I'm sure we are on the same page wrt usage of Qt types in generic code. My remark was along the lines that when I, and maybe other programmers too, read code, it is easier to work with one concept, rather than switching between multiple ones. `size()` on two dimensional object immediately translates to width x height for me. The point is that having to remember that `Ocd::IconV*`'s `size()` is a scalar is an extra thing to keep in mind.

src/fileformats/ocd_file_export.h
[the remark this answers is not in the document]
dg0yt: Returning by value instead of passing an output parameter makes it impossible to simply overload the function name for different icon types. We will either have to use templates, or we may use an extra type, moving the export implementation to conversion operators:

~~~
struct OcdIcon
{
	const Map& map;
	const Symbol& symbol;
	
	operator Ocd::IconV8() const;
	operator Ocd::IconV9() const;
};

...

ocd_base_symbol.icon = OcdIcon{map, symbol};
~~~

I admit there is some beauty in this approach: It eliminates static functions from OcdFileExport, and `OcdIcon` totally encapsulates functionality which might be used and tested separately.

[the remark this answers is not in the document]
dg0yt: OcdIcon is key contribution of #1136.

src/fileformats/ocd_file_import.cpp
dg0yt: I leave this as an exercise to be done.

src/gui/symbols/icon_properties_widget.cpp
dg0yt: ad 1) already implemented  
ad 2) already implemented  
ad 3) now an explicit button  
ad 4) not sure about this one. We rescale on display, everything else is better handled in graphics software. However, I relabeled the spinbox to "Preview width", and moved the "width" lines to before the icons.  
ad 5) future anyway.

lpechacek:
> However, I still don't know how to deal with OCD 8 compressed icons.

I played a bit with the compressed icons and came to the conclusion that they are perhaps (TIFF style) [LZW](https://en.wikipedia.org/wiki/Lempel%E2%80%93Ziv%E2%80%93Welch) compressed.  A half black, half white icon which has 242 zeros and 242 0x7c's encodes as follows:

```
Half black square, binary notation

000000000 << 0x00 for black
100000010 << 0x102, first dictionary entry
100000011
100000100
100000101
100000110
100000111
100001000
100001001
100001010
100001011
100001100
100001101
100001110
100001111
100010000
100010001
100010010
100010011
100010100
100010101
100001011
001111100 <<< 0x7c for white
100011000
100011001
100011010
100011011
100011100
100011101
100011110
100011111
100100000
100100001
100100010
100100011
100100100
100100101
100100110
100100111
100101000
100101001
100101010
100101011
100100001
100000001 << (likely) end of data marker
111111000 << trailing bits
```

Unix [compress](https://github.com/vapier/ncompress) program gives very similar output with the exception of the first and last code. The first dictionary code from `compress` entry is 0x101, as opposed to 0x102 here, and there is no termination code. Searching around, I found out that [TIFF uses 0x101 as End Of Input marker](http://code.qt.io/cgit/qt/qtimageformats.git/tree/src/3rdparty/libtiff/libtiff/tif_lzw.c#n74) and its first dictionary code is 0x102, which is in line with what I see in the .ocd.

Will you look into implementation or shall I start coding (which may take long time)?

```
---- code scrap for later reuse
$ perl -e 'print "\0" x 242; print "|" x 242' | compress | xxd -g1
$ cat /dev/shm/binout.py 
#!/usr/bin/python3

import sys

for line in sys.stdin:
    print(line.strip(), format(int(line, 16), '08b'))
```

dg0yt:
> I played a bit with the compressed icons and came to the conclusion that they are perhaps (TIFF style) LZW compressed.

GIF/TIFF/LZW were my suspects too, but I always stopped at some point. Now I finally understood a) how to decode a LZW code stream when the original data was like aSaSa (such as a plain white line), and b) how GIF turns the code stream in a bit stream.

> Will you look into implementation or shall I start coding (which may take long time)?

I look into implementation again now. I will drop a note if I should put this aside again.

lpechacek:
The initial 16 bytes have some meaning too. They are often 0xFF's but in some cases there are different values. Two examples from OCAD 8 symbol set:

```
In symbol symbol 527.0 'Settlement': Unsupported compressed icon
0000   2b 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b ff b2 b2 b2 b2 5a 
0010   04 0e 24 58 d0 e0 41 84 09 15 2e 64 d8 d0 e1 43 88 
0020   11 25 4e a4 58 d1 e2 45 8c 19 35 6e e4 d8 d1 e3 c4 
0030   80 2b 2b 2b 2b 2b 2b 2b ff b2 b2 b2 b2 b2 b2 b2 b2 
0040   b2 b2 b2 11 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b ff b2 
0050   b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 11 2b 2b 2b 2b 2b 2b 
0060   2b 2b 2b 2b 2b ff b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 
0070   11 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b ff b2 b2 b2 b2 
0080   b2 b2 b2 b2 b2 b2 b2 11 2b 2b 2b 2b 2b 2b 2b 2b 2b 
0090   2b 2b ff b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 11 2b 2b 
00a0   2b 2b 2b 2b 2b 2b 2b 2b 2b ff b2 b2 b2 b2 b2 b2 b2 
00b0   b2 b2 b2 b2 11 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b ff 
00c0   b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 11 2b 2b 2b 2b 2b 
00d0   2b 2b 2b 2b 2b 2b ff b2 b2 b2 b2 b2 b2 b2 b2 b2 b2 
00e0   b2 11 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b 2b ff b2 b2 b2 
00f0   b2 b2 b2 b2 b2 b2 b2 b2 11 

In symbol symbol 401.0 'Open land': Unsupported compressed icon
0000   b9 bb b9 bb b9 bb b9 bb b9 bb b9 bb bb bb bb bb 73 
0010   04 0e 24 58 d0 e0 41 84 09 15 2e 64 d8 d0 e1 43 88 
0020   11 25 4e a4 58 d1 e2 45 8c 19 35 6e e4 d8 d1 e3 c4 
0030   80 b9 bb b9 bb b9 bb b9 bb bb bb bb bb bb bb bb bb 
0040   bb bb bb bb bb b9 bb b9 bb b9 bb b9 bb b9 bb bb bb 
0050   bb bb bb bb bb bb bb bb bb bb b9 b9 bb b9 bb b9 bb 
0060   b9 bb b9 bb b9 bb bb bb bb bb bb bb bb bb bb bb bb 
0070   bb bb b9 bb b9 bb b9 bb b9 bb b9 bb bb bb bb bb bb 
0080   bb bb bb bb bb bb bb b9 b9 bb b9 bb b9 bb b9 bb b9 
0090   bb b9 bb bb bb bb bb bb bb bb bb bb bb bb bb bb b9 
00a0   bb b9 bb b9 bb b9 bb b9 bb bb bb bb bb bb bb bb bb 
00b0   bb bb bb bb b9 b9 bb b9 bb b9 bb b9 bb b9 bb b9 bb 
00c0   bb bb bb bb bb bb bb bb bb bb bb bb bb b9 bb b9 bb 
00d0   b9 bb b9 bb b9 bb bb bb bb bb bb bb bb bb bb bb bb 
00e0   bb b9 b9 bb b9 bb b9 bb b9 bb b9 bb b9 bb bb bb bb 
00f0   bb bb bb bb bb bb bb bb bb 
```
There are 7 icons of this kind in the whole set.

src/core/symbols/symbol.h
```diff
@@ -345,7 +363,9 @@ class Symbol
 	/**
 	 * Clear the symbol's cached icon.
 	 * 
-	 * It will be recreated the next time getIcon() gets called.
+	 * Call this function after changes to the symbol definition, custum icon,
```
lpechacek: Typo here ("custum").

dg0yt:
> The initial 16 bytes have some meaning too.

To me, the unused data looks to me like left-over uncompressed garbage. But I have almost finished decompression, so you can test soon.

lpechacek:
> To  me, the unused data looks to me like left-over uncompressed garbage.

Indeed. All icons, including the above examples, load fine with changes from `ocd-compressed-icons`.

src/fileformats/ocd_file_import.cpp
```diff
@@ -800,6 +802,12 @@ void OcdFileImport::importSymbols(const OcdFile< F >& file)
 		symbol_index[ocd_symbol.number] = symbol;
 	}
 	resolveSubsymbols();
+	
+	for (auto ocd_symbol_entry : file.symbols())
+	{
+		auto& ocd_symbol = *ocd_symbol_entry.entity;
+		dropRedundantIcon(symbol_index[ocd_symbol.number], ocd_symbol);
```
lpechacek: Error case needs to be handled here.

If symbol import fails above, the corresponding `symbol_index` entry is `nullptr`. `dropRedundantIcon()` does not handle the `symbol == nullptr` case well.

dg0yt:
Now all comments should be resolved. This will be merged with the compression changes, once they are complete.

dg0yt:
Merged.

PR #1136, opened 2018-09-15
Add OcdIcon unit

dg0yt:
These changes move generic OCD icon properties (dimensions, palette) to the types in namespace Ocd, and add another class `OcdIcon` in namespace OpenOrienteering which deals with converting icons between OCD and Mapper/Qt.
(This is split out from PR #1129, taking into account review comments.)


dg0yt:
>  I'm surprised by the extra code preventing function template instantiation. I infer that text size is a concern for Mapper. If that is the case, it should be mentioned in coding style.

These function templates expand an array for the given type. I admit that it is a little bit overstretched here, given the net number of bytes to be saved. At least it also gives some consistency.

Issue #1125, opened 2018-08-19
Support line symbol distance from start/end


ollesmaps:
### Steps to reproduce
1. Import symbols that have round caps. As for example cliffs.
2. In case the round cap have in Ocad defined distance from start/ end - this distance does not get transferred.
3. This is problem especially by cliffs as minimum sizes, distances are effected

![rounded_caps](https://user-images.githubusercontent.com/20535899/44307716-a5b88780-a3a7-11e8-9c91-e95228ffd8cb.jpg)

### Expected behaviour
1. I suggest to have the round cap length adjustable also in Mapper.
2. I realize the double definition in Ocad - distance from start and also from end - is a problem. Perhaps if both are same it get transferred?

### Configuration
Mapper Version: 0.8.0
Windows 8

ollesmaps:
Is it really implemented? At 0.8.4 I can not mak it work.
The cap length for round cap is unadjustable.
![caplength](https://user-images.githubusercontent.com/20535899/59252115-a159a580-8c2b-11e9-97c4-3ea6b8f3f384.jpg)

dg0yt:
This issue is marked "Milestone v0.9.0". So this is not implemented in v0.8.4.

ollesmaps:
Thank you for clarification. I will try to read better.

ollesmaps:
As an interim solution. There is the option to use a point symbol at the end.
![hedge2](https://user-images.githubusercontent.com/20535899/59267047-7b42fe00-8c49-11e9-995f-77eb1a39b574.jpg)
OOM file: [hedge_roundedcap.zip](https://github.com/OpenOrienteering/mapper/files/3276163/hedge_roundedcap.zip)

Issue #1111, opened 2018-06-30
"Start symbol" & "End symbol" always combined with "Mid symbol"


Symbian9:
### Steps to reproduce

1. Run *OpenOrienteering Mapper*;
2. Create new symbol set;
3. Create new "Line" symbol;
4. Set only "Mid symbol" (left "Start symbol" & "End symbol" color blank) in this "Line" symbol settings.

### Actual behavior

When only "Mid symbol" defined both "Start symbol" & "End symbol" use it's settings.

![img 1](http://i.imgur.com/NfaL4kO.png)

![img 2](http://i.imgur.com/CNzLyoT.png)

In same time we has option for "Suppress the dash symbol at line start and line end", but it only suppress "Dash symbol" (if set) and not suppress "Mid symbol".

![img 3](http://i.imgur.com/nuIXsI8.png)

If try define "Start symbol" (for rewrite automatic "Mid symbol" settings applying) it still combined with "Mid symbol"

![img 4](http://i.imgur.com/cbtoKvq.png)

### Expected behavior

Add option that will give possibility for creating lines with defined "Mid symbol", but in same time with suppressed "Start symbol" & "End symbol".

### Configuration

Mapper Version: `0.8.x`, `0.9.x`
Operating System: Debian 9.x "Stretch" (Linux) , x86_64 (64bit)

dg0yt:
You need to specify the "Distance from line end" value. It is zero in your screenshot, but non-zero e.g. for ISOM2017 105 Earth wall.

(This distance applies to both "line ends" at the moment.)

PR #1138, opened 2018-09-19
Line symbols: Review combination of dashed line with pointed line caps

dg0yt:
In analysis of #1125, I stumbled over the implementation of pointed line caps, as used by the "Erosion gully" symbol. Generating the actual graphic primitives for dashed line features is one of the more complex parts of code in Mapper, taking quite some CPU and memory resources.

Given that:
- ISOM et al. don't need the combination of dashed line with pointed line caps,
- in Mapper, *pointed line caps already don't work properly with dashed lines* (cf. screenshot),
- in Mapper, pointed line caps behave different from other line caps (i.e. pointed line caps don't extend to the outside of dash lines, resulting in a dash layout which differs from common line caps),
- the layout is even more complex in the presence of "dash points",
- in graphics frameworks such as Qt, PDF and SVG, pointed line caps do not exist,
- at least some variants of dashed lines with pointed line caps might be implemented with "mid symbols" instead (similar to dot patterns),
- export to OCD format will not support the combination of dashed line styles with pointed line caps,

I propose to restrict the support of pointed line ends to continuous lines only, and as an alternative to the drawn (visible) line having an (maybe zero) offset from the start and end of the line feature (#1125). This change would:
- ensure the consistency of the layout of dashed line features,
- simplify the implementation,
- facilitate the export to OCD format.

Are there significant objections?

(This PR doesn't yet include the proposed change.)

cap-variants


ollesmaps:
I like the idea of zero offset from the start and end. However with hill/ depression contours it does not work well: 
hill_contour

dg0yt:
Zero offset will leave closed paths closed, so no problem with closed contour lines or even closed cliffs.

Non-zero positive offsets and pointed line ends are quite pointless for closed paths IMO, but IIRC the OCAD Viewer does handle this differently. Maybe you can test in OCAD a closed Gully or Cliff (#1125) and see if what you get is what you want .

Symbian9:
> ISOM et al. don't need the combination of dashed line with pointed line caps,

We don't know what could be added in next revisions of ISOM, so if something noe work, lets (at least) just keep it dissabled.

dg0yt:
This is where the changes lead so far:
 

As I wrote in the last commit, pure offsets at path start and end (#1125) can be implemented as a minimal variation, by simply not drawing the pointed ends but otherwise using the same layout. Of course with different length at start and end.

src/core/symbols/line_symbol.cpp
```diff
@@ -967,20 +977,18 @@ void LineSymbol::processDashedLine(
...
-	
-	auto groups_start = SplitPathCoord::begin(path_coords);
+	auto groups_start = start;
 	auto line_start   = groups_start;
 	for (bool is_part_end = false; !is_part_end; )
 	{
 		auto groups_end_path_coord_index = path_coords.findNextDashPoint(groups_start.path_coord_index);
 		auto groups_end_index = path_coords[groups_end_path_coord_index].index;
-		auto groups_end = SplitPathCoord::at(path_coords, groups_end_path_coord_index);
+		auto groups_end = groups_end_index < end.index ? SplitPathCoord::at(path_coords, groups_end_path_coord_index) : end;
```
dg0yt: This turned out to be to simple. Revised version of this commit is coming.

src/core/symbols/line_symbol.cpp
```diff
@@ -1298,11 +1308,11 @@ void LineSymbol::createMidSymbolRenderables(
...
 	}
 	
-	auto part_end = path.last_index;
-	while (groups_start.index != part_end)
+	while (groups_start.index != end.index)
 	{
 		auto groups_end_path_coord_index = path_coords.findNextDashPoint(groups_start.path_coord_index);
-		auto groups_end = SplitPathCoord::at(path_coords, groups_end_path_coord_index);
+		auto groups_end_index = path_coords[groups_end_path_coord_index].index;
+		auto groups_end = groups_end_index < end.index ? SplitPathCoord::at(path_coords, groups_end_path_coord_index) : end;
```
dg0yt: This turned out to be to simple. Revised version of this commit is coming.

src/core/symbols/symbol.cpp
```diff
@@ -453,7 +453,7 @@ QImage Symbol::createIcon(const Map& map, int side_length, bool antialiasing, qr
 			}
 			else if (line->getCapStyle() == LineSymbol::PointedCap)
 			{
-				line_length_half = std::max(line_length_half, 0.0012 * line->getPointedCapLength());
+				line_length_half = std::max(line_length_half, 0.006 * (line->startOffset() + line->endOffset()));
```
dg0yt: 0.0012 / 2 = 0.0006. Fix will be force-pushed.

Symbian9:
> LineSymbol: Add support for offset at start/end

So now Mapper has option to define different offset for line end & start, right?

If so, it would be useful for my request ;-)

- https://github.com/OpenOrienteering/mapper/issues/1111#issuecomment-403269920

dg0yt:
> So now Mapper has option to define different offset for line end & start, right?

Once this is merged, yes.

lpechacek:
> dg0yt requested a review from lpechacek 2 days ago

Will have to get up to speed on this one as I didn't follow the conversation closely. Will try to finish the icon pull request first and then I'll get back on this one.

Issue #904, opened 2017-07-28
Sorting of files in Mapper start menu on Android?


krticka:
Is it possible to sort files (for example by date) in Mapper start menu on Android? It seems to me, that there is no regularity.
I would also welcome option to hide files with file extensions like GPX or OCD.

Issue #1099, opened 2018-05-28
Android: mapper as standard application to open .omap files 


mlerjen:
Not sure if this issue is a mapper issue, but
### Steps to reproduce
1. Using nextcloud app to synchronize .omap file between desktop and phone
2. double-click on .omap file in nextcloud-app

### Actual behaviour
-> getting two suggested apps  to open the .omap file 
(mapper is not among them, both suggestions are unable to open .omap file)
(workaround: double-click on .omap file in the generic file manager -> opens with mapper)

### Expected behaviour
The link between .omap file and mapper should be disclosed to all kind of app

### Configuration
Mapper Version: 0.8.1
Operating System: Android 4.4.2.

dg0yt:
I didn' find out how to do that. It is not just about the .omap extension. IIRC we need to register a MIME type, then Mapper may handle this type.

PR #1150, opened 2018-10-07
Android: Show recent files, browse folders

dg0yt:
With this change, the mobile UI no longer show a single list of all maps in known locations, but:
1. Lists most recently used maps first (as on desktop).
2. If there are no such maps, show a link to the manual page.
3. List the known storage locations (unique full path).
4. Show a link to the examples folder.
    
When clicking on a folder, the folder contents will be displayed as follows:
1. Show ".." entry which navigates back.
2. List map files in this directory.
3. List subdirectories.
    
This change does not allow browsing to other locations.
    
Geospatial vector files will no longer be listed when they are handled by GDAL/OGR.
The "File list" headline and the redundant "Examples" button are removed.
    
This change at resolves or at least mitigates #904. However, this change does not add custom filtering or sorting, but just changes the way file types and folder structure are exposed to the user.

## Initial Home screen (no recent files)
home-screen-initial

## Home screen after clicking the "Examples" folder
home-screen-examples

## Home screen with recent files (read-only examples here)
home-screen-recent

### DIFF
--- a/src/core/storage_location.cpp
+++ b/src/core/storage_location.cpp
@@ -285,7 +285,7 @@ QString StorageLocation::fileHintTextTemplate(Hint hint)
 	switch (hint)
 	{
 	case HintNormal:
-		return tr("'%1' is stored in a regular location.");
+		return {};  // No text for a regular location.
 		
 	case HintApplication:
 		return tr("'%1' is located in app storage. The files will be removed when uninstalling the app.");
@@ -298,6 +298,12 @@ QString StorageLocation::fileHintTextTemplate(Hint hint)
 	}
 	
 	Q_UNREACHABLE();
+}
+
+
+QString OpenOrienteering::StorageLocation::hintText() const
+{
+	return hint() == HintNormal ? QString{} : fileHintTextTemplate(hint()).arg(path());
 }
 
 
--- a/src/core/storage_location.h
+++ b/src/core/storage_location.h
@@ -65,13 +65,21 @@ public:
 	/** Returns the path of this location. */
 	QString path() const;
 	
-	/** Returns the hint for this location. */
+	/**
+	 * Returns the hint for this location.
+	 * 
+	 * This function returns an empty string for HintNormal.
+	 */
 	Hint hint() const;
 	
 	/** Returns the text representing the hint for this location. */
 	QString hintText() const;
 	
-	/** Returns a text template for giving the hint for the given path. */
+	/**
+	 * Returns a text template for giving the hint for the given path.
+	 * 
+	 * This function returns an empty string for HintNormal.
+	 */
 	static QString fileHintTextTemplate(Hint hint);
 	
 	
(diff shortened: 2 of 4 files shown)


dg0yt:
I would appreciate some feedback before I merge this into 0.8.3.  
@Zerbembasqwibo, @mlerjen, @ollesmaps, @lpechacek ?

Zerbembasqwibo:
Suggestion:
"2. If there are no such maps, show a link to the manual page."
Think this is unnessecary sine "HELP"button exist.

Question:
"2. List map files in this directory."
What will be the order here? I suggest in order of letters.

Suggestion:
Does the known storage locations have to be shown with unique full path? Possible to shorten it with some ... to be able see the last part?

Otherwise fine with me.

dg0yt:
> "2. If there are no such maps, show a link to the manual page."
> Think this is unnessecary sine "HELP"button exist.

Note that unlike the Help button, this item links to the [page on Android storage](https://www.openorienteering.org/mapper-manual/pages/android-storage.html), and it goes away as soon as there are recent files to display. So it is there only when it is needed most: the first time a user installs the app.

> Question:
> "2. List map files in this directory."
> What will be the order here? I suggest in order of letters.

Files in directories are alphabetical, case insensitive. Recent files are most recent file first.

> Suggestion:
> Does the known storage locations have to be shown with unique full path? Possible to shorten it with > some ... to be able see the last part?

I started with that. It resulted in two entries "files" because it is "Android/data/org.openorienteering.mapper/files" on device storage and on memory card. So the unique part is in the beginning indeed. (I wouldn't mind improving this in later releases, but I don't want to put more time into it at the moment.)

Zerbembasqwibo:
OK good! *3


Den tors 11 okt. 2018 kl 22:05 skrev Kai Pastor <notifications@github.com>:

lpechacek:
I consider mixing "recent files" with real storage locations somewhat confusing. I'd welcome separation of these two entities into tabs or a separator line in the list. This is, however, not a blocker for merge from my POV. The conflation merely feels weird to me.

If anyone wants to see the feature live, my test build is at http://liborpechacek.wz.cz/oomapper/oomapper-mobile-ui-gadgets-20181011.apk .

dg0yt:
> I'd welcome separation of these two entities into tabs or a separator line in the list.

Tabs are not really optimal on small screens IMO. Separators are somewhat difficult with this widget, unfortunately. At least we never mix different types of folders.

And users need to be careful when switching app versions (testing your build) on Android, as uninstalling may remove files from folders.

lpechacek:
> Tabs are not really optimal on small screens IMO.

Yes, that came to my mind too. I'm wondering, however, how much common are small displays among Mapper users.

> Separators are somewhat difficult with this widget, unfortunately. At least we never mix different types of folders.

Ack.

> And users need to be careful when switching app versions (testing your build) on Android, as uninstalling may remove maps and data from some folders.

Thanks for the hint.

mlerjen:
Thanks a lot!

As I store my projects in separate folders mostly (1.) will be of great use to me. (3.) might also be, but here I wonder, how these "storage locations" are defined (it must be "mainfolders" somehow f.eg. nextcloud main folder, rather than each single projects folder)

For me this also would pretty much fix my issue #1099 as I would open the files from the application instead of the application from the .omap file.

mlerjen:
Thank you very much! (in hindsight, I would ban the sample maps from the recent list. They are in the samples folder anyway.)

PR #1152, opened 2018-10-08
Spot color properties

dg0yt:
I will merge this without waiting for approval, but for review and commenting, an explicit PR still seems beneficial.

### DIFF
--- a/src/core/map_color.cpp
+++ b/src/core/map_color.cpp
@@ -110,6 +110,8 @@ MapColor* MapColor::duplicate() const
 	copy->rgb_color_method  = rgb_color_method;
 	copy->flags = flags;
 	copy->spot_color_name = spot_color_name;
+	copy->screen_angle = screen_angle;
+	copy->screen_frequency = screen_frequency;
 	copy->components = components;
 	return copy;
 }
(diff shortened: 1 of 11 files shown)


lpechacek:
Celebration day - 128 closed pulls requests and 8 open! :)

Reading ISOM 2017, section 2.12.2 recommends stochastic and frequency modulated screens over traditional screen patterns. Mapper does not use these values for printing and export. That said, I wonder if there are still RIPs using these properties and I'd like to see screen properties being marked as OCAD compatibility feature in the UI.

dg0yt:
> I'd like to see screen properties being marked as OCAD compatibility feature in the UI.

Even in most OCD files I use for testing, these properties almost all use the same defaults (45°, 150 lpi) except for Yellow (30°, 150 lpi). And even if we would use it for PDF export, the print shop might use something else when it combines the map with other jobs on a single page. (The recent visit at Žaket made this much clearer.)

So for now, how about hiding these properties for files which do not already have them set (from OCD), or at least collapse them to a single expandable "> Screen properties"?

lpechacek:
> So for now, how about hiding these properties for files which do not already have them set (from OCD), or at least collapse them to a single expandable "> Screen properties"?

The proposal is to clarify usage of these fields in the UI. Hiding them is not serving the clarification purpose.

It should be clear for a new Mapper user that these values do not influence anything in normal Mapper operation. The user should know immediately that he/she does not have to adjust these values unless they want to influence OCD export. Marking these settings as "expert" ones would do the job too. Thanks!

PR #1153, opened 2018-10-09
Allow position change of a georeferenced template

dg0yt:
When using georeferenced images, e.g., from Google Maps (Satellite), sometimes
different images do not line up properly. It is therefore useful to be able to
move a template after it has been opened as a georeferenced one.

### DIFF
--- a/src/gui/widgets/template_list_widget.cpp
+++ b/src/gui/widgets/template_list_widget.cpp
@@ -221,12 +221,6 @@ TemplateListWidget::TemplateListWidget(Map* map, MapView* main_view, MapEditorCo
 	up_down_layout->addWidget(move_up_button);
 	up_down_layout->addWidget(move_down_button);
 	
-	// TODO: Fix string
-	georef_button = newToolButton(QIcon(QString::fromLatin1(":/images/grid.png")), tr("Georeferenced: %1").remove(QLatin1String(": %1")));
-	georef_button->setToolButtonStyle(Qt::ToolButtonTextBesideIcon);
-	georef_button->setCheckable(true);
-	georef_button->setChecked(true);
-	georef_button->setEnabled(false); // TODO
 	move_by_hand_action = new QAction(QIcon(QString::fromLatin1(":/images/move.png")), tr("Move by hand"), this);
 	move_by_hand_action->setCheckable(true);
 	move_by_hand_button = newToolButton(move_by_hand_action->icon(), move_by_hand_action->text());
@@ -238,6 +232,7 @@ TemplateListWidget::TemplateListWidget(Map* map, MapView* main_view, MapEditorCo
 	position_action = edit_menu->addAction(tr("Positioning..."));
 	position_action->setCheckable(true);
 	import_action =  edit_menu->addAction(tr("Import and remove"), this, SLOT(importClicked()));
+	georef_action = edit_menu->addAction(tr("Release georeference"), this, SLOT(releaseGeorefClicked()));
 	
 	edit_button = newToolButton(QIcon(QString::fromLatin1(":/images/settings.png")),
 	                            ::OpenOrienteering::MapEditorController::tr("&Edit").remove(QLatin1Char('&')));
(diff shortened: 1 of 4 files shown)


src/gui/widgets/template_list_widget.cpp
```diff
@@ -806,7 +800,9 @@ void TemplateListWidget::updateButtons()
 		if (temp->isTemplateGeoreferenced())
 		{
 			georef_visible = true;
-			georef_active  = true;
+			// Unsafe to release unloaded template
+			// See comment in Template::releaseGeoref().
+			georef_active = main_view->getTemplateVisibility(temp).visible;
```
dg0yt: Don't enable when the row is not checked. (If it is checked, then it is loaded!)
~~~
                   && template_table->item(visited_row, 0)->checkState() == Qt::Checked
~~~

src/templates/template.h
```diff
@@ -308,6 +308,11 @@ Q_OBJECT
 	 */
 	void unloadTemplateFile();
 	
+	/**
+	 * Release georeferencing flag for the template file.
+	 * Must not be called before the original georeferencing information is available.
+	 */
+	void releaseGeoreference();
```
dg0yt: I would prefer this to be place in the vicinity of `isTemplateGeoreferenced()`/`isTemplateGeoreferenced`.

src/templates/template.cpp
```diff
@@ -604,6 +604,18 @@ void Template::unloadTemplateFile()
 	emit templateStateChanged();
 }
 
+void Template::releaseGeoreference()
+{
+	// Calling this method on an unloaded template results in
+	// the template position being set to (0, 0, 1, 1, 0).
+	// XXX This assert is insufficient. FIXME!!!
```
dg0yt: The assert is okay for me ATM if the precondition is stated in the documentation. (We have a number of methods like this.)

src/gui/widgets/template_list_widget.h
```diff
@@ -140,6 +141,7 @@ protected slots:
 	QAction* move_by_hand_action;
 	QAction* position_action;
 	QAction* import_action;
+	QAction* georef_action;
```
dg0yt: And `georef_button` a few lines below can be removed now.

src/gui/widgets/template_list_widget.cpp
```diff
@@ -1064,6 +1060,27 @@ void TemplateListWidget::importClicked()
 	}
 }
 
+void TemplateListWidget::releaseGeorefClicked()
+{
+	auto temp = getCurrentTemplate();
+	if (!temp || !temp->isTemplateGeoreferenced())
+		return;
+	
+	temp->releaseGeoreference();
```
dg0yt: The remaining code might be better replaced by calling `updateButtons()`. And with that change, it could be clearer to invert the operation of th preceding `if` guard.

src/templates/template.cpp
sfroyen: I could not get the assert to trigger, hence the FIXME comment. In other words, I could still release an unloaded template.

src/gui/widgets/template_list_widget.cpp
sfroyen: That was what I intended, please replace the visibible test with the checkState test.

src/templates/template.cpp
sfroyen: Originally, I was looking for a way to load an unloaded template here. I think that is a better option. Is there an easy way to accomplish this?

src/gui/widgets/template_list_widget.cpp
dg0yt: You are right. I was confused by the different implementation in the other branch of the `if` statement.

src/templates/template.cpp
dg0yt: It actually just a call to `loadTemplateFile(false)`.

dg0yt:
Replaced by #1154.

sfroyen:
Thanks!

PR #1154, opened 2018-10-10
Allow changing the georeferenced state of template images

### LINKED ISSUE #1148: Enhancement: Changing position of a georeferenced template
When using georeferenced images, e.g., from Google Maps (Satellite), sometimes different images do not line up properly. It is therefore useful to be able to move a template after it has been opened as a georeferenced one. I have tweaked the code to always display the "move_by_hand_button" instead of the "georef_button". This allows me to move any template. However, saving and reopening the map causes any adjustment to be lost. I suspect there is some simple change (perhaps unsetting the is_georeferenced flag) that will make the new position permanent. The is_georeferenced flag is protected, however,  so it will be more involved than a one line change and I would appreciate some feedback or comments on how best to make the new position stick.

The final user experience should probably be something like this:

1) Tap the (renamed) Georeferenced button.
2) Tap away some type of warning.
3) Tap the move or edit buttons.
4) Move or edit the template position.
Step 4 side effect: template becomes not georeferenced,
5) If the template is removed and opened (not reopened) again, the original georeferenced position is restored

dg0yt:
Resolves #1148.  
Work in progress. known issues:
-  Closing the dialog when reenabling georeferencing after save/load should leave georeferencing disabled.


dg0yt:
Finished, and complemented with some cleanup. 

The buttons for manual adjustment now remain alyway visible, they just get disabled. From here, it might be the next step to leave them enabled, disabling georeferencing when the first adjustment is made, possibly with asking for confirmation first. But I don't want to change that much at the moment.

src/gui/widgets/template_list_widget.cpp
```diff
@@ -1064,6 +1060,31 @@ void TemplateListWidget::importClicked()
...
 
+void TemplateListWidget::changeGeorefClicked()
+{
+	auto* templ = getCurrentTemplate();
+	if (templ && templ->canChangeTemplateGeoreferenced())
+	{
+		auto new_value = !templ->isTemplateGeoreferenced();
+		if (new_value)
+		{
+			// Properly tear down positioning activities
```
lpechacek: Good!

src/gui/widgets/template_list_widget.cpp
```diff
@@ -235,8 +229,11 @@ TemplateListWidget::TemplateListWidget(Map* map, MapView* main_view, MapEditorCo
 	adjust_button->setCheckable(true);
 	
 	auto edit_menu = new QMenu(this);
+	georef_action = edit_menu->addAction(tr("Georeferenced"), this, SLOT(changeGeorefClicked()));
```
lpechacek: Suggest a more descriptive label for the menu entry. "Use georeferencing" or similar might better express changes associated with checking and unchecking the item than plain "georeferenced"..

src/gui/widgets/template_list_widget.cpp
```diff
@@ -784,66 +787,42 @@ void TemplateListWidget::updateButtons()
...
-		georef_action->setChecked(is_georeferenced);
-		georef_visible = true;
-		georef_active = is_visible && temp->canChangeTemplateGeoreferenced();
-		
-		custom_visible = !is_georeferenced || temp->canChangeTemplateGeoreferenced();
-		custom_active = !is_georeferenced && is_visible;
+		bool is_georeferenced = false;
+		bool edit_enabled   = false;
+		bool georef_enabled = false;
+		bool custom_enabled = false;
```
lpechacek: `custom_` is a very generic name prefix. How about `position_change_` or something better expressing that the variable controls visibility of template position controls?

Issue #986, opened 2017-10-06
Gpx thickness and colour


ollesmaps:
### Steps to reproduce
At the moment the thickness and colour of line that indicates position is hard-set. Sometimes the lines is hard to spot. It is a very thin line.

### Expected behaviour
I suggest to make the thickness of gps line adjustable (in percents as other templates). 

Additionaly it would be nice to tell retrospectively how bad/ good a gps signal was in a certain .gpx segment. That could be indicated by .gpx line color.

### Configuration
Android app 0.7.0.

dg0yt:
With GDAL based GPX support, the track is loaded as regular map internally. My idea is to add a hook to adjust default colors and symbols/dimensions for these types of templates (i.e. per template). At least color should be doable fairly easy. 

This needs additional elements in the file format.

ollesmaps:
Great, there is a way.

I noticed recently - gpxs from previous days got thicker line. Is it on purpose? I got used to thin ones - I could have them loaded all the time and have an idea where I had been. Now I must the previous one hide.

yevhenmazur:
Gpxs from previous days are not just thicker. They are scaled as other map content - very annoying behaviour.
![screenshot_20180205-114250-1](https://user-images.githubusercontent.com/10380431/35831910-fde8a40c-0ad3-11e8-9f8b-88b9f8daca9b.png)

dg0yt:
IIRC correctly we do have a little inconsistency in GPX tracks: The currently recorded track on Android still uses the legacy GPX code which draws "cosmetic" lines. Cosmetic lines, i.e. lines of one pixel width, are barely visible on high-dpi Android screens. When loading the map later, it may be loaded with GDAL/OGR, using a regular line symbol, so the line scales with zoom.

The trouble is that there is no single line width, fixed or not, which fits at all zoom scales (whole map, 1x, etc.), especially when working with many tracks in a single map.

dg0yt:
Above discussion seems to prefer the thin "cosmetic" lines. But I am quite unhappy with that poor visibility of the current track. I don't work on 105x zoom. For older tracks (thick lines when loaded with GDAL/OGR), I just reduce the opacity.

So what is the main concern at the moment:  
a) the current track being thinner than old ones, or
b) really just the significant width and scaling behaviour of the old ones (when loaded with GDAL/OGR)?  

[...]

PR #1157, opened 2018-10-14
GPX tracking corrections

dg0yt:
As a last set of changes for v0.8, this is to improve and fix issues with GPX tracks, including the pen width.


dg0yt:
The last change in this PR would resolve the line width inconsistency discussed in #986. While it improves track visibility for medium zoom levels, it establishes what was called "annoying" by @yevhenmazur. (For me, 105x zoom is not as important as track visibility for normal zoom. For reference, we have 1x and 2x as explicit options.) 

I wouldn't mind dropping this change, or perhaps toggle it with the GDAL/OGR GPX checkbox setting, for optimum consistency.

dg0yt:
This is the stable master branch, and I won't touch more then neccessary. Regarding track thickness, I see only two options for v0.8.3:
a) Leave everything as is. (Technically, it means not to merge the last two commits.)
b) Merge as proposed. Users may choose one of two consistent options by (un)checking GPX in the GDAL/OGR settings.

Any further improvement is for v0.9.

lpechacek:
OK, I somehow missed that this goes into stable branch. The additional changes should of course go into `dev`.

Regarding the two last patches, I don't see how they change behavior. With or without them GDAL imported GPX shows as thick line and legacy imported ones are drawn with cosmetic lines. I don't mind if they get merged though.

dg0yt:
> After the rewrite two variables in `MapEditorController::enableGPSDisplay()` are set in pair. `new_template` and `track` are either `true/nullptr` or `false/<valid pointer>`. IMHO `new_template` is obsolete now and also code at lines 3335-3340 can be removed.

This is incorrect. There is also `false/nullptr` when the existing template was loaded via GDAL/OGR (cannot qobject_cast OgrTemplate to TemplateTrack). 

> With or without them GDAL imported GPX shows as thick line and legacy imported ones are drawn with cosmetic lines.

This is incorrect. There would be either only cosmetic lines when GDAL/OGR is disabled for GPX, or only scaling lines when it is enabled, at least the next time the map is loaded. 
However, there seems to be little to no interest in the second option, so there is no need to add it now. We should rather put our focus on doing the right thing in dev.

dg0yt:
Moved the last two changes in a commit at https://github.com/OpenOrienteering/mapper/tree/gpx-pen-width.

lpechacek:
> This is incorrect.
...
> This is incorrect.

FWIW, I'm investing my time and effort into reviews in the hope that they will be useful, and not to be graded like an university student. Yes, I confused the qobject_cast() with the nearby new() operator invocation, making my review less valuable. And like I refrain on commenting the original code quality, for the author, who is likely following these conversations, could feel offended, I'd be glad if my mistakes were handled with politeness or ignored completely.

puzzlepaint:
At least for me, it is sometimes easy to overlook that something can be interpreted in an offensive way when focusing on technical details. Perhaps the "This is incorrect" statements could be edited out and then it immediately sounds much more constructive while not changing the information in the comment.

By the way, as a previous developer on this project, I think that it might be hard to offend me based on the quality of code that has been written many years ago ;) (not sure how much is left over anyway.) I am happy that the project is still being worked on.

lpechacek:
dg0yt asked me to revisit my previous comment. I wholeheartedly agree with Thomas, that without the introductory sentences the feedback is pretty factual and to the point.

It was the labeling that triggered my response. I pushed myself to keep pace with the incoming review requests and after I stumbled, admitted my failure and approved the change for `master`, saying that the other changes should be done in `dev`, I received two additional FAILs. That indeed was unexpected outcome of my time donation.

I guess my review feedback on other pull requests accelerated implementation of the OCD file handling and I'm happy for that. There are still some wrinkles to iron out but the base looks solid to me. Patch review is a demanding job without tangible outcome. For now I'll take a break and invest time and effort into things that give me joy. Thanks!

Issue #1121, opened 2018-08-11
Extend "Templates setup window" to show map parts 


Symbian9:
### Actual behavior

Currently *map parts* list exist as separate tool bar. It is not efficient and not best practice.

### Expected behavior

Show *map parts* in "Templates setup window" as separate sub-entries of `- Map -` entry. No need keep separate "Map parts" tool bar (that is unusable).

In future it would be better replace "Templates setup window" by "Project setup window" that will include all templates & map parts. With look like as "Outliner" in Blender or as "Layers" in Inkscape

![pic1](http://i.imgur.com/wIvuwHO.png)

![pic2](http://i.imgur.com/FMjq9Es.png)

dg0yt:
> No need keep separate "Map parts" tool bar (that is unusable).

Why do you think it is not needed or unusable? When using map parts, it seems very important to have a way to quickly *identify* and *switch* the active map part.

Symbian9:
> it seems very important to have a way to quickly
`identify` and `switch` the active map part.

"Map parts" tool bar unusable because it impossible do any other actions using this tool bar such as:

- "Add new part..."
- "Rename current part..."
- "Remove current part"
- "Move selected objects to"
- "Merge this part with"
- "Merge all parts"
- [ ] "Hide current part" (for temporary dissable any pats)

Function "Hide current part" is very important and similar option presented on "Templates setup window" for each *template* or for the whole `- Map -`.

"Move selected objects to" is not needed menu entry at all as we can replace it by doing next steps:

1. "Cut" selected objects on 1st map part;
2. Switch to 2nd map part;
3. "Paste" objects on 2nd map part;

### Project setup window

"Project setup window" should be unifed project organizer dashboard for control all *map parts*, *templates*, etc.

dg0yt:
I see, the tool bar is doing what I expect it to do, but you have very diffent ideas.
This toolbar is intentionally minimal. It is a separate toolbar a) to keep it hidden when there is just a single map part, and b) allow individual placement.

> "Move selected objects to" is not needed menu entry at all as we can replace it by doing next steps:

That is a poor argument. There are lots of features which would not be needed, e.g. Edit > Cut, which is the same as Edit > Copy + Delete.
In addition, the original feature does not change the current map part. You can immediately select the next objects and move them to another part.

Symbian9:
> In addition, the original feature does not change the current map part.

[...]

PR #1160, opened 2018-10-23
Remove legacy DXF and OSM support

dg0yt:
Mapper's initial support for DXF and OSM was implemented as addition to the GPX track feature, adding lots of complications and quirks which hinder maintenance and development. I propose to completely remove this legacy code now. There is better support of DXF and OSM through GDAL/OGR. (However, the visual representation is different from the wire-frame GPX track style.)

Who still depends on the old DXF/OSM implementation, enabled by unchecking DXF/OSM on the GDAL/OGR settings page?


yevhenmazur:
The waypoints description is not visible when GDAL for GPX import is enabled. However I don't think that it is a real blocker for this task. Just FYI.
#### GDAL Disabled
gdal-disabled
#### GDAL Enabled
gdal-enabled

dg0yt:
The idea is to remove the DXF/OSM cruft from the well-accepted existing GPX support ("GDAL disabled"), and then to improve this GPX support (color, line width) instead of the GDAL one.  At the moment it is just too complex.

(Personally, I'm not yet happy with how areas are rendered for OSM via GDAL. I think I' have to reactivate transparency for map colors, or implicitly apply the CRT files and symbol sets. But this is another task.)

Symbian9:
> I think I' have to reactivate transparency for map colors

+1 for transparency reactivating.

Also "map layers" (with opacity), as described in https://github.com/OpenOrienteering/mapper/issues/1121, should be part of this complex task.

dg0yt:
For the record, while Mapper should use the GDAL/OGR import by default since v0.7.92 (d9bca2bece47cf254ae0401f800932b222bec566), I had to learn recently that the legacy DXF and OSM support was used in fresh installations, in contradiction to the settings' checkbox state (#1166). So some users will notice the "new" import only in v0.8.4 or v0.9.0.

PR #1170, opened 2018-10-31
Sensors review

dg0yt:
Depends on #1169 "Cleanup Track".


dg0yt:
Still needs testing with real sensors.

PR #1179, opened 2018-11-15
DrawPathTool: Fix corner point handling

lpechacek:
I'm not completely sure I understood the code in draw path tool right but made my best to test all cases which came to my mind. Please review with extra care.

### DIFF
--- a/src/tools/draw_path_tool.cpp
+++ b/src/tools/draw_path_tool.cpp
@@ -291,19 +291,7 @@ bool DrawPathTool::mouseMoveEvent(QMouseEvent* event, MapCoordF map_coord, MapWi
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


dg0yt:
I tend to adopt this change.

IIUC the original (removed) code tries to restore the state which the tool had between mouse button press and drag start (> minimal distance). I.e. it was meant to create a straight edge on mouse button release, or allow to create a non-corner point.

However, it seems that we don't retain enough original state to successfully do so. At least `previous_drag_map` may be lost.

lpechacek:
> IIUC the original (removed) code tries to restore the state which the tool had between mouse button press and drag start (> minimal distance).

I see. Thanks for the explanation.

Issue #241, opened 2013-05-19
Windows: Limited printing resolution causes distortions


dg0yt:
**dg0yt reported on Sourceforge [[tickets:#241](https://sourceforge.net/p/oorienteering/tickets/241)]:**

When printing on Windows in vector mode, there are visible distortions for small items, such as dashes of 50x paths, green dots of 412 Orchards or black dots of 415 Cultivated land. This can be easily seen with virtual PDF printers.

_Background:_
The resolution from the printing dialog is applied as physical resolution but has no (direct) effect on the resolution of the printing coordinate system. The GDI-based print engine of Qt uses the "logical" resolution  reported by the printer, which is 300 dpi in many cases (but 600 dpi for the virtual XPS printer). All coordinates are _rounded to integers_ based on this resolution. 

For 300 dpi, the effective resolution is app. 0.085 mm, and so is the maximum error in distance of two neighboring points. A straight dashed path may no longer appear straight, given the fine offsets and irregular shapes of the indidual dashes. A black dot in 415 Cultivated land might shrink from nominal 0.2 mm to effective 0.17 mm or grow to effective 0.25 mm. The actual result is even worse because each dot is a path made of 12 distinct points.

The print preview and the direct PDF export are not affected, because they are implemented by other print engines.

![win-print-resolution](https://user-images.githubusercontent.com/13567791/48760034-30f63180-eca5-11e8-9862-5190c31fadcf.png)

PR #1182, opened 2018-11-19
Review printing

### LINKED ISSUE #1163: Bug when printing in vecteur mode with background raster
### Steps to reproduce
1. Open a raster template without géoréférencing. (aérial photo .jpg)
2. Move the angle
3. Draws some vectors on it (contours)
4. Print it in vertor mode with "show template" activated

### Actual behaviour
The preview is good, but the template does not appear in the final printed map.

I have an other issue with the same map imported from Ocad 2018:
The preview is good.
In the final printed map, the template appear corectly, but the vectors are very smal in the top left corner 

### Expected behaviour
If i make an PDF export, all is good.

### Configuration
Mapper Version: 0.8.3
Operating System: Windows seven

edit:
screenshots with printing/export dialog settings widget:
image
Résult with a virtual or réal printer:
image
zoom top left corner:
image

dg0yt:
This is a series of patches make the core MapPrinter::drawPage() function much easier to understand. It is a significant modification to the `master` branch, but I don't see a simpler way to eventually fix #1163.

There is still room for improvement, now more obvious than before, and more easy to add on top:
- Reduced opacity of the map "layer" will still trigger rasterization even when vector mode is selected - i.e. we must enforce 100% opacity for vector mode.
- In a similar way, rasterization might possibly be triggered from vector templates which are not 100% opaque. We might handle them like raster templates and rasterize them in the background, or we may handle them like the map "layer" and draw them 100% opaque in the foreground.
- I still need to verify if non-opaque colors in a vector drawing cause rasterization, too.
- From here, it is easy to make the code reuse the same temporary buffer allocation for drawing the map and foreground "layers".

### DIFF
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -40,6 +40,7 @@
 #include <QLatin1String>
 #include <QLocale>
 #include <QMessageBox>
+#include <QPaintEngine>
 #include <QPainter>
 #include <QPoint>
 #include <QPointF>
@@ -1100,20 +1101,22 @@ void Map::drawTemplates(QPainter* painter, const QRectF& bounding_box, int first
 	for (int i = first_template; i <= last_template; ++i)
 	{
 		const Template* temp = getTemplate(i);
-		bool visible  = temp->getTemplateState() == Template::Loaded;
+		if (temp->getTemplateState() != Template::Loaded)
+			continue;
+		
 		double scale  = std::max(temp->getTemplateScaleX(), temp->getTemplateScaleY());
-		float opacity = 1.0f;
+		auto visibility = TemplateVisibility{ 1, true };
 		if (view)
 		{
-			auto visibility = view->getTemplateVisibility(temp);
-			visible &= visibility.visible;
-			opacity  = visibility.opacity;
-			scale   *= view->getZoom();
+			visibility = view->getTemplateVisibility(temp);
+			visibility.visible &= visibility.opacity > 0;
+			scale *= view->getZoom();
 		}
-		if (visible)
+		if (visibility.visible)
 		{
+			Q_ASSERT(visibility.opacity == 1 || painter->paintEngine()->hasFeature(QPaintEngine::ConstantOpacity));
 			painter->save();
-			temp->drawTemplate(painter, bounding_box, scale, on_screen, opacity);
+			temp->drawTemplate(painter, bounding_box, scale, on_screen, visibility.opacity);
 			painter->restore();
 		}
 	}
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -1091,9 +1091,9 @@ void Map::drawColorSeparation(QPainter* painter, const RenderConfig& config, con
 	renderables->drawColorSeparation(painter, config, spot_color, use_color);
 }
 
-void Map::drawGrid(QPainter* painter, const QRectF& bounding_box, bool on_screen)
+void Map::drawGrid(QPainter* painter, const QRectF& bounding_box)
 {
-	grid.draw(painter, bounding_box, this, on_screen);
+	grid.draw(painter, bounding_box, this);
 }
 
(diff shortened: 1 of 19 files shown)


dg0yt:
While the particular template issue is solved, I'm still fighting with keeping vector output vectors. The behaviour is different across platforms:
- For macOS (Cocoa), QMacPrintEngine does not rasterize.
- For Linux (CUPS), QCupsPrintEngine is based on QPdfEngine (via QPdfPrintEngine)
which does not rasterize, at least for PDF 1.4 output.
  However, the native print system (CUPS) may rasterize, even when spooling to a PDF file. (The vector output is preserved when redirecting the output to a file in   
  Print... > Preview... > Print > Output file [...].)
- For Windows, QWin32PrintEngine is based on QAlphaPaintEngine.

QAlphaPaintEngines rasterizes:
- with the device's logical resolution but at least 300 dpi,
  cf. QAlphaPaintEnginePrivate::drawAlphaImage(const QRectF &rect)
- in tiles of 2044 pixel width and heigth,
  cf. QAlphaPaintEnginePrivate::drawAlphaImage(const QRectF &rect)
- when the painter is non opaque,
  cf. QAlphaPaintEngine::updateState(const QPaintEngineState &state)
- when the pen's brush is not opaque,
  cf. QAlphaPaintEngine::updateState(const QPaintEngineState &state)
- when the brush is not opaque,
  cf. QAlphaPaintEngine::updateState(const QPaintEngineState &state)

For undistorted vector output on Window, we must use an significantly increased resolution (cf. #241). This resolution causes (memory) problems when Qt's rasterization kicks in. Thus we must catch all rasterization triggers. But we only handle opacity (painter state) for templates, not painter state for the map, and not pens and brushes.

lpechacek:
This change set is growing large (`17 files changed, 484 insertions(+), 206 deletions(-)` as of today). @dg0yt, do you want me to devote some time for review of the patches?

And many thanks for jumping into fixing the printer code! This was one of the areas which yielded a few facepalms from me while crawling through it.

dg0yt:
I certainly appreciate review and independent testing. I spent a lot of time on breaking larger changes into traceable commits, mostly in order to facilitate such review efforts.

dg0yt:
I think it is ready now.

dg0yt:
> I think it is ready now.

No, I'm still seeing issues with order of templates in PDF export.

dg0yt:
Libor, I would like to merge this to `master` now (without review being finished) so that it becomes available for public testing via `master` builds. Do you accept?

lpechacek:
LGTM. Thanks for the diligent split of patches. This series was indeed easy to review.

I've got a remark though.

* 9cff2c055de8 MapPrinter: Rasterize in vector mode if needed
 I'm not sure if silent option setting is the right course of action. The user should be told that the option was set - either via an info dialog or a confirmation.

dg0yt:
Thanks for taking the time to review.

> I'm not sure if silent option setting is the right course of action. The user should be told that the option was set - either via an info dialog or a confirmation.

"Silent option setting" is what may happen to naive users of the API, indeed. However, for regular GUI users, this commit also adds a warning in PrintWidget stating that raster mode is needed because of transparent elements, and as a consequence it activates raster mode. I think this should match user expectations.

dg0yt:
BTW I studied Qt's print engine implementations. Some implement drawImage() using drawPixmap(), some do it just the opposite way. Our buffer images get really large at high resolutions. So I wonder if I shouldn't use QPixmap instead of QImage for those engines that prefer this. And of course, QWin32PrintEngine prefers pixmaps, by not implementing drawImage():  
https://code.woboq.org/qt5/qtbase/src/printsupport/kernel/qprintengine_win_p.h.html#QWin32PrintEngine

lpechacek:
> However, for regular GUI users, this commit also adds a warning in PrintWidget stating that raster mode is needed because of transparent elements, and as a consequence it activates raster mode. I think this should match user expectations.

Now I can see how the part changing `PrintWidget::setOptions()` fits in. It implements exactly the functionality I was missing. Thanks for clarification!

PR #1190, opened 2018-12-06
TemplateMap: Fix RenderConfig::scaling

### LINKED ISSUE #1186: Misrendered line objects when .omap file is used as a template
### Steps to reproduce
1. Add .omap file as a template
2. Zoom out a bit (approx. 2x)

### Actual behaviour
Line objects are misrendered. A fence became a tag-less line, erosion gully disappeared completely.

Example - see control site with code 41:
2x zoom:
image

4x zoom:
image

### Expected behaviour
The map should render in the same way as if it is opened directly.

The template at 2x zoom when opened directly:
image

### Configuration
Mapper Version: origin/master
Operating System: Android 7

dg0yt:
When drawing the map, RenderConfig::scaling must be set to the number
of pixels per mm. Fixes #1186.

### DIFF
--- a/src/templates/template_map.cpp
+++ b/src/templates/template_map.cpp
@@ -25,6 +25,7 @@
 
 #include <QtGlobal>
 #include <QByteArray>
+#include <QPaintEngine>
 #include <QPainter>
 #include <QRectF>
 #include <QStringList>
@@ -36,6 +37,7 @@
 #include "core/map.h"
 #include "core/map_coord.h"
 #include "core/renderables/renderable.h"
+#include "gui/util_gui.h"
 #include "util/transformation.h"
 #include "util/util.h"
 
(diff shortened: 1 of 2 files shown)


src/templates/template_map.cpp
```diff
@@ -139,9 +141,21 @@ void TemplateMap::drawTemplate(QPainter* painter, const QRectF& clip_rect, doubl
...
-	RenderConfig config = { *(template_map.get()), transformed_clip_rect, scale, options, qreal(opacity) };
+		scaling = Util::mmToPixelPhysical(scale);
+	}
+	else
+	{
+		auto dpi = painter->device()->physicalDpiX();
+		if (!dpi)
+			dpi = painter->device()->logicalDpiX();
+		if (dpi > 0)
+			scaling = dpi / 25.4;
```
dg0yt: `… *= …`

lpechacek: The only question left at this place is what is the exact meaning of `RenderConfig.scaling` and what is its unit of measurement. Without this information it is difficult to tell what is the appropriate   default for `scaling` in case the painter device does not provide DPI information. However, if you don't know the answer from top of your head, feel free to disregard the question and push the change into `master` as it indeed resolved the issue at hand.

dg0yt: Well, I already found the answer, and I added it to the documentation now.

lpechacek:
Assigning to Kai for final merge as I don't have the rights for pushing into `master`.

PR #1201, opened 2019-01-31
[WIP] SymbolRenderWidget: Add Yes To All action for symbol delete

lpechacek:
Minimal implementation of the enhancement.

### DIFF
--- a/src/gui/widgets/symbol_render_widget.cpp
+++ b/src/gui/widgets/symbol_render_widget.cpp
@@ -834,12 +834,26 @@ void SymbolRenderWidget::deleteSymbols()
 	}
 	
 	// delete symbols in order
+	bool yes_to_all = false;
 	for (auto* symbol : saved_selection)
 	{
-		if (map->existsObjectWithSymbol(symbol))
+		if (!yes_to_all && map->existsObjectWithSymbol(symbol))
 		{
-			if (QMessageBox::warning(this, tr("Confirmation"), tr("The map contains objects with the symbol \"%1\". Deleting it will delete those objects and clear the undo history! Do you really want to do that?").arg(symbol->getName()), QMessageBox::Yes | QMessageBox::No) == QMessageBox::No)
+			auto response = QMessageBox::warning(this, tr("Confirmation"),
+			                                     tr("The map contains objects with the symbol \"%1\"."
+			                                        " Deleting it will delete those objects and clear the undo history!"
+			                                        " Do you really want to do that?").arg(symbol->getName()),
+			                                     QMessageBox::Yes | QMessageBox::No | QMessageBox::YesToAll);
+			switch (response)
+			{
+			case QMessageBox::No:
 				continue;
+			case QMessageBox::YesToAll:
+				yes_to_all = true;
+				break;
+			default:
+				;
+			}
 		}
 		map->deleteSymbol(map->findSymbolIndex(symbol));
 	}
--- a/src/gui/widgets/symbol_render_widget.cpp
+++ b/src/gui/widgets/symbol_render_widget.cpp
@@ -46,6 +46,10 @@
 #include "core/symbols/text_symbol.h"
 #include "gui/symbols/symbol_setting_dialog.h"
 #include "gui/widgets/symbol_tooltip.h"
+#include "core/objects/object_operations.h"
+#include "undo/object_undo.h"
+#include "undo/symbol_undo.h"
+#include "undo/undo.h"
 #include "util/backports.h"
 #include "util/overriding_shortcut.h"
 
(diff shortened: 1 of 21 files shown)


src/gui/widgets/symbol_render_widget.cpp
```diff
@@ -834,12 +834,25 @@ void SymbolRenderWidget::deleteSymbols()
...
+			                                     tr("The map contains objects with the symbol \"%1\"."
+			                                        " Deleting it will delete those objects and clear the undo history!"
+			                                        " Do you really want to do that?").arg(symbol->getName()),
+			                                     QMessageBox::Yes | QMessageBox::No | QMessageBox::YesToAll);
+			switch (response)
+			{
+			case QMessageBox::No:
 				continue;
+			case QMessageBox::YesToAll:
+				yes_to_all = true;
```
dg0yt: break;

src/gui/widgets/symbol_render_widget.cpp
```diff
@@ -834,12 +834,25 @@ void SymbolRenderWidget::deleteSymbols()
 	}
 	
 	// delete symbols in order
+	bool yes_to_all = false;
 	for (auto* symbol : saved_selection)
 	{
-		if (map->existsObjectWithSymbol(symbol))
+		if (map->existsObjectWithSymbol(symbol) && !yes_to_all)
```
dg0yt: I would suggest to put the cheap boolean test first. No need to walk through the objects with yes_to_all.

lpechacek: Sure. Will include both changes in the next iteration.

src/gui/widgets/symbol_render_widget.cpp
lpechacek: Thanks for the express review, Kai!

lpechacek:
> I would like to propose having just a single `std::unique_ptr<...> takeObject(...)` (i.e. the same signature as `releaseObject` above).
...
> I didn't find any best practice hints on this topic.

I can see we are aligned. I would say all the guidelines I was able to find about expressiveness and intuitive design are very broad worded. Looks like it's our call to decide about a pattern and apply it on the code.

> The code example illustrates a change which asks for trouble: There will be event handlers/slots triggered synchonously when removing the object from the selection, and in this moment the object won't belong to any part. But that is an invariant which should hold when triggering unknown paths.

The remark confirms my spaghetti-code feeling I sometimes have. I'll keep signals in mind.

> My version:
>
> ```
>			auto object = target->getObject(i);
>			if (map->getCurrentPart() == target && map->isObjectSelected(object))
>				map->removeObjectFromSelection(object, false);
>			source->addObject(target->takeObject(i)); // No explicit move needed!
> ```

Understood. My preference was expressed by my version - take owning pointer once, do the processing and pass it on. Ownership transfer is clear between the two map parts and code performing the transfer. In this version object ownership is blurred to me. Code arranging the transfer never takes ownership of the map object and it can manipulate the object and its surroundings until `object` variable goes out of scope.

Anyway, here comes the first working version of symbol delete undo. Interestingly enough, I was able to eradicate `std::unique_ptr` from the patch set upon splitting my original large patch. It perhaps illustrates that owning pointers are a programming aid rather than requirement for a working program. I'll further refine the change with your comments in mind.

lpechacek:
So I happily implemented undo for all symbol operations until I looked into undo step persistence. The in-memory objects can be identified by its memory allocation pointer but stable symbol identifier seems to be required for saving into the on-disk format. Looking into the source, I can see there is unique `id` attribute which is however volatile and `code` attribute which is stable but nonspecific. For a while I contemplated on unique numeric id's for the symbols and even considered UUIDs but then gave up on the idea after realizing that symbol codes are unique in the standard and the other map drawing software.

At this point I'm wondering why are non-unique symbol codes allowed in Mapper. I appreciate that commit f1493cd190738 ("Handle arbitrary symbol IDs") implements just the single change but its commit message does not explain why support for colliding symbol codes has been introduced. @dg0yt, can you please shed light into the matter or [attach an explanatory note to the commit](https://git-scm.com/docs/git-notes)?

Regarding the owning pointers, I still keep the topic in focus. However, as the object ownership itself turned out to be tangential to this enhancement, I propose moving the discussion to the [refactoring project](https://github.com/OpenOrienteering/mapper/projects/6). The method naming part should perhaps be resolved here. I'm fine with any of `releaseXYZ() / takeXYZ()` as long as the scheme is documented and applied to the code base consistently. (looking around no clear naming pattern stands out - Linux uses `get / put` for resource life cycle management, Qt uses its object tree model, STL smart pointers assume ownership on construction and assignment and relinquishes it through `release`)

lpechacek:
Closed by mistake while cleaning up branches in my repo. Experience teaches me that these mistakenly closed PR's cannot be resurrected after the branch is restored. I'll place a new PR for tracking.

lpechacek:
Apologies for the noise.

PR #1202, opened 2019-01-31
Map: Prevent selecting hidden and protected symbols

lpechacek:
Alteration of protected objects comes as a surprise to users.

### DIFF
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -989,6 +989,12 @@ void Map::drawSelection(QPainter* painter, bool force_min_size, MapWidget* widge
 void Map::addObjectToSelection(Object* object, bool emit_selection_changed)
 {
 	Q_ASSERT(!isObjectSelected(object));
+
+	// we omit hidden and protected objects from any kind of selection
+	const auto* object_symbol = object->getSymbol();
+	if (object_symbol->isProtected() || object_symbol->isHidden())
+		return;
+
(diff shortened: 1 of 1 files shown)


dg0yt:
The code change looks fine. Do we need to amend the documentation part, too?

I had some strange effects on undo after a sequence of selecting/protecting/moving objects, but I'm unable to reproduce it, so I even cannot attribute it to this change ATM. (This involves `ObjectSelector`, which also uses (the result of) `Map::toggleObjectSelection()`, and we may get non-trivial behavioral changes here.)

lpechacek:
Pushed to a wrong branch. This one should have stayed at a01cd0f.

src/core/map.cpp
```diff
@@ -989,6 +989,12 @@ void Map::drawSelection(QPainter* painter, bool force_min_size, MapWidget* widge
 void Map::addObjectToSelection(Object* object, bool emit_selection_changed)
 {
 	Q_ASSERT(!isObjectSelected(object));
+
+	// we omit hidden and protected objects from any kind of selection
+	const auto object_symbol = object->getSymbol();
```
dg0yt: As stated in the other PR, I would prefer `auto*` for pointers.

lpechacek: Fixed in code. Thanks for the review!

Issue #1035, opened 2018-01-23
Ideas incubator


ollesmaps:
I miss the opportunity to incubate ideas somewhere, to talk about OOMapper new possible functions or ask some silly questions others might find obvious. Could that be here? And I mean here as in comments bellow this issue as an ever expanding conversation.

dg0yt:
Discussion is meant to happen here. 

 - There are 32 people "watching" this project, so I assume they get notified even without being mentioned explicitly.
 - Please try to keep different ideas in separate issues. (Cross-referencing is  easy, and this is another reason why discussion works well here.)
 - Please take the time look for existing issues.

For mere questions like this one, I attach the label 'question'.

ollesmaps:
Committing, Pushing and Pulling does it relate to code only? Or could you commit let's say to improving visual style of icons? If there is a good explanation of Github, please point me there.

dg0yt:
Everything file in the repository is subject to committing, pushing and pulling. Of course, features such as displaying "diffs" work much better with text files.

I don't know about a single document in [Github Help](https://help.github.com/). You will have to learn how to 
- create your "fork" of our git repository
- create and maintain "branches" within your fork
- finally create "pull requests" for incorporating your changes into our repository.

Partly this can be managed even in the Github website, but there are limits. In order to work with a local clone of the repositories on your computer, its probably most convenient for non-developers to use the [Github desktop software](https://desktop.github.com/).

ollesmaps:
Thank you Kai!

ollesmaps:
Once a map is opened in Anroid app, is there a way where to find map scale?
Once a map is opened on desktop, is the only place where to find map scale at Change map scale?

lpechacek:
> Discussion is meant to happen here.

My personal preference is a mailing list. [openorienteering-devel at SourceForge](https://sourceforge.net/p/oorienteering/mailman/) still works and might IMHO serve as a good place for discussing future directions. Question remains what are the preferences of the other thirty watchers.

Ad the map scale display in Android version, current display scale (zoom * map scale) could perhaps be part of the information toast, like "2x, 1:7500".

ollesmaps:
@lpechacek With regards to scale - I meant a defined scale for the whole file. You can not find it anywhere while using Android.

ollesmaps:
Is there a way how to add dash points while drawing in app? Not after drawing is finished, during drawing on Android device?

lpechacek:
[...]

PR #1203, opened 2019-02-01
[RFC] Resize multiple objects at once

lpechacek:
While making a 1:15 000 map, I noticed that most of the buildings imported from OSM are below minimum dimensions in my country. That led to necessity to enlarge objects en-mass to 160 % of their original size. First iteration was done by hand, then I thought how to automate that.

This morning I gave a shot to implementation and this is where it led. I added a new mode of operation to _Scale tool_. When holding `Ctrl` key, the resize operation uses object center (center of the enclosing rectangle) as the origin. As a bonus, double click bring in a dialog allowing to enter the scaling value in numeric format.

`Ctrl` + drag to resize multiple objects:
image

Double click to enter numeric value:
image

Known patch deficiencies:
- it does combine two changes in one (new mode of operation and numeric input)
- status line text is not updated
- double click for numeric input is not aligned with [a seemingly reasonable UI design rule](https://stackoverflow.com/a/4628577): double click function should be extension of function triggered by single click
- when holding `Ctrl` during double click, `Ctrl` key release is consumed by the modal dialog

How do you like the idea? Any opinions about the user interface?

### DIFF
--- a/src/tools/scale_tool.cpp
+++ b/src/tools/scale_tool.cpp
@@ -24,6 +24,8 @@
 #include <Qt>
 #include <QtGlobal>
 #include <QCursor>
+#include <QInputDialog>
+#include <QKeyEvent>
 #include <QLocale>
 #include <QPainter>
 #include <QPixmap>
@@ -75,6 +77,7 @@ void ScaleTool::updateStatusText()
 
 void ScaleTool::clickRelease()
 {
+	prev_scaling_center = scaling_center;
 	scaling_center = cur_pos_map;
 	updateDirtyRect();
 	updateStatusText();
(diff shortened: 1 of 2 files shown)


puzzlepaint:
Nice, this seems useful to me for that purpose (given that the status text is also updated such that users have a chance to learn about it). Regarding the user interface, how about making the object scaling dialog accessible via the program menu? Triggering the dialog via first selecting the scaling tool and then double clicking the map somewhere seems a bit quirky to me. For example, Inkscape seems to have something similar accessible via "Object -> Transform ..." (where also other transformations apart from scaling are possible). Another alternative would be to have a drop-down next to the scale tool in the same way the cut-hole and draw-on-template tools have one, and add a menu entry there. This wouldn't fit as well together with extending the dialog to a generic numerical transformation dialog as the other alternative would, however.

lpechacek:
@puzzlepaint: Thanks, Thomas, for the comment. In the meantime I realized that with the double click  behavior I perhaps tried to mimic the legacy OCAD pattern for drawing straight line segments. Let's forget about it. I also gave a shot to the separate menu entry and it didn't look good either. In the end, I opted for dropping this part. Setting scaling factor through mouse drag is sufficiently precise after all.

I can see that I mistakenly deleted the code branch and the pull request got closed. Now I'm unable to resurrect it for continuation, therefore I'm filing a new pull request with updated patch set.

ollesmaps:
Would be possible to use this tool also as "Move parallel"? 
Which can be used like this (staying with your @lpechacek  map sample) : snapping along the roads with fence symbol and then move parallel inside to get it further from the road. Possible?

lpechacek:
Let's move the discussion to Ideas incubator (#1035), Aleš.

PR #1205, opened 2019-02-07
README: Add Repology and Weblate badges

luzpaz:
* Repology badge shows the openorienteering package available in downstream repositories.
* Weblate badge shows the percentage of the openorienteering project that has been translated.

Preview commit at https://github.com/luzpaz/mapper/blob/README-badges/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
 # OpenOrienteering Mapper
+[![Packaging status](https://repology.org/badge/tiny-repos/openorienteering-mapper.svg)](https://repology.org/metapackage/openorienteering-mapper/versions)
+[![Translation status](https://hosted.weblate.org/widgets/openorienteering/-/svg-badge.svg)](https://hosted.weblate.org/engage/openorienteering/?utm_source=widget)
 
 ![Mapper Screenshot](https://www.openorienteering.org/mapper-manual/pages/images/main_window.png)
 


dg0yt:
Thanks for the proposal.

So far, I intentionally did not add badges to third-party resources to the README which is always displayed when viewing the Gh repository front page, in order to protect visitors from being tracked there. The information is available in developer wiki pages:
- https://github.com/OpenOrienteering/mapper/wiki/Target-systems#packages-in-linux-distributions
- https://github.com/OpenOrienteering/mapper/wiki/Translation

Maybe this concern is mitigated by GH acting as a proxy for such resources: The images seem to be loaded from https://camo.githubusercontent.com instead. Is this a documented GH feature we can rely on?

dg0yt:
> Maybe this concern is mitigated by GH acting as a proxy for such resources: The images seem to be loaded from https://camo.githubusercontent.com instead. Is this a documented GH feature we can rely on?

I could verify that image proxying for privacy is a documented Github feature, since 2014:
- https://help.github.com/en/articles/about-anonymized-image-urls
- https://github.blog/2014-01-28-proxying-user-images/

dg0yt:
@luzpaz Thank you very much for opening this pull request.

I will close this now without merging:
- I don't think the Repology badge provides significant value to most visitors of the GH page.
- I don't think the Weblate widget on the GH page will be effective in attracting additonal capable translators.

For similar considerations, I added other badges to separate wiki pages, not to the repository.
Of course, this is just my personal opinion.

Issue #642, opened 2016-03-30
.proj4 strings to EPSG is not working


marcovandersson:
I have an old map created in 0.5.96 in which I have used the .proj4 string definition for the swedish SWEREF99TM, to be able to import and use Kartaapullautin results in a geospatially correct manner. But when I try to change the definition in 0.6.1. to the corresponding EPSG:3006 the true coordinate system seems to be lost. The background template files (georeferenced) is no longer alinged with the maps features.

Not a really big issue since I can keep the .proj4 defintion. But anyway.

dg0yt:
Could you at least provide your old .proj4 definition?

marcovandersson:
The used definition is 
+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs
Which is according to ESPG:3006 @ http://spatialreference.org/ref/epsg/3006/

dg0yt:
I was not able to reproduce an error with a map created in Mapper 0.6.0 with the Proj4 definition, and then changing georeferencing to EPSG:3006. You may send test files to dg0yt AT darc.de.

marcovandersson:
The original file was probably created in 0.5.96 and is working in 97. But I cannot redefine to EPSG. The very same map also crashes when trying to open it with 0.6.1 (prob described in #618). I'll email you the file!

dg0yt:
It is not the change from PROJ.4 to EPSG definition which is causing the trouble. The original georeferencing setting have declination and grivation values of zero degrees. Any change to the georeferencing will trigger a recalculation of grivation. If you adjust the declination to a value such that grivation becomes zero (and don't have the map rotated), the georeferenced templates will stay in their current position, relative to the map

Mapper is designed for IOF standard maps, i.e. top is magnetic north. So the grivation results in a rotation of the grid. There is already ticket #331 for map north other than magnetic, but for now you will have to use a fake declination if you want a different arrangement.

dg0yt:
I would like to hear opinions on how to handle this. From my point of view it is the old file which has wrong data. In general it is correct to keep the declination when changing the coordinate reference system. However, we have a special case here, with both declination and grivation being zero - this is likely to be wrong and could trigger a dialog. Leaving the wrong data untouched is not an option IMO.

marcovandersson:
[...]

PR #1206, opened 2019-02-07
Case of convergence change affecting declination

pkturner:
I had a case in which a template became misaligned with the map, and tracked it down to the relationship between declination and grivation getting messed up by a georeferencing dialog.
To reproduce with my data in Tennessee, USA
  1.  Create map, open the georeferencing dialog, and set EPSG6575 into the georeferencing.
  2.  Set the reference point with EPSG coordinates
      922801  222772
  3.  DO NOT touch the declination, and click OK on the georeferencing dialog.
  4.  Import Willow_Springs_Park_contours.dxf to the map.
      Observe contour tile is drawn with EPSG6575 grid aligned to screen, i.e. grivation is 0.
  5.  Open Template control and add Willow_Springs_Park_slope.gif (with .gfw)
      with same georef as map.
  6.  Open georeferencing dialog. Shows both declination and grivation = 0.
  7.  Change declination to -6.87, choose option to "rotate map".
  8.  Observe template no longer aligned to map.
The above mentioned files can be downloaded from https://pkturner.org/1206.zip.

This PR has a suggested fix. In the existing code, when a CRS is added to Georeferencing, the declination and grivation are set if grivation is not locked. These are set via a call to setGeographicReferencePoint or setProjectedReferencePoint. I believe that any time either of these functions is called for a valid CRS, the convergence should be calculated and the declination and grivation should end up differing from one another accordingly. When one of these functions does not update grivation, it should update declination.

### DIFF
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -456,27 +456,33 @@ void GeoreferencingDialog::reset()
 
 void GeoreferencingDialog::accept()
 {
-	float declination_change_degrees = georef->getDeclination() - initial_georef->getDeclination();
-	if ( !grivation_locked &&
-	     declination_change_degrees != 0 &&
-	     (map->getNumObjects() > 0 || map->getNumTemplates() > 0) )
+	if (grivation_locked)
 	{
-		int result = QMessageBox::question(this, tr("Declination change"), tr("The declination has been changed. Do you want to rotate the map content accordingly, too?"), QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
-		if (result == QMessageBox::Cancel)
+		georef->updateGrivation();
+	}
+	else
+	{
+		float declination_change_degrees = georef->getDeclination() - initial_georef->getDeclination();
+		if ( declination_change_degrees != 0 &&
+		     (map->getNumObjects() > 0 || map->getNumTemplates() > 0) )
 		{
-			return;
-		}
-		else if (result == QMessageBox::Yes)
-		{
-			RotateMapDialog dialog(this, map);
-			dialog.setWindowModality(Qt::WindowModal);
-			dialog.setRotationDegrees(declination_change_degrees);
-			dialog.setRotateAroundGeorefRefPoint();
-			dialog.setAdjustDeclination(false);
-			dialog.showAdjustDeclination(false);
-			int result = dialog.exec();
-			if (result == QDialog::Rejected)
+			int result = QMessageBox::question(this, tr("Declination change"), tr("The declination has been changed. Do you want to rotate the map content accordingly, too?"), QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
+			if (result == QMessageBox::Cancel)
+			{
 				return;
+			}
+			else if (result == QMessageBox::Yes)
+			{
+				RotateMapDialog dialog(this, map);
+				dialog.setWindowModality(Qt::WindowModal);
+				dialog.setRotationDegrees(declination_change_degrees);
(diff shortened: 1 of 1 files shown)


dg0yt:
Thanks for your PR. The code changes look good to me, but I didn't find the time to actually verify problem and resolution. Maybe some other user can at least verify the problem in the current `stable` version, for his regional CRS.

In the meantime, I would like to ask you to clean up the pull request. There shall be only your changes (to `master` or `dev`), but no merge commits. (I think this can be achieved by rebasing on `master`, or to make it safe, cherry-pick the changes in a new branch based on `master`, and finally reset the `bump-declination` branch to the head of the new branch.) Thanks again.

pkturner:
OK. I'm a newbie to git, but it seems to have come out cleanly now.

pinnabilli:
I can reproduce the misalignment in current _stable_ with UTM 32 N.

Notice that, based on the lower right coordinate box, it is actually the map which becomes misaligned after the rotation mentioned at step 7: this can be seen by comparing geographic coordinates of any location on the template, which stay the same as one would expect, while for map objects the geographic coordinates change.

pkturner:
I redid the bump-declination branch again, having missed a comment change in the former round of cherry-picking.
@pinnabilli Thanks for taking a look. I double-checked that following the requested fix, the map and template are displayed with the correct geographic coordinates.

dg0yt:
I can confirm the problem. However, I think it might be the wrong spot to fix the issue. These `Georeferencing` functions are used from various location, including file loading, with `update_grivation = false`. OTOH the issue is only related to the `GeoreferencingDialog` AFAICT. So it probably must be fixed there, perhaps handling the grivation-locked input state in a different way when closing the dialog.

A second question is what to do then. I would tend to do the `updateGrivation()` call that was intentionally skipped during the grivation-locked input phase, thus respecting the user-supplied declination of 0. Your proposal shows the alternative: Update the declination, leaving grivation unchanged. The difference is whether your grid, map and templates appear rotated or not.

Regarding the proposed changes, I think we can safely assume `(state == Normal && isValid())` when we reach these else branches. Did you have a case where this assumption did not hold?

pkturner:
I'm glad to accept that `(state == Normal && isValid())` is unnecessary.
To limit the fix so that it only changes the behavior of `GeoreferencingDialog` makes sense. File loading makes me nervous, because the file could give you a declination and grivation that are inconsistent with the CRS.
However, if the behavior change occurs when closing the dialog, that is not great. The `GeoreferencingDialog` has the nice feature that it updates the displays of Declination and Grivation on the fly, so that when the user clicks "OK", they get what they saw. In that sense, the non-zero declination _is_ user-supplied. Once we decide what _should_ happen on "OK", the values under the "Map north" title can be continuously updated to reflect that.
I chose the "Update the declination, leaving grivation unchanged" alternative thinking that was the best way to respect the `grivation_locked` mode, which I do not understand very well.

pkturner:
Updated this PR to drop the `(state == Normal && isValid())` condition.

pkturner:
In #642 @dg0yt said,

> The dialog's "grivation locked" feature handles the fact that there is no point in determining grivation while you are entering the CRS and reference points, before you enter the declination. However it needs to be done in the end, instead of leaving all angles at zero.

Now I understand better what you're looking for. Since the declination is one of the user's primary points of control, it's preferable to leave the declination unchanged and waiting for the user to fix it. I will revise the PR code to update the grivation "in the end".

pkturner:
Revised the PR code to update the grivation "in the end".

dg0yt:
Replaced by #1222. Please review. Thanks!

pkturner:
I reviewed how this was incorporated in #1222.

PR #1210, opened 2019-03-02
Automatic Grid Scale Factor

pkturner:
### Automatic Grid Scale
Having enjoyed orienteering for many years, last September I began my first attempt at creating an orienteering map. Retirement has given me time for mapping, and beyond that I hope to contribute to the OpenOrienteering Mapper software development.

Just as Mapper automatically figures the convergence of the CRS at the reference point by examining the partial derivative of the projection along a meridian, it can use similar analysis to figure the grid scale factor.

This PR supports the development of existing maps without “correcting” their grid scale factor, or reinterpreting the georeferencing with which they were created. For existing maps, the map-to-projected transformation is generated exactly as it has been.
### User Control of Scale Factor
While strictly speaking, “grid scale factor” is a property of the projected coordinate system, the user has been able to provide a “grid scale factor” which may have included reduction due to elevation. I experimented with allowing the user to set an elevation scale factor, and decided to leave it out of the PR because my intent is to make the user interface easier, and to avoid obscure features.
### Compensation for Deformed Projections
As long as the code is finding the full derivative of the projection, it is possible also to handle projections that deform local shapes, such as plate carrée and web mercator. Since `QTransform` supports arbitrary affine transformations, the `from_projected` transformation can undo any local deformation present in the projected coordinates, resulting in a map that is free of deformation.
### Code Details
The analysis of grid scale factor needs the distance scale of the ellipsoid. This implementation gets the distance using a double stereographic projection centered on the reference point. This projection is supported by **PROJ** as `+proj=sterea`. With the distance scale established, partial derivatives of the projection at the reference point form a `grid_compensation` matrix. The convergence and grid scale factor can be readily calculated from this matrix.

Other notable elements of the code changes are:

- The flag `use_grid_compensation` is `false` for compatibility with existing maps, and is set to `true` for new maps. The user controls this setting in the Georeferencing dialog. When `grid_compensation` is enabled, the map-to-projected transformation is based directly on the declination, rather than on grivation.
- When the user enables automatic grid compensation, the "grid scale factor" displayed to the user is the calculated value.
- When saving and loading .omap files, Georeferencing that uses grid compensation is distinguished by the presence of a `grid_compensation` XML element. Although this matrix can be derived from the projected CRS, it is essential to Mapper, and saving it ensures numerical consistency.

### Comments Welcome
Comments on approach, user interface, code structure and coding style are welcome.


dg0yt:
I think these commits needs to be rebased on dev, and probably edited, to facilitate a review and further changes.

dg0yt:
I will clean this up from intermediate steps, and properly rebase on `dev`.

dg0yt:
Replaced by revised PR for `dev` in #1277.

ollesmaps:
I have Samsung Galaxy Tab Active 2 (Android 8.1.0) for a month now. And it originaly also stopped recording when screen went asleep (Mapper 0.8.4). I did some adjustments and it now works also in the sleep mode.  I assume the trick was.  Screen lock type. None. 
I also switched off: Floating icon. Pointer.

PR #1211, opened 2019-03-08
Scale tool enhancement

lpechacek:
This is continuation of PR#1203 which got closed inadvertently.

This pull request adds a new operation mode to Scale tool. When holding `Ctrl` key, the resize operation uses object center (center of the enclosing rectangle) as the origin. Further patches in the series refine the code change in sense of code deduplication and TODO removal.

`Ctrl` + drag to resize multiple objects:
image

### DIFF
--- a/src/gui/util_gui.cpp
+++ b/src/gui/util_gui.cpp
@@ -38,6 +38,8 @@
 #include <QLabel>
 #include <QLocale>
 #include <QMessageBox>
+#include <QPainter>
+#include <QPen>
 #include <QProcess>
 #include <QProcessEnvironment>
 #include <QSpacerItem>
@@ -230,6 +232,27 @@ namespace Util {
 	
 	
 	
+	namespace Marker
+	{
+		void drawCenterMarker(QPainter* painter, const QPointF& center)
+		{
+			const auto saved_hints = painter->renderHints();
+			painter->setRenderHint(QPainter::Antialiasing, true);
+			
+			painter->setPen(Qt::white);
+			painter->setBrush(Qt::NoBrush);
+			
+			/// \todo Use dpi-scaled dimensions
+			painter->drawEllipse(center, 3, 3);
+			painter->setPen(Qt::black);
+			painter->drawEllipse(center, 4, 4);
+			
+			painter->setRenderHints(saved_hints);
+		}
+	}  // namespace Marker
+
+
+
 	QSpacerItem* SpacerItem::create(const QWidget* widget)
 	{
 		const int spacing = widget->style()->pixelMetric(QStyle::PM_LayoutTopMargin);
--- a/src/gui/util_gui.cpp
+++ b/src/gui/util_gui.cpp
@@ -237,15 +237,19 @@ namespace Util {
 		void drawCenterMarker(QPainter* painter, const QPointF& center)
 		{
 			const auto saved_hints = painter->renderHints();
-			painter->setRenderHint(QPainter::Antialiasing, true);
+			painter->setRenderHint(QPainter::Antialiasing, Settings::getInstance().getSettingCached(Settings::MapDisplay_Antialiasing).toBool());
 			
-			painter->setPen(Qt::white);
+			const auto larger_radius = mmToPixelPhysical(1.1);
+			const auto smaller_radius = larger_radius*3/4;
+			
+			auto pen = QPen(Qt::white);
+			pen.setWidthF(larger_radius - smaller_radius);
+			painter->setPen(pen);
 			painter->setBrush(Qt::NoBrush);
-			
-			/// \todo Use dpi-scaled dimensions
-			painter->drawEllipse(center, 3, 3);
-			painter->setPen(Qt::black);
-			painter->drawEllipse(center, 4, 4);
+			painter->drawEllipse(center, smaller_radius, smaller_radius);
+			pen.setColor(Qt::black);
+			painter->setPen(pen);
+			painter->drawEllipse(center, larger_radius, larger_radius);
 			
 			painter->setRenderHints(saved_hints);
 		}
--- a/src/gui/util_gui.h
+++ b/src/gui/util_gui.h
@@ -30,6 +30,8 @@ class QCheckBox;
 class QDoubleSpinBox;
 class QLabel;
 class QObject;
+class QPainter;
+class QPointF;
 class QSpacerItem;
 class QSpinBox;
 class QWidget;
@@ -203,6 +205,13 @@ namespace Util
 	
 	
 	
+	namespace Marker
+	{
+		/** Center marker sign for rotate and scale tools. */
+		void drawCenterMarker(QPainter* painter, const QPointF& center);
+	}  // namespace Marker
+
+
 	namespace Headline
 	{
 		/**
--- a/src/tools/scale_tool.cpp
+++ b/src/tools/scale_tool.cpp
@@ -24,6 +24,7 @@
 #include <Qt>
 #include <QtGlobal>
 #include <QCursor>
+#include <QKeyEvent>
 #include <QLocale>
 #include <QPainter>
 #include <QPixmap>
@@ -36,6 +37,8 @@
 #include "core/map_view.h"
 #include "gui/map/map_widget.h"
 #include "core/objects/object.h"
+#include "gui/modifier_key.h"
+#include "gui/util_gui.h"
 #include "tools/tool.h"
 #include "util/util.h"
 
@@ -65,10 +68,19 @@ void ScaleTool::initImpl()
 void ScaleTool::updateStatusText()
 {
 	if (editingInProgress())
+	{
 		setStatusBarText(tr("<b>Scaling:</b> %1%").arg(QLocale().toString(scaling_factor * 100, 'f', 1)));
+	}
 	else
-		setStatusBarText(tr("<b>Click</b>: Set the scaling center. ") +
-		                 tr("<b>Drag</b>: Scale the selected objects. "));
+	{
+		auto status_text = tr("<b>Drag</b>: Scale the selected objects. ");
+		if (using_scaling_center)
+			status_text = tr("<b>Click</b>: Set the scaling center. ") +
+			              status_text +
+			              tr("<b>%1</b>: Switch to individual object scaling. ").arg(ModifierKey::control());
+
+		setStatusBarText(status_text);
+	}
 }
 
 
@@ -100,8 +112,17 @@ void ScaleTool::dragMove()
 	
 	auto scaling_length = (cur_pos_map - scaling_center).length();
 	scaling_factor = qMax(minimum_length, scaling_length) / qMax(minimum_length, reference_length);
-	for (auto object : editedObjects())
-		object->scale(scaling_center, scaling_factor);
+
+	if (using_scaling_center)
+	{
+		for (auto* object : editedObjects())
+			object->scale(scaling_center, scaling_factor);		
+	}
+	else
+	{
+		for (auto* object : editedObjects())
+			object->scale(MapCoordF(object->getExtent().center()), scaling_factor);
+	}
 	
 	updatePreviewObjects();
 	updateStatusText();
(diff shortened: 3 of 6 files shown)


src/tools/scale_tool.h
```diff
@@ -65,6 +68,7 @@ Q_OBJECT
 	MapCoordF scaling_center;
 	double reference_length = 0;
 	double scaling_factor   = 1;
+	bool group_scaling = true;
```
dg0yt: IMHO `group_scaling` isn't really naming the difference. We are dealing with a group of objects in both cases. How about `use_common_center`?

src/tools/scale_tool.cpp
```diff
@@ -115,18 +133,61 @@ void ScaleTool::dragFinish()
...
+		painter->drawEllipse(center, 4, 4);
+	};
+
+	if (group_scaling)
+	{
+		draw_center_sign(scaling_center);
+	}
+	else
+	{
+		for (const auto object : map()->selectedObjects())
```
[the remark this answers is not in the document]
lpechacek: I'd say it's a useful habit. I've played with `std::set`'s element type choice and I can see it determines the choice of member access operators. I like the expressiveness of  `auto*`, which is signalizing that subsequent `object` access uses `->` operator.

```
#include <set>
#include <memory>

struct Object
{
        void doNothing() const
        {
        }
};

int main()
{
        std::set<Object*> pointer_set;
        for (const auto* object_ptr : pointer_set)
                object_ptr->doNothing();

        std::set<Object> object_set;
        for (const auto object : object_set)
                object.doNothing();

        std::set<std::unique_ptr<Object>> unique_ptr_set;
        for (const auto& object_owning_ptr : unique_ptr_set)
                object_owning_ptr->doNothing();
}               
```

And there is [supporting blog post](http://www.nuonsoft.com/blog/2018/07/06/const-auto-versus-const-auto-for-pointer-types/) explaining the difference between `const auto var` and `const auto* var`. Indeed, modifying the loop body provides argument for using `const auto*`:

```
// compiles but should not
		for (const auto object : map()->selectedObjects())
			object->rotate(0);
```

```
// fails to compile, which is desired
		for (const auto* object : map()->selectedObjects())
			object->rotate(0);
```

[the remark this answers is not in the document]
lpechacek: Documented use of `auto*` in Coding conventions wiki page and fixed the code line. Thanks!

src/tools/scale_tool.cpp
```diff
@@ -67,8 +69,15 @@ void ScaleTool::updateStatusText()
...
-		                 tr("<b>Drag</b>: Scale the selected objects. "));
+	{
+		auto status_text = tr("<b>Drag</b>: Scale the selected objects. ");
+		if (group_scaling)
+			status_text = tr("<b>Click</b>: Set the scaling center. ") +
+			              status_text +
+			              tr("<b>%1</b>: Switch to individual object scaling. ").arg(ModifierKey::control());
+
+		setStatusBarText(status_text);
+	}
```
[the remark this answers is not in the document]
lpechacek: Good catch! Fixed.

src/tools/scale_tool.h
lpechacek: Right. `group_scaling` says too little about the difference between the two modes. `use_common_center` sounds a lot better. If you don't mind, I'd take your proposal a step further. Opting for present participle form of "use" and replacing "common" with "selection" would IMHO enhance fluency in reading the code. For example, the following code lines
```
if (using_selection_center)
    do_something(selection);
```
would read "when using selection center, do something with selection". How does the adjusted name sound to you?

src/gui/util_gui.h
```diff
@@ -203,6 +205,13 @@ namespace Util
 	
 	
 	
+	namespace GuiElements
```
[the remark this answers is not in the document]
lpechacek: Ack. `GuiElements` is unnecessarily broad.

lpechacek:
Proposed changes merged into a local branch. Once we agree on the variable naming, I'll post updated patch set. Thanks for the diligent review, Kai!

src/tools/scale_tool.h
dg0yt: 👍

lpechacek: Pushed updated patch set. Thanks for the smooth cooperation!

src/tools/scale_tool.cpp
```diff
@@ -100,8 +112,17 @@ void ScaleTool::dragMove()
 	
 	auto scaling_length = (cur_pos_map - scaling_center).length();
 	scaling_factor = qMax(minimum_length, scaling_length) / qMax(minimum_length, reference_length);
-	for (auto object : editedObjects())
-		object->scale(scaling_center, scaling_factor);
+
+	if (using_selection_center)
+	{
+		for (auto object : editedObjects())
```
dg0yt: Here and in the loop below, I would be glad if you could turn this into `auto object*`.

src/tools/scale_tool.h
```diff
@@ -65,6 +69,7 @@ Q_OBJECT
 	MapCoordF scaling_center;
 	double reference_length = 0;
 	double scaling_factor   = 1;
+	bool using_selection_center = true;
```
dg0yt: I spotted the slight different naming of `scaling_center` and `using_selection_center`. What I found now is that while the selection center is the default, the use may explicitly choose another scaling_center. From that point of view, renaming `scaling_center` is not an option, and `using_selection_center` is not 100% at target.

src/tools/scale_tool.cpp
lpechacek: I recall checking this exact place and I tend to believe `editedObjects` returns iterator and not a pointer. I'll study the interactions again.

src/tools/scale_tool.h
lpechacek: Ack. Will give it another attempt.

src/tools/scale_tool.cpp
dg0yt: Don't get sidetracked by the `ObjectsRange` class: It is the `for` loop which derefences the iterators it operates on internally.

lpechacek: Changed.

src/tools/scale_tool.h
lpechacek: So I resorted to `using_scaling_center`. Looks reasonably consistent to me at the moment.

lpechacek:
Changes in this iteration:
- rename of `using_selection_center` to `using_scaling_center`
- use of `auto*`

src/gui/util_gui.cpp
```diff
@@ -348,18 +348,22 @@ namespace Util {
...
-		painter->setRenderHint(QPainter::Antialiasing, true);
+		painter->save();
+		
+		const auto larger_radius = mmToPixelPhysical(1.1);
+		const auto smaller_radius = larger_radius*3/4;
 
-		painter->setPen(Qt::white);
+		painter->setRenderHint(QPainter::Antialiasing, Settings::getInstance().getSettingCached(Settings::MapDisplay_Antialiasing).toBool());
+		auto pen = QPen(Qt::white);
+		pen.setWidth(larger_radius - smaller_radius);
```
dg0yt: `QPen::setWidth` takes an `int` but the actual argument is a `qreal`. Either we need to call `QPen::setWidthF`, or round explicitly.

src/gui/util_gui.cpp
```diff
@@ -348,18 +348,22 @@ namespace Util {
 	
 	void Marker::drawCenterMarker(QPainter* painter, const QPointF& center)
 	{
-		const auto saved_hints = painter->renderHints();
-		painter->setRenderHint(QPainter::Antialiasing, true);
+		painter->save();
```
dg0yt: Note that `QPainter::save()` is convenient but will [allocate](https://code.woboq.org/qt5/qtbase/src/gui/painting/qpainter.cpp.html#_ZN8QPainter4saveEv) a new `QPainterState`. I try to avoid it when I can - which usually means when I touch just hints, pen and/or brush, as in this function.

src/gui/util_gui.cpp
```diff
@@ -343,6 +345,22 @@ namespace Util {
 		}
 		return maybe_markup;
 	}
+	
+	void Marker::drawCenterMarker(QPainter* painter, const QPointF& center)
```
dg0yt: Which style shall we eventuall adopt to implement namespaced free functions: `::` style as here, or explicit `namespace` blocks as elsewhere in this file?

src/gui/util_gui.cpp
lpechacek: Before I look into addressing implicit rounding, style polishing and micro-optimizations, there is one topic which may have escaped your attention. This code is the only instance which bothers itself with assembling markers from graphic primitives. Markers displayed during line editing are scaled bitmaps provided by `PointHandles` class. Isn't extending and reusing `PointHandles` for the center markers more efficient?

src/gui/util_gui.cpp
dg0yt: This is already flagged with a warning in Qt Creator.

src/gui/util_gui.cpp
dg0yt: I didn't want to insist on a change here. It is just that I became careful about what I do i during `paint()`. I can't say that `PointHandles` has been designed or reviewed from this point of view. So I wouldn't recommend an immediate switch to that class at the moment. However, having these "decorations" in a single place would facilitate future improvements.

src/gui/util_gui.cpp
dg0yt: I didn't want to insist on a change here. It is just that I stumbled about the free function looking like class member, different from the style used for other free functions in this particular file.

lpechacek: The file is [already inconsistent](https://github.com/OpenOrienteering/mapper/blob/9adf5c7053de8dafeb03219815d1c89efe40eba4/src/gui/util_gui.cpp#L217-L232) and needs to be checked for style anyway.

I've introduced `namespace` in my change but my personal taste is towards use of scope resolution operator (`::`) in function definitions. Supporting arguments are:
- Creator uses this notation when triggering the "Add function definition to .cpp file". If we use enclosing name spaces, programmers can forget about this editor function or have to rewrite the generated code anyway.
- The accompanying name space block enforces code indentation which is limiting usable horizontal space. We have given up on indentation within global `OpenOrienteering` name space but not the deeper nested ones.

Whichever approach we take, existing code base still needs adjustments.

src/gui/util_gui.cpp
lpechacek: Using the more lightweight `setRenderHint()` over the catch-all `Painter::save()` makes sense to me.

src/gui/util_gui.cpp
lpechacek: Interesting that my GCC 9 reported dozens of other warnings on current `origin/dev` but not this particular one. Which compiler/code analyzer complained about this line?

On related note, what are the plans towards static code analyzers and code changes arising from that kind of automated review? On QuickEvent Coverity [did quite good job](https://github.com/fvacek/quickbox/commits?author=lpechacek).

lpechacek:
Changes in this iteration:
- use `setWidthF()`
- enclose function definitions in `namespace`
- avoid use of `Painter::save()`

Please note that each patch set respin involves manual conflict resolution and another round of testing on every patch. I'd be glad for review feedback covering as many requested changes as possible in one round so that I can invest time currently spent on rebasing into application enhancements and fixes.

dg0yt:
> Which compiler/code analyzer complained about this line?

It was one of the static checks which are now a regular feature of Qt Creator. I guess it is the "Clang-only check for almost everything", `-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-unused-macros -Wno-newline-eof -Wno-exit-time-destructors -Wno-global-constructors -Wno-gnu-zero-variadic-macro-arguments -Wno-documentation -Wno-shadow -Wno-switch-enum -Wno-missing-prototypes -Wno-used-but-marked-unused`
(Now that I looked it up, I can see how to turn on clang-tidy and clazy checks)

> The file is already inconsistent and needs to be checked for style anyway.

Indeed, thanks for pointing out.

> Please note that each patch set respin involves manual conflict resolution

I guess this set of commits whould have benefitted from rebasing onto a more current state of `dev`.

> I'd be glad for review feedback covering as many requested changes as possible in one round so that I can invest time currently spent on rebasing into application enhancements and fixes.

That's what I normally try to do.

PR #1215, opened 2019-03-20
Rewrite object tags undo save/load

dg0yt:
The first commit is just a modernization. The second one fixes a clazy warning which pointed to a DRY violation. This should result in one compiler warning disappearing in the new CI, so that's why I'm inviting more reviewers than usual.


dg0yt:
I had to edit the Azure pipeline settings to actually be able to get a CI build for this PR.

PR #1218, opened 2019-03-22
Use PROJ 5.1 API

pkturner:
This updates Georeferencing to use the PROJ 5.1 API. I'm initiating this PR to show the status of this work. I've tested it on Linux, and will test more thoroughly including on Android.

The `pj_strerrno` function is replaced with `proj_errno_string`, which was introduced in 5.1.
Search uses the environment's `PROJ_LIB`, as suggested by @dg0yt.

### DIFF
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -870,9 +870,6 @@ QDebug operator<<(QDebug dbg, const Georeferencing &georef)
 }
 
 
-}  // namespace OpenOrienteering
-
-
 
 #if defined(Q_OS_ANDROID)
 
@@ -919,3 +916,6 @@ void copyProjAssets()
 }
 
 #endif  // defined(Q_OS_ANDROID)
+
+
+}  // namespace OpenOrienteering
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -37,7 +37,7 @@
 #include <QXmlStreamReader>
 #include <QXmlStreamWriter>
 
-#include <proj_api.h>
+#include <proj.h>
 
 #include "core/crs_template.h"
 #include "fileformats/file_format.h"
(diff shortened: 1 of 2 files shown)


dg0yt:
Please, base you work on `dev`, not `master`.

I'm not sure how fast we actually want to remove compatibility with PROJ 4.9. Ubuntu 18.04 will be LTS at least until 2020. I don't know when Debian 9 will be replaced.
https://github.com/OpenOrienteering/mapper/wiki/Target-systems#component-version-matrix

pkturner:
I brought the proj-5 branch up-to-date with dev. Now it looks like a mess. Maybe it needs a different branch and/or a new pull request.

Also, checks are failing due to PROJ 5.0's `<proj.h>` not found in Azure.

pkturner:
@dg0yt 
> I'm not sure how fast we actually want to remove compatibility with PROJ 4.9.
Understood. Seeing the component version matrix, you are juggling a great number of dependencies.

dg0yt:
I think you could do the following now:
~~~
git rebase ed39289 --onto dev
~~~

This should cherry-pick your new commits (which are after ed39289) on top of dev.

I'm about to skip PROJ 5 for PROJ 6 in Superbuild if it works for Android, too. However, I'm struggling with the GDAL update to 2.4 which I want to do at the same time. Maybe it is easier to wait a few days.

pkturner:
Changing the PR's base branch in GitHub from `master` to `dev` cleared up the "mess".

dg0yt:
On the PROJ mailing was a [thread about porting to PROJ 6](https://lists.osgeo.org/pipermail/proj/2019-July/008740.html). Some takeaways:
- We need to explicitly handle axis order, at least by axis `proj_normalize_for_visualization`.
- In the long run, we may need to handle (at least for parity with GIS)
  - dynamic datum
  - late binding

dg0yt:
This is now entirely replaced by #1341. Thanks for the excellent starting point. It even helped find bugs in PROJ.

PR #1226, opened 2019-04-06
SymbolRenderWidget: Fix crash when deleting multiple symbols

xc-racer99:
When deleting more than half of the symbols at one go (and the second one that is clicked on is after the first one), and then creating a new symbol; Mapper crashes.

This crash is due to the fact that current_symbol_index is still at the index of the last symbol clicked on and is now out of bounds so AddSymbol crashes when trying to add a new symbol there.

Solve it by explicitly setting the selected symbol (and therefore current_symbol_index) to the invalid -1

### DIFF
--- a/src/gui/widgets/symbol_render_widget.cpp
+++ b/src/gui/widgets/symbol_render_widget.cpp
@@ -841,8 +841,12 @@ void SymbolRenderWidget::deleteSymbols()
 		map->deleteSymbol(map->findSymbolIndex(symbol));
 	}
 	
-	if (selected_symbols.empty() && map->getFirstSelectedObject())
-		selectSingleSymbol(map->getFirstSelectedObject()->getSymbol());
+	if (selected_symbols.empty()) {
+		if (map->getFirstSelectedObject())
+			selectSingleSymbol(map->getFirstSelectedObject()->getSymbol());
+		else
+			selectSingleSymbol(-1);
+	}
 }
 
 void SymbolRenderWidget::duplicateSymbol()


xc-racer99:
Note that I've based this against master as it is a relatively trivial bugfix, but if you'd prefer it to go in dev I can easily rebase.

lpechacek:
> When deleting more than half of the symbols at one go (and the second one that is clicked on is after the first one), and then creating a new symbol; Mapper crashes.

I've tried to replicate the crash with a custom v0.8.4 build but didn't succeed, perhaps due to misunderstanding the instructions. What I did:
1) Created new file with ISOM 2017 symbol set.
2) Selected the first symbol and Shift-clicked somewhere beyond the half of the table to select "more than half of the symbols".
3) Clicked right mouse button above the second symbol in the table to get the context menu and chose Delete.
4) Created new point symbol with number "1" and description "x", clicked OK. Symbol was created without any application crash or other noticeable misbehavior.

@xc-racer99, would you be so kind as to provide more detailed information about replication procedure and frequency of crashes?

xc-racer99:
> I've tried to replicate the crash with a custom v0.8.4 build but didn't succeed, perhaps due to misunderstanding the instructions. What I did:
> 
>     1. Created new file with ISOM 2017 symbol set.
> 
>     2. Selected the first symbol and Shift-clicked somewhere beyond the half of the table to select "more than half of the symbols".
> 
>     3. Clicked right mouse button above the second symbol in the table to get the context menu and chose Delete.
> 
>     4. Created new point symbol with number "1" and description "x", clicked OK. Symbol was created without any application crash or other noticeable misbehavior.
> 
> 
> @xc-racer99, would you be so kind as to provide more detailed information about replication procedure and frequency of crashes?

Interesting, that's essentially how I can trigger the crash 100% of the time (running Debian Stretch and version 0.6.7, packaged master, latest self-compiled master, latest self-compiled dev).  Here's a complete list of steps:

1. Create a new map, using any symbol set (eg ISOM2017_1500)
2. Left click on first symbol, shift-left click on last symbol (or one just a couple from the end)
3. Right click on selected symbols, click Delete
4. Immediately right click in symbol widget area, create new symbol (of any sort) giving it number 1 and description x.
5. Hit Ok, and Mapper immediately exits with no warning and no chance to save

When debugging through, I found that https://github.com/OpenOrienteering/mapper/blob/master/src/gui/widgets/symbol_render_widget.cpp#L1088 creates a new symbol at index current_symbol_index when current_symbol_index >= 0 - this index is used as-is in https://github.com/OpenOrienteering/mapper/blob/master/src/core/map.cpp#L1791 for a simple insert operation into symbols vector - however, current_symbol_index never gets updated when the symbols are deleted so it will try to insert a new symbol way past the end of the vector leading to the crash.

Hopefully this explanation makes things a little more clear.

dg0yt:
I could reproduce the problem now. Note that you have to click into the "white" area of the symbol widget to create the new symbol.

lpechacek:
Now I got it and can replicate the bug too. Very specific procedure which should IMHO be reflected in patch description. Good job done catching such tiny insect!

lpechacek:
Looking at the proposed fix I can see multiple ways to fix the issue. xc-racer99's is the minimal one.

I can see two deeply ingrained issues causing this crash. First, `Map::*Symbol()` methods are doing no sanity checking on provided index, even in debug build. Second, routines manipulating `SymbolRenderWidget::selected_symbols` do not update associated `current_symbol_index`. How about updating `current_symbol_index` close to the `selected_symbols.erase(it)` in `SymbolRenderWidget::symbolDeleted()` so that the index value is always valid?

dg0yt:
I tend to approve the PR as is.

However, I agree that the way how `SymbolRenderWidget` is designed and implemented needs a second look, and some work.

lpechacek:
The code change is fine with me. I'd like to see the replication instructions and reference to this ticket (GH-1226) mentioned in the commit message though.

@xc-racer99, can you please refresh the patch?

xc-racer99:
Patch refreshed to contain the more detailed instructions on how to reproduce the crash.

Thanks!

Issue #1248, opened 2019-05-15
FileFormatTest occasionally crashes on Windows


dg0yt:
In native Windows builds (MSYS2 MinGW CI), FileFormatTest occasionally crashes when/after testing the new OCD format implementation. The error does not occur on Linux or macOS.

Context:
- mingw32 and mingw64
- after passing the OCD10 stage or the OCD11 stage for 'forest sample.omap'

If this isn't just a bug in this particular test, other OCD related bugs (e.g. #1234) might have the same root cause, even if our current Windows binaries are cross-compiled (and the test couldn't be run this way.)

Example:
~~~
PASS   : FileFormatTest::saveAndLoad(data:/examples/forest sample.omap <> OCD8)
PASS   : FileFormatTest::saveAndLoad(data:/examples/forest sample.omap <> OCD9)
PASS   : FileFormatTest::saveAndLoad(data:/examples/forest sample.omap <> OCD10)
A crash occurred in D:\a\1\s\build\test\file_format_t.exe.
Function time: 6977ms Total time: 7056ms

Exception address: 0x00007ff8edf64860
Exception code   : 0xc0000005
Nearby symbol    : memcpy

Stack:
#  1: ZN5QTest15toPrettyUnicodeEPKti() - 0x0000000061d82760
#  2: UnhandledExceptionFilter() - 0x00007ff8ec2c64e0
#  3: memset() - 0x00007ff8effa6440
#  4: _C_specific_handler() - 0x00007ff8eff8fca0
#  5: _chkstk() - 0x00007ff8effa4570
#  6: RtlWalkFrameChain() - 0x00007ff8eff03730
#  7: KiUserExceptionDispatcher() - 0x00007ff8effa33d0
#  8: memcpy() - 0x00007ff8edf64840
#  9: Unable to obtain symbol
# 10: Unable to obtain symbol
# 11: Unable to obtain symbol
# 12: Unable to obtain symbol
# 13: Unable to obtain symbol
# 14: Unable to obtain symbol
# 15: Unable to obtain symbol
# 16: Unable to obtain symbol
# 17: ZNK11QMetaMethod6invokeEP7QObjectN2Qt14ConnectionTypeE22QGenericReturnArgument16QGenericArgumentS5_S5_S5_S5_S5_S5_S5_S5_S5_() - 0x00000000098dcce0
# 18: ZN5QTest5qInitEP7QObjectiPPc() - 0x0000000061d87780
[...]

PR #1250, opened 2019-05-22
Ocd: Review and test string assignments

dg0yt:
Resolves GH-1248.

### DIFF
--- a/src/fileformats/ocd_types.h
+++ b/src/fileformats/ocd_types.h
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2013, 2015-2018 Kai Pastor
+ *    Copyright 2013, 2015-2019 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -22,7 +22,6 @@
 
 #include <algorithm>
 #include <cstddef>
-#include <cstring>
 #include <iterator>
 #include <type_traits>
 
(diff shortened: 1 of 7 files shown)


src/fileformats/ocd_types.h
```diff
@@ -125,7 +125,7 @@ namespace Ocd
 		
 		Utf16PascalString& operator=(const QString& value)
 		{
-			memcpy(reinterpret_cast<void*>(data), reinterpret_cast<const void*>(value.utf16()), std::max(std::size_t(value.length())+1, 2*N));
+			std::memcpy(data, value.utf16(), 2*std::min(static_cast<std::size_t>(value.length())+1, N));
```
dg0yt: Note: it is `min`, not `max`.

dg0yt:
I wasn't yet able to test with OC*D Viewer.

dg0yt:
Last push: Updated copyright, removed extra `;` (-Wpedantic).

lpechacek:
Will review. Hoping to get to it by the end of week.

dg0yt:
Thanks.

What I could have done differently from the start is omitting the ocd_t.h file and put everything into the ocd_t.cpp file (cf. symbol_t.cpp). But I leave it as is for now.

dg0yt:
- Removed `ocd_t.h`
- Reviewed commit messages and source code comments.
- Fixed GH issue reference.

I didn't add explanations for "why things are done as they are at the moment" because these strings like other OCD stuff are simply based on the documentation of this external file format. Maybe the naming of the string types could have been done differently, but this is out of scope now.

I intentionally did not introduce a function for common test data, because there are significant differences anyway, with regard to length and encoding.

I intentionally didn't want to mix commits changing behaviour with commits moving code around.

lpechacek:
Fine with me. Thanks for the quick turnaround!

Issue #1245, opened 2019-05-07
New OCAD 10 export changes color index


jmacura:
### Steps to reproduce
1. Create map in OCAD 10
2. Open in Mapper
3. Export to OCAD 10 (new file, do not rewrite)
4. Open in OCAD 10
5. Copy some objects from map created in step no. 1

### Actual behaviour
Index of colors in the table is renumbered => The file is no longer color-compatible with its previous versions. Image is for thousand words:

Objects I am copying from the original file:
![ocd10_exporting](https://user-images.githubusercontent.com/5598693/57262590-03435000-706d-11e9-9c72-1a3be3f333ef.jpg)

Objects how they appear in the new file (exported from Mapper):
![ocd10_importing](https://user-images.githubusercontent.com/5598693/57262545-e3ac2780-706c-11e9-972d-b34718d915c0.jpg)

How does the table of colors look like in the original file:
![ocd10_colors_before](https://user-images.githubusercontent.com/5598693/57262621-2a9a1d00-706d-11e9-8c5b-d553ff17cadb.PNG)

How does it look like in the new file (exported from Mapper):
![ocd10_colors_after](https://user-images.githubusercontent.com/5598693/57262627-31289480-706d-11e9-91ee-25453fad33d9.PNG)

### Expected behaviour
The index in table of colors should remain the same. Why this is important? Because many maps are created in teams. Some team-members use OCAD and some use Mapper, these files must remain mutually compatible.

### Configuration
Mapper Version: 0.9.20190326
Operating System: Windows 10

lpechacek:
Is step 5 (copy objects from original map) mandatory or are the display colors skewed in step 4, please?

dg0yt:
I don't think we do preserve the index numbers at the moment. But we should eventually add a ID field (and the same for symbols, objects).

lpechacek:
I've just played with OCAD 8 a bit, assuming the behavior is the same in higher versions.

So I set up a new map from orienteering symbol set, created a few objects. Saved a copy of that map, deleted color entries and added new colors with the same indices at their place (**Delete** color entry followed by **New**). Opened the previously  saved copy, selected an object, copied it and pasted into the modified map. No surprise, the pasted object came out in a different color.

Bottom line is that OCAD blindly references color entries by their indices upon pasting objects from a different map. For good interoperability Mapper should retain the color indices.

[...]

PR #1251, opened 2019-05-29
[WIP] SymbolRenderWidget: Add Yes To All action for symbol delete #1201

lpechacek:
Continuation of #1201. The cherry-pick of "Accept use of deprecated PROJ API" is there solely for the purpose of compiling Mapper on my system. Will be left out from the final patch set.

Currently there is open question about symbol code uniqueness. Without unique symbol id the undo operations cannot be serialized. The solution is to either assign symbols another id, make symbol codes unique again, or some other clever idea which has not yet occurred to me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,7 @@ if(NOT TARGET PROJ4::proj)
 	set(PROJ4_FOUND false)
 	find_package(PROJ4 MODULE REQUIRED)
 endif()
+add_definitions(-DACCEPT_USE_OF_DEPRECATED_PROJ_API_H)
 
 if(Mapper_USE_GDAL)
 	find_package(GDAL REQUIRED)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -203,6 +203,7 @@ set(Mapper_Common_SRCS
   
   undo/map_part_undo.cpp
   undo/object_undo.cpp
+  undo/symbol_undo.cpp
   undo/undo.cpp
   undo/undo_manager.cpp
   
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -2081,15 +2081,24 @@ int Map::addObject(Object* object, int part_index)
 
 void Map::deleteObject(Object* object, bool remove_only)
 {
+	auto object_ptr = releaseObject(object);
+	if (!remove_only)
+	{
+		delete object_ptr;
+	}
+}
+
+Object* Map::releaseObject(Object* object)
+{
 	for (MapPart* part : parts)
 	{
-		if (part->deleteObject(object, remove_only))
-			return;
+		if (auto object_found = part->releaseObject(object))
+			return object_found;
 	}
 	
-	qCritical().nospace() << this << "::deleteObject(" << object << "," << remove_only << "): Object not found. This is a bug.";
-	if (!remove_only)
-		delete object;
+	qCritical().nospace() << this << "::deleteObject(" << object << "): Object not found. This is a bug.";
+
+	return nullptr;
 }
 
 void Map::setObjectsDirty()
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -1600,6 +1600,11 @@ void Map::setSymbol(Symbol* symbol, int pos)
 
 void Map::deleteSymbol(int pos)
 {
+	delete releaseSymbol(pos);
+}
+
+Symbol* Map::releaseSymbol(int pos)
+{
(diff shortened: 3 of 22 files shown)


dg0yt:
IMO the first commit looked fine within the scope outlined in the title. Instead of adding review comments, I just submitted a similar change which adds a "Cancel" option (#1259).

The rest should become a separate PR (Or you change the subject of this one, as it is still quite young). To keep things simple, I suggest you rebase these API changes onto the latest `dev`.

> Without unique symbol id the undo operations cannot be serialized. The solution is to either assign symbols another id, make symbol codes unique again, or some other clever idea which has not yet occurred to me.

Maybe the actual problems is that the widget changes the order of the symbols in the Map when all it needs to do is change the order of the symbols in the widget? With a stable order of symbols in the map, it might be enough to carefully handle the order of undo/redo. We do the same with objects IIUC. And of course, there would be less noise for files maintained in git.

lpechacek:
> > Without unique symbol id the undo operations cannot be serialized. The solution is to either assign symbols another id, make symbol codes unique again, or some other clever idea which has not yet occurred to me.
>
> Maybe the actual problems is that the widget changes the order of the symbols in the Map when all it needs to do is change the order of the symbols in the widget?

Yes, that's the point where the card collapses.

> With a stable order of symbols in the map, it might be enough to carefully handle the order of undo/redo. We do the same with objects IIUC. And of course, there would be less noise for files maintained in git.

Yes, that is a possible solution. That way the programmer will have to be careful not to change symbol positions while e.g. importing another map into existing one, and there will have to be an attribute indicating symbol icon position in the symbol widget. As a result, unique symbol id will be obscured (expressed as the element position in map file) while auxiliary information will be recorded explicitly. On the positive side, symbol id collisions cannot occur. :) While this is a working solution, I'm still inclined to give symbols unique identity - in database world it would be called table primary key.

I see three options at the moment - use ISOM symbol codes (Mapper has to ensure uniqueness while exporting to OCD anyway so why couldn't these numbers be unique in OMAP too?), use UUID (ugh, heavy hammer but should work) or use the table position as an id.

dg0yt:
We seem to agree that the symbol order in the widget should be decoupled from the identification of the symbols ("primary key") and from the order of symbols in our file format.

Regarding the type of ID:
- I think it is sufficient to have IDs which are unique within the file. From that point of view, UUID would be overkill.
- I wouldn't like to use the symbol codes. Enforcing them to be unique within a file would require to immediately resolve collisions on all kind of data import.
- I would propose to use a auto incrementing integer, more less continuing what we now use as "index", but now as an explicit attribute. 
   - Probably most the index-based interface could be reused (perhaps with some renaming).
   - Given common symbol set sizes and frequency of changes, the IDs would be fairly small, and we could perhaps still use the ID as an array index.
   - Deleting symbols would create "gaps" which might be reused when a new symbol is added.
   - We might even provide some limited compatibility for older Mapper versions by filling the gaps with placeholder symbols which would be ignored by newer Mapper versions.

Regarding the order of symbols in the widget:  
I would rather not put the position in a (single) symbol attribute, but save the whole collection as a separate entity: This would allow for handling multiple arrangements.

Last not least, these ideas apply as well to the map colors, where we want to solve #1245. However, I am more doubtful about the ID being a good array index, depending on how small and dense IDs are in OCD files. Anyway, colors might be an alternative starting point to explore these ideas.

lpechacek:
A quick ack on the proposal. I also contemplated on the auto-incrementing integer id but then stopped at the vision of symbol entity decorated with unique description, unique appearance, unique ISOM code and unique internal id. That is why I looked in the direction of reusing the ISOM symbol number.

Let's go down the integer id path, although collisions will have to be resolved on imports anyway.

lpechacek:
Merely sharing my surprise and slight amusement arising from my realization that symbol id's in .omap file are in fact plain strings. While mapper assigns numeric id's to symbols, it is still possible to rewrite id's to arbitrary strings and the map loads fine.

Example:
```
<?xml version="1.0" encoding="UTF-8"?>
<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="9">
<symbols count="1">
<symbol type="4" id="***AREA_SYMBOL***" code="1" name="A"><area_symbol inner_color="-900" min_area="0" patterns="0"/></symbol>
</symbols>
<parts count="1" current="0">
<part name="default part"><objects count="1">
<object type="1" symbol="***AREA_SYMBOL***"><coords count="4">-48598 -26520;-25443 -4173;-6058 -30828;-48598 -26520 18;</coords></object>
</objects></part>
</parts>
</map>
```

Issue #1200, opened 2019-01-30
Allow deleting all symbols in one run


lpechacek:
Suggested enhancement.

### Steps to reproduce
1. Select many symbols
2. Trigger Delete symbols operation

### Actual behaviour
Every symbol that is associated with an object in the map triggers a Yes/No dialog asking for confirmation.

### Expected behaviour
A "Yes to all" option in the dialog would be convenient.

### Configuration
Mapper Version: stable branch build 0.8.20181215
Operating System: openSUSE Tumbleweed

lpechacek:
I've placed a pull request against `dev` containing a straightforward implementation of the enhancement. However, while testing the change, I can see the user experience is not yet perfect. When a large symbol set is selected and delete operation is triggered by mistake, the operation can neither be canceled nor undone. I can imagine this might be a fearful moment for some users spiced with uncertainty how to restore lost symbols.

I'm wondering what are your preferences for handling this situation. I can see following options:
1) Add Cancel button to the dialog. That way only unused symbols up to the first used one are lost.
2) Implement undo for symbol deletion.
3) Implement undo capability for all symbol changes while we are at it.
4) _... any better idea_

Feel free to share your thoughts in comments.

lpechacek:
> In the long term, we always need undo (i.e. 2, 3).

I'm working on 3 and I'm wrestling with map object ownership transfer between `MapPart` and `UndoStep`. I'm tempted to turn `MapPart::ObjectList` into `std::vector<std::shared_ptr<Object>>` for the manual resource management is just annoying. Any advice or information about preferred solution would be welcome.

dg0yt:
I wouldn't mind modernizing ownership, however this is a separate development topic with huge impact. 
- It works even better with full C++14 (having `std::make_unique`), which requires the successful switch to clang for Android (work in progress).
- In this particular spot, I would rather expect a `std::unique_ptr`, because there is quite clear, exclusive ownership. And with passing and returning `std::unique_ptr`, it becomes clear where ownership is transfered. There are spots where this is done already, although not always according to best practise (e.g. `void UndoManager::push(std::unique_ptr<UndoStep>&& step)`, which should be `void UndoManager::push(std::unique_ptr<UndoStep> step)`.
- IMO the key problem here is not object ownership but identity. We use both pointers and index numbers, and this what is so fragile. We need a stable identity, also to facilitate tracking and versioning of map data.

lpechacek:
[...]

PR #1259, opened 2019-06-10
Add YesToAll and Cancel to symbol deletion

dg0yt:
When deleting multiple symbols, users may be asked very often to
confirm deleting the affected objects. This changes adds "Yes to all"
and "Cancel" options which will both cause the dialog to not appear
again for the current operation.

For "Cancel" to affect the whole operation, the confirmation request
is moved to the first sub-step where the symbols are collected for
later deletion.

Resolves GH-1200.

### DIFF
--- a/src/gui/widgets/symbol_render_widget.cpp
+++ b/src/gui/widgets/symbol_render_widget.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2014 Kai Pastor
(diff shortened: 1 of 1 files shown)


dg0yt:
> Looks good to me except the commit message. According to Coding Conventions subject line should start with name of affected unit. It is not the case here. 

I removed the unit prefix because I felt it was better to not have the end cut off where git subjects get limited space, and I didn't see how to solve it in another way. Will try again without "symbol".

> I'd also welcome a brief Suggested-by or something like that. Otherwise approved. Thanks!

Using full name or Github mention.

PR #1268, opened 2019-06-21
Improve map scale and rotate dialogs

dg0yt:


src/gui/util_gui.cpp
```diff
@@ -275,7 +275,7 @@ namespace Util {
 		}
 #endif
 		
-		QSpinBox* create(int min, int max, const QString &unit, int step)
+		QSpinBox* create(int min, int max, const QString &unit, int step, bool wrapping)
```
lpechacek: Please [avoid `bool` parameters](https://wiki.qt.io/API_Design_Principles#The_Boolean_Parameter_Trap). Function/method calls stuffed with `true/false` are pretty difficult to read. Some sort of "creation flags" might be an alternative.

dg0yt: Fully acknowledged. I'm still somewhat reluctant to go the way of adding flags. All these parameter should not be used outside of `template<class P>create()`. I'm thinking about moving more code to the template function (in the header), with direct access to the actual `InputProperties` type instead of passing parameters. However, this would mean that the unit calling `create<ActualInputProperties>()` would have to include the matching spinbox header.

dg0yt: Wrapping now handled in `template<class P>create()`.

PR #1289, opened 2019-07-18
Fix transform used when importing a template map

pkturner:
When importing a template map, objects need their coordinates transformed in the steps: scale, rotate, translate. This order needs to be the opposite of how they are applied to QPainter for drawing templates.

To experience the bug,

1. Open a map and add a template map.

1. Use the template dialog's positioning widget to set obvious rotation and non-isotropic scaling, such as rotation of 90, scale-x of 2, and scale-y of 0.5.

1. Observe that the template map is rotated counter-clockwise, and is greatly elongated vertically.

1. Use the template dialog's feature to import and remove the template map.

1. Observe that the imported map is now greatly elongated horizontally.

The imported features should not have changed their location/scaling. The rotation and scaling were applied correctly to the template map, and were applied incorrectly during import.

### DIFF
--- a/src/gui/widgets/template_list_widget.cpp
+++ b/src/gui/widgets/template_list_widget.cpp
@@ -110,8 +110,8 @@ struct ApplyTemplateTransform
 	
 	void operator()(Object* object) const
 	{ 
-		object->rotate(transform.template_rotation);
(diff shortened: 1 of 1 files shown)


dg0yt:
Oops, touched Enter key before finishing editing the commit message.
Thanks for the PR.

PR #1291, opened 2019-07-19
Compare TemplateTransforms will check template_y member

pkturner:
TemplateTransform::operator== compares all the other members, but missed comparing the template_y member. Simple fix.


dg0yt:
Good catch, thanks.

PR #1292, opened 2019-07-19
TemplateTransform: allow any affine transformation

pkturner:
Enhances `fromQTransform` and `struct TemplateTransform` to fully reflect the
affine transformation that defines them.

This PR does not require #1277, though the need for each PR may arise in related circumstances.

In comparison to the approach taken with #1277, this PR is lighter weight. Rather than add to or replace TemplateTransform with an affine `QTransform`, a scalar `template_shear` member is added to the existing data structure.

### Example

[wintri-maps.zip](https://github.com/OpenOrienteering/mapper/files/3412647/wintri-maps.zip)
[7_tile_slope.zip](https://github.com/OpenOrienteering/mapper/files/3412645/7_tile_slope.zip)

The example maps illustrate a case in which the alignment of a template image can be improved. As with #1277 this case uses the Winkel III projection. For the current PR, it's not a matter of straightening out the map, but of aligning the template with the map that is skewed by its projection.

1. Open and view the junior.2019-05-24a.wintri.deformed.omap in Mapper.  This map is correctly georeferenced with the Winkel III projection that stretches the map along a southwest-to-northeast axis.
1. Load  7_tile_slope.gif as a template, with world file 7_tile_slope.gfw. Set the CRS to EPSG 6575. Note that the image does not align with the map.
1. Alternatively, load the map junior.2019-05-24a.original.omap as a template. Note that the template map does not align with the "deformed" map. (This map is georeferenced with EPSG 6575. The "deformed" map was created by importing this.)

This PR changes the behavior, so that the template image and template map show up aligned with the map.

Regarding #1277 (Determine grid scale automatically) @dg0yt wrote:

> What we need for all type of maps is the reasonable precise placement of geospatial data (templates or import). The junior "deformed" map example shows that Mapper seems to work okay for GPX tracks. Of course we won't achieve satisfying results with rectangular raster images when just the corners are aligned to fit.

In the "deformed" example, the current Mapper does not work okay with the template image. This PR takes care of that limitation.

### DIFF
--- a/src/gui/widgets/template_list_widget.cpp
+++ b/src/gui/widgets/template_list_widget.cpp
@@ -110,7 +110,16 @@ struct ApplyTemplateTransform
 	
 	void operator()(Object* object) const
 	{ 
-		object->scale(transform.template_scale_x, transform.template_scale_y);
+		/// \todo Move this if...else into a constructor that sets up a transform member.
+		if (qFuzzyIsNull(transform.template_shear))
+			object->scale(transform.template_scale_x, transform.template_scale_y);
+		else
+		{
+			QTransform scaling(transform.template_scale_x, transform.template_shear,
+			                   transform.template_shear, transform.template_scale_y,
+			                   0, 0);
+			object->transform(scaling);
+		}
 		object->rotate(transform.template_rotation);
 		object->move(transform.template_x, transform.template_y);
 	}
--- a/src/templates/template.cpp
+++ b/src/templates/template.cpp
@@ -94,14 +94,18 @@ Q_STATIC_ASSERT(std::is_nothrow_move_assignable<TemplateTransform>::value);
 // static
 TemplateTransform TemplateTransform::fromQTransform(const QTransform& qt) noexcept
 {
-	const auto scale_x  = std::hypot(qt.m11(), qt.m12());
-	const auto scale_y  = std::hypot(qt.m21(), qt.m22());
-	const auto rotation = std::atan2(qt.m21() / scale_y, qt.m11() / scale_x);
+	const auto rotation = std::atan2(qt.m21()-qt.m12(), qt.m11()+qt.m22());
+
+	// Determine the non-rotating, pure scaling aspect.
+	// Note TemplateTransform rotation is opposite to QTransform.
+	// This rotation makes the shear components equal.
+	QTransform scaling = qt * (qFuzzyIsNull(rotation) ? QTransform() : QTransform().rotateRadians(rotation));
 	return { qRound(1000 * qt.m31()),
 	         qRound(1000 * qt.m32()),
 	         rotation,
-	         scale_x,
-	         scale_y };
+	         scaling.m11(),
+	         scaling.m22(),
+	         (scaling.m12()+scaling.m21())/2 };
 }
 
 
@@ -114,6 +118,8 @@ void TemplateTransform::save(QXmlStreamWriter& xml, const QString& role) const
 	xml.writeAttribute(QString::fromLatin1("y"), QString::number(template_y));
 	xml.writeAttribute(QString::fromLatin1("scale_x"), QString::number(template_scale_x));
 	xml.writeAttribute(QString::fromLatin1("scale_y"), QString::number(template_scale_y));
+	if (!qFuzzyIsNull(template_shear))
+		xml.writeAttribute(QString::fromLatin1("shear"), QString::number(template_shear));
 	xml.writeAttribute(QString::fromLatin1("rotation"), QString::number(template_rotation));
 	xml.writeEndElement(/*transformation*/);
 }
@@ -130,6 +136,7 @@ void TemplateTransform::load(QXmlStreamReader& xml)
 	template_y = coord.nativeY();
 	template_scale_x = element.attribute<double>(QLatin1String("scale_x"));
 	template_scale_y = element.attribute<double>(QLatin1String("scale_y"));
+	template_shear = element.attribute<double>(QLatin1String("shear")); // 0 if absent
 	template_rotation = element.attribute<double>(QLatin1String("rotation"));
 }
 
@@ -140,7 +147,9 @@ bool operator==(const TemplateTransform& lhs, const TemplateTransform& rhs) noex
 	        && (qFuzzyCompare(lhs.template_rotation, rhs.template_rotation)
 	            || (qIsNull(lhs.template_rotation) && qIsNull(rhs.template_rotation)))
 	        && qFuzzyCompare(lhs.template_scale_x, rhs.template_scale_x)
-	        && qFuzzyCompare(lhs.template_scale_y, rhs.template_scale_y);
+	        && qFuzzyCompare(lhs.template_scale_y, rhs.template_scale_y)
+	        && (qFuzzyCompare(lhs.template_shear, rhs.template_shear)
+	            || (qFuzzyIsNull(lhs.template_shear) && qFuzzyIsNull(rhs.template_shear)));
 }
 
 bool operator!=(const TemplateTransform& lhs, const TemplateTransform& rhs) noexcept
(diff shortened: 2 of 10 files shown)


src/templates/template.cpp
```diff
@@ -94,14 +94,18 @@ Q_STATIC_ASSERT(std::is_nothrow_move_assignable<TemplateTransform>::value);
...
 	return { qRound(1000 * qt.m31()),
-	         qRound(1000 * qt.m32()),
-	         rotation,
-	         scale_x,
-	         scale_y };
+			 qRound(1000 * qt.m32()),
+			 rotation,
+			 scaling.m11(),
+			 scaling.m22(),
+			 (scaling.m12()+scaling.m21())/2 };
```
dg0yt: Wrong indent.

src/templates/template.cpp
```diff
@@ -130,16 +136,23 @@ void TemplateTransform::load(QXmlStreamReader& xml)
 	template_y = coord.nativeY();
 	template_scale_x = element.attribute<double>(QLatin1String("scale_x"));
 	template_scale_y = element.attribute<double>(QLatin1String("scale_y"));
+	if (element.hasAttribute(QLatin1String("shear")))
+		template_shear = element.attribute<double>(QLatin1String("shear"));
+	else
+		template_shear = 0.0;
```
dg0yt: `element.attribute<double>(QLatin1String("shear"))` will return 0 anyway if the attribute doesn't exist. So no `if else` needed here.

src/templates/template.cpp
```diff
@@ -621,8 +634,19 @@ bool Template::trySetTemplateGeoreferenced(bool /*value*/, QWidget* /*dialog_par
 void Template::applyTemplateTransform(QPainter* painter) const
 {
 	painter->translate(transform.template_x / 1000.0, transform.template_y / 1000.0);
+	// Rotate counter-clockwise.
 	painter->rotate(-transform.template_rotation * (180 / M_PI));
-	painter->scale(transform.template_scale_x, transform.template_scale_y);
+
+	// Scale
+	if (qFuzzyIsNull(transform.template_shear))
+		painter->scale(transform.template_scale_x, transform.template_scale_y);
```
dg0yt: Please add parentheses around this line, for symmetry with the `else` branch.

src/templates/template.cpp
```diff
@@ -905,8 +929,11 @@ void Template::applyTemplatePositionOffset()
 {
 	QTransform t;
 	t.rotate(-qRadiansToDegrees(getTemplateRotation()));
-	t.scale(getTemplateScaleX(), getTemplateScaleY());
-	setTemplatePosition(templatePosition() - MapCoord{t.map(QPointF{accounted_offset})});
+	const double shear = getTemplateShear();
+	QTransform scaling(getTemplateScaleX(), shear,
+					   shear, getTemplateScaleY(),
+					   0, 0);
```
dg0yt: Wrong indent.

src/gui/widgets/template_list_widget.cpp
```diff
@@ -110,8 +110,16 @@ struct ApplyTemplateTransform
 	
 	void operator()(Object* object) const
 	{ 
+		if (qFuzzyIsNull(transform.template_shear))
+			object->scale(transform.template_scale_x, transform.template_scale_y);
+		else
+		{
+			QTransform scaling(transform.template_scale_x, transform.template_shear,
+							   transform.template_shear, transform.template_scale_y,
+							   0, 0);
```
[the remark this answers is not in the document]
pkturner: Fixed indent and added `\todo`.

src/templates/template.cpp
pkturner: Fixed indent.

src/templates/template.cpp
pkturner: Added braces.

src/templates/template.cpp
pkturner: Fixed indent.

src/templates/template.cpp
pkturner: Simplified as indicated.

pkturner:
Cleaned up the PR as requested.

dg0yt:
I'm still a bit undecided. Frankly, I didn't want to add more complexity to templates before doing a significant refactoring on the template classes. 

> Cleaned up the PR as requested.

I'm fully satisfied with the code of this PR. Just two more questions.

1. When extending `TemplateTransform`, why limiting ourselves to `m12 == m21`? Is this PR enough to handle a random photograph of an orienteering map, e.g. for analysing GPX tracks? 
(Yes, this what I am doing sometimes, indeed.)
2. At least for a transitional period, how shall template which use this feature be handled in older versions of Mapper? 
With the current code, they will display the template in a misplaced way. As an alternative, they might refuse to display the template (something like an "unknown template type").

I tend to ignore question 1 for now (v0.9), but I would like to discuss question 2.

@lpechacek What it is your opinion?

puzzlepaint:
>     1. When extending `TemplateTransform`, why limiting ourselves to `m12 == m21`? Is this PR enough to handle a random photograph of an orienteering map, e.g. for analysing GPX tracks?
>        (Yes, this what I am doing sometimes, indeed.)

For a random photograph of a map, in general one would need to account for arbitrary lens distortion. In this case, for best results, one would want to calibrate the camera, for example with something like [Kalibr](https://github.com/ethz-asl/kalibr/wiki).

If the map is planar (not bent) and your camera is a perfect pinhole camera, or sufficiently close to it to ignore any lens distortion (which I guess may often be the case for small-field-of-view smartphone cameras), then you would still need a [homography](https://en.wikipedia.org/wiki/Homography_(computer_vision)) to be able to undistort the map in general. This has 8 degrees of freedom, and may be stored in a 3x3 matrix. Seems unlikely that once can constrain `m12 == m21` while not restricting oneself. Often, homographies are normalized by scaling them such that the bottom-right element is one (while I guess only restricting them a little). In that case, one would still need to store all other 8 elements of the matrix.

pkturner:
Agree with @puzzlepaint. Regarding the other part of question 1, the constraint `m12 == m21` means that the average rotation is 0. In the code, rotation is analyzed first, and the remaining matrix is guaranteed to satisfy the constraint, at least fuzzily.

pkturner:
> I'm still a bit undecided. Frankly, I didn't want to add more complexity to templates before doing a significant refactoring on the template classes.

Understood.

When you start on the refactoring, let me know if there is some grunt work that I could help with.

dg0yt:
> 2. At least for a transitional period, how shall template which use this feature be handled in older versions of Mapper?
With the current code, they will display the template in a misplaced way. As an alternative, they might refuse to display the template (something like an "unknown template type").

I could implement the latter, and then merge.

dg0yt:
> > 1. At least for a transitional period, how shall template which use this feature be handled in older versions of Mapper?
> >    With the current code, they will display the template in a misplaced way. As an alternative, they might refuse to display the template (something like an "unknown template type").
> 
> I could implement the latter, and then merge.

I didn't really find a hook how to do help older versions. The best thing we could do is select one of the following:
1. Write an alternativ "legacy" transform (as an approximation without shearing) to the file and ignore this in newer versions.
This adds more legacy stuff handling for a long time. :-1: :-1:
2. Support only reading from files for now, but postpone the creation of such transformations to some minor release.
This adds technical debt and doesn't help testing. :-1:
3. Do nothing, assuming that shearing is close to zero in most cases, and people don't switch back to the old version very often, so the lack of backward compatibility does no harm.
The last option means I merge this now as is, just before a last 'master' release before v0.9.0, with the file format version already incremented to 9, so old version of Mapper show a warning anyway.
This my preferred option.

pkturner:
Either option 2 (for now, support only reading of files containing shear transformations) or option 3 (merge these changes now as is) would be satisfactory. I agree with the preference for the latter.

Issue #1293, opened 2019-07-20
Android: Toolbar buttons very small on some devices


dg0yt:
### Steps to reproduce
Compare toolbar button sizes on various devices. Example:
![android-toolbar-size](https://user-images.githubusercontent.com/13567791/61577589-e35aee80-aae9-11e9-91f0-113ecc1c20fd.jpg)

### Actual behaviour
On some devices (cf. tablet above), the buttons are tiny, less than 5 mm.

### Expected behaviour
- [ ] The button should not be smaller than a certain size.
- [ ] Optionally, the button size could be a configuration option.

### Configuration
Mapper Version: 0.8.4, `dev v201906xx`
Operating System: Android

dg0yt:
Sidenote: The Measure widget title bar and its icons are a separate story, partly visible in #787.

lpechacek:
Ack. Feel free to cherry-pick https://github.com/OpenOrienteering/mapper/commit/df0a41dce3057d2a5877c5e2cd1840aee281cd9d and https://github.com/OpenOrienteering/mapper/commit/64b1f471377de553bd10238036b95592738b2cdf.

Zerbembasqwibo:
Very small toolbar buttons in 0.9.0 now. Android 6.0.1

![image](https://user-images.githubusercontent.com/16168975/67633138-5378cd80-f8ac-11e9-9bc1-6e6ff3b71380.png)

dg0yt:
@Zerbembasqwibo Please go to the settings and verify/adjust the size there.

Zerbembasqwibo:
Size of "Åtgärdsknapp"-value solved it. Great!

Issue #787, opened 2016-10-30
Android: Improper size of window and crash with measure tool


krticka:
### Steps to reproduce
1. Try to measure length of object
2. Result is located in very small window, not readable
3. Move window to upper dock position and try to move window to different position.
### Actual behaviour

Result of measuring is not readable, window is too small. Resize of window does not work. After move of window to upper dock position window is enlarges automatically, result is readable, but application accidentally crashes after move of window back above map field.
![screenshot_2016-10-30-09-21-50 1](https://cloud.githubusercontent.com/assets/15969434/19835354/268ff030-9e84-11e6-968b-baabacde3b14.jpg)
### Expected behaviour

Readable result of measurement, resize of window is working, no crash on window move.
### Configuration

Mapper Version: 0.6.6
Operating System: Android 6.0.1 (Samsung Galaxy Note 4)

dg0yt:
It is a little bit more complicated, and probably outside of our source code.
- The measure widget is a "dock widget". It is initially "floating" (undocked), but it gets docked and resized when you move it to the top (or any other border). (I even contributed a change to Qt to have a titlebar at all, for closing and moving.) There shouldn't be a crash, though.
- The initial size should get some review.
- Basically, resizing does work as well, as it does on the desktop.
- One issue especially on Android devices is insufficient high-dpi support in the version Qt we are using at the moment (Qt 5.5.1). The tiny icons in the widget title bar are an outcome of this. Thus I won't look into this issue before I can make relases based on Qt 5.6 (#673). (I think I can release my local work in the next weeks.)

This issue reminds us of the fact that the Android port of Mapper is still very much a proof-of-concept.

Issue #576, opened 2015-12-31
Track recording when screen is turned off


rugk:
[In your doc](https://github.com/OpenOrienteering/mapper-manual/blob/gh-pages/pages/android-app.md#the-mobile-map-editor) (part 3) you say:

> Note that track recording is interrupted while the screen is off or the app is in the background to save battery. If you want to record a track, you need to keep the app active.

However as far as I remember the app also recorded when the screen is off (which also makes sense IMO).
And if this is not true then I would recommend to add this feature.

dg0yt:
Maybe you test and report the current state?

Note that "screen off" is different from "in the background". The app will be in the background when another app is active. It will need a different approach to make recordings while the app is in the background (i.e. stopped).

rugk:
If it is stopped (swiped out of the "recent apps" list) you can of course not track. But if you leave the app in the foreground and just turn the device off it works AFAIK. So maybe the documentation is wrong here.

yevhenmazur:
I've changed my work device to Mi 9 powered by Android Pie (MIUI 10.2) with internal dual-band GNSS receiver. In this setup track is recorded only when the screen is on. Unfortunately this behavior makes the dual-band GNSS receiver almost useless. Looking for a solution or workaround, I found a [similar issue with Osmand](https://github.com/osmandapp/Osmand/issues/5587) (resolved). I would be very grateful if track recording when the screen is off will be possible on new devices with internal GNSS reciver.

dg0yt:
@yevhenmazur Thanks for pointing this out.  Which OOMapper version were you refering to?
FTR I would like to add the link to https://github.com/osmandapp/Osmand/issues/5632.

yevhenmazur:
Tested on OOMapper 0.8.4 and dev-20190609.3
Tomorrow I going to test it with external Bluetooth GNSS receiver.

dg0yt:
dev-20190609.3 has target API level 28 so it will probably significantly increase our "obligations" on newer devices. This could be hard to solve. (We can move GPX tracking to a service, but I'm more concerned about the app being killed while you are taking photos.) At least, I do have an 8.1 test device.

dg0yt:
Android 9 (API level 28) uses a rigid [power management](API level 28). With a foreground service, an app is considered to be in the *active bucket*.

[...]

PR #1296, opened 2019-07-22
Improve Android user experience

dg0yt:
- Minor build system adjustments
- Fixes for icon and dock widget issues
- Relaxed permission handling on the home screen

Resolves GH-1293.
Closes GH-787.


pkturner:
> Resolves GH-1292.
This looks unintentional, as #1292 is unrelated to this PR.

dg0yt:
> > Resolves [GH-1292](https://github.com/OpenOrienteering/mapper/pull/1292).
> 
> This looks unintentional, as #1292 is unrelated to this PR.

GH-1293, of course. Done.

src/core/storage_location.cpp
```diff
@@ -59,6 +59,11 @@ bool storageAccessGranted()
 	return AppPermissions::checkPermission(AppPermissions::StorageAccess) == AppPermissions::Granted;
 }
 
+/**
+ * Tells Android to update its media scanner cache.
+ * 
+ * This helps to avoid corrupted files when transfering files with MTP.
+ */
```
dg0yt: Already documented in the header file. I will remove this commit.

src/gui/widgets/mapper_proxystyle.cpp
```diff
@@ -36,6 +37,29 @@
...
+namespace  {
+
+/**
+ * This function recognizes dock widget related widgets by the class name
+ * matching "DockWidget".
+ * 
+ * This function helps to customize the style for classes like
+ * "QDockWidgetTitleButton" or "MapEditorDockWidget".
+ */
+bool isDockWidgetRelated(const QWidget* widget)
```
dg0yt: This causes an unused-function warning on all systems but Android. I will have to add `#ifdef Q_OS_ANDROID`

dg0yt: Added `Q_DECL_UNUSED` instead.

lpechacek:
Will do my best to review until end of week.

dg0yt:
Android foreground service is coming, too: 
https://github.com/OpenOrienteering/mapper/compare/dg0yt/android-service

However, it cannot pass CI until the superbuild update to compile SDK >= 26 is integrated. 
https://github.com/OpenOrienteering/superbuild/pull/37

src/gui/main_window.cpp
```diff
@@ -498,7 +499,19 @@ void MainWindow::setHasUnsavedChanges(bool value)
 		has_unsaved_changes = value;
 		setAutosaveNeeded(has_unsaved_changes && !has_autosave_conflict);
 	}
+	else
+	{
```
dg0yt: The following needs more work.

dg0yt:
Thanks a lot for the review.

> Please make code self-modification in "Android: Fix BuildConfig and R access" more visible to a future developer trying to debug that part. [...] Example: […]

Well, this really far from perfect. Guess how long it took me to realize why R and BuildConfig wouldn't work out of the box.
I did consider something like you propose, but then you would need to remember to add these headers to new classes. However, from an Android developer point of view, you would expect them to be available right in the app's namespace, without explicit import. 
So I really would like to come to the proper solution which seems to be "flavours" in gradle, completely eliminating this hack. But it is out of scope for the current PR. The interplay of gradle, androiddeployqt, local setups and CI deserve its own share of time.

> Button for hiding the top Mapper menu bar changes size. It's smaller when the menu is hidden.

Good catch.

> Mapper crashed three times during testing. Android system log contains following trace for all three crashes: [...] 

Maybe this is unrelated to our source code. but a bug in Android >= P, cf. 
https://stackoverflow.com/questions/51956971/illegalstateexception-of-toast-view-on-android-p-preview
even with a google issue link. I can try to create a similar workaround.
Similar:
https://stackoverflow.com/questions/54624116/android-java-lang-illegalstateexception-in-toast-handleshow

dg0yt:
> So I really would like to come to the proper solution which seems to be "flavours" in gradle, completely eliminating this hack.

However, androiddeployqt and/or Qt Creator are not ready for changing the app ID in build.gradle:  
https://bugreports.qt.io/browse/QTCREATORBUG-20802

dg0yt:
>     * Please provide more context for the GitHub issue references in commit messages. The message should be self-contained, therefore understandable  without having to connect to Internet (I do most of my Mapper work offline on train). Example: `This is expected to resolve #576.` -> `This is expected to resolve GH-576 ("Track recording when screen is turned off").`

Done.

>     * Please make code self-modification in "Android: Fix BuildConfig and R access" more visible to a future developer trying to debug that part. Efforts to figure out why there are differences between the source code repository and compile-time sources typically take half day to complete with facepalm and swear words at the end. I'd welcome if code added by the `sed` script were present in the source file and commented out with a remark that a build script activates this code during build. Example: [...]

Created #1306 instead.

>     * Button for hiding the top Mapper menu bar changes size. It's smaller when the menu is hidden.

Done.

>     * Mapper crashed three times during testing. Android system log contains following trace for all three crashes: [...]

Created #1305 instead.

In addition, fixed a crash with large buttons on tiny screens.

src/gui/widgets/editor_settings_page.cpp
```diff
@@ -44,6 +45,12 @@ EditorSettingsPage::EditorSettingsPage(QWidget* parent)
 {
 	auto layout = new QFormLayout(this);
 	
+	if (MainWindow::mobileMode())
+	{
+		button_size = Util::SpinBox::create(1, 3.0, 26.0, tr("mm", "millimeters"), 0.1);
```
dg0yt: @lpechacek How about making this an integer, similar to the following 'Symbol icon size'? It is always a little bit difficult to edit this setting. At least I would like to change the step size from 0.1 to 0.5.

PR #1298, opened 2019-07-26
MainWindow: Start an Android foreground service

### LINKED ISSUE #576: Track recording when screen is turned off

dg0yt:
To protect unsaved map changes and to keep continuity of track
recording when the screen is off or the app is in the background,
let MainWindow start an Android foreground service when the it
is told that the map is modified.
This is expected to resolve #576. However this needs to be verified
in the field, across build and device configurations.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -86,6 +86,8 @@
             <meta-data android:name="android.app.extract_android_style" android:value="default"/>
             <!-- extract android style -->
         </activity>
+        <service android:name="org.openorienteering.mapper.MapperService"
+                 android:exported="false" />
     </application>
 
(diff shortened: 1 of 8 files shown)


yevhenmazur:
Verified in the field: track recording works as expected on my device.

PR #1301, opened 2019-07-31
OCD: Fix string type 1039 (georeferencing) import

lpechacek:
Recently I received an OCAD 12 georeferenced map at scale 1:10000. After opening it in Mapper-dev I could not match the real world coordinates from Mapper to coordinates from other map services. The root cause was a failure to parse the georeferencing string. Mapper failed to parse "10000.000000" and silently filled in "15000" as map scale. This commit fixes the parsing part but not the silent breakage.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -363,7 +363,7 @@ namespace {
 void tryParamConvert(int& out, const QString& param_value)
 {
 	bool ok;
-	auto value = param_value.toInt(&ok);
+	auto value = qRound(param_value.toFloat(&ok));
 	if (ok)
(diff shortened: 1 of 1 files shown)


dg0yt:
I reformatted the commit message to lines of max. 70 characters. There are some contexts where the message is intended (e.g. `git log`) or otherwise limited (e.g. when expanded in the PR conversation tab).

lpechacek:
Good point, Kai. It also reminds me that the last line about debug warning should have been dropped. The first patch version contained some extra code which I dropped before submission. But now that it's in `dev`, let's leave it as is unless you are fine with force updating the branch. Mistakes happen...

Thanks and wish you a good day!

PR #1309, opened 2019-08-01
MapGrid: Fix offset direction

pkturner:
When the Map Grid is enabled, a negative rotation leads to offsets being interpreted in the wrong direction.

To reproduce,

1. Configure the grid to be visible with an additional rotation of -20°, and a vertical offset of 30 meters.
1. The horizontal lines show up offset downwards, whereas a positive vertical offset should move the lines toward the top.
1. The snap-to-grid feature snaps to the _correct_ grid intersections, which do not match the visible horizontal lines.

Alternatively,
1. Set georeferencing such that grivation is negative, and the origin of projected coordinates is offset from the map origin.
1. Projected coordinates of the visible horizontal lines do not match multiples of the vertical spacing.
1. The snap-to-grid feature snaps to the _correct_ grid intersections, which do not match the visible horizontal lines.
***
`hatchingOperation` should recognize the full circle of possible rotation values. Also fix the direction of the quarter-turn when calling `hatchingOperation` for horizontal lines. This provides correct offset to the origin of projected coordinates.

### DIFF
--- a/src/core/map_grid.cpp
+++ b/src/core/map_grid.cpp
@@ -136,8 +136,8 @@ void MapGrid::draw(QPainter* painter, const QRectF& bounding_box, Map* map, qrea
 	if (display == AllLines)
 		Util::gridOperation<ProcessLine>(bounding_box, final_horz_spacing, final_vert_spacing, final_horz_offset, final_vert_offset, final_rotation, process_line);
 	else if (display == HorizontalLines)
-		Util::hatchingOperation<ProcessLine>(bounding_box, final_vert_spacing, final_vert_offset, final_rotation + M_PI / 2, process_line);
-	else // if (display == VeritcalLines)
+		Util::hatchingOperation<ProcessLine>(bounding_box, final_vert_spacing, final_vert_offset, final_rotation - M_PI / 2, process_line);
+	else // if (display == VerticalLines)
 		Util::hatchingOperation<ProcessLine>(bounding_box, final_horz_spacing, final_horz_offset, final_rotation, process_line);
 }
 
--- a/src/util/util.h
+++ b/src/util/util.h
@@ -173,9 +173,14 @@ namespace Util
 	void hatchingOperation(const QRectF& extent, double spacing, double offset, double rotation, T& processor)
 	{
 		// Make rotation unique
-		rotation = fmod(1.0 * rotation, M_PI);
+		rotation = fmod(1.0 * rotation, 2 * M_PI);
 		if (rotation < 0)
-			rotation = M_PI + rotation;
+			rotation += 2 * M_PI;
+		if (rotation >= M_PI)
+		{
+			rotation -= M_PI;
+			offset = -offset;
+		}
 		Q_ASSERT(rotation >= 0 && rotation <= M_PI);
 		
 		if (qAbs(rotation - M_PI/2) < 0.0001)
@@ -187,9 +192,11 @@ namespace Util
 				processor.processLine(QPointF(cur, extent.top()), QPointF(cur, extent.bottom()));
 			}
 		}
-		else if (qAbs(rotation - 0) < 0.0001)
+		else if (rotation < 0.0001 || rotation > M_PI - 0.0001)
 		{
 			// Special case: horizontal lines
+			if (rotation > M_PI/2)
+				offset = -offset;
 			double first = offset + ceil((extent.top() - offset) / (spacing)) * spacing;
 			for (double cur = first; cur < extent.bottom(); cur += spacing)
 			{
@@ -301,7 +308,7 @@ namespace Util
 	                   double horz_offset, double vert_offset, double rotation, T& processor)
 	{
 		hatchingOperation(extent, horz_spacing, horz_offset, rotation, processor);
-		hatchingOperation(extent, vert_spacing, vert_offset, rotation + M_PI / 2, processor);
+		hatchingOperation(extent, vert_spacing, vert_offset, rotation - M_PI / 2, processor);
 	}
 	
 }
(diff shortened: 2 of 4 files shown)


src/core/map_grid.cpp
```diff
@@ -136,7 +136,7 @@ void MapGrid::draw(QPainter* painter, const QRectF& bounding_box, Map* map, qrea
 	if (display == AllLines)
 		Util::gridOperation<ProcessLine>(bounding_box, final_horz_spacing, final_vert_spacing, final_horz_offset, final_vert_offset, final_rotation, process_line);
 	else if (display == HorizontalLines)
-		Util::hatchingOperation<ProcessLine>(bounding_box, final_vert_spacing, final_vert_offset, final_rotation + M_PI / 2, process_line);
+		Util::hatchingOperation<ProcessLine>(bounding_box, final_vert_spacing, final_vert_offset, final_rotation - M_PI / 2, process_line);
 	else // if (display == VeritcalLines)
```
dg0yt: Could you perhaps spell-correct this comment when you edit your original commit?

src/util/util.h
```diff
@@ -173,10 +173,15 @@ namespace Util
 	void hatchingOperation(const QRectF& extent, double spacing, double offset, double rotation, T& processor)
 	{
 		// Make rotation unique
-		rotation = fmod(1.0 * rotation, M_PI);
+		rotation = fmod(1.0 * rotation, 2 * M_PI);
 		if (rotation < 0)
-			rotation = M_PI + rotation;
-		Q_ASSERT(rotation >= 0 && rotation <= M_PI);
+			rotation = 2 * M_PI + rotation;
```
dg0yt: I would prefer to read
~~~
rotation += 2 * M_PI;

src/util/util.h
```diff
@@ -173,10 +173,15 @@ namespace Util
...
+		Q_ASSERT(rotation >= 0 && rotation <= 2 * M_PI);
+		while (rotation > M_PI - 0.0001)
```
dg0yt: I see how it works, but I don't feel like `while` is the best expression here.
- The loop can run exactly zero, one, or two times, given the previous assertion.
- When running two times, it will move rotation out of the previously asserted range.

IIUC this loop can be removed when an extra `|| qAbs(rotation > M_PI - 0.0001)` is added in line 195?

dg0yt: Okay, updated the comment twice, but still it is not correct: Instead of the `while` loop, it still takes an `if`:
~~~
		if (rotation >= M_PI)
		{
			rotation -= M_PI;
			offset = -offset;
		}
		Q_ASSERT(rotation >= 0 && rotation < M_PI);
~~~
and later for horizontal lines:
~~~
		else if (rotation < 0.0001 || rotation > M_PI - 0.0001)

pkturner: @dg0yt Yes, the `while` was tricky. It would be more straightforward to do it as you say. With the 
`		else if (rotation < 0.0001 || rotation > M_PI - 0.0001)`
that you suggest, it would require another
`		if (rotation > M_PI / 2) offset = -offset;`
which is what the trickiness avoided. Straightforward is good.

pkturner:
The test data looks good as far as it goes. I experimented with adding a few tests of `hatchingOperation` with rotations of 37° and -37° to catch the general case.

dg0yt:
> The test data looks good as far as it goes. I experimented with adding a few tests of `hatchingOperation` with rotations of 37° and -37° to catch the general case.

:+1:  The biggest part of hatchingOperation was not covered yet. I hope there is a set of nearly integer test parameters for the general case, or maybe you could precalculate an offset which gives integer line coordinates.

pkturner:
> It is really interesting which bugs you discover, 

I've been playing with ideas related to the georeferencing dialog. I was looking at how the reference point gets involved in the MapGrid calculation, and noticed that the grid line did not have the expected value of Northing.

Finding the bug among the dozen or so plus and minus signs which could have thrown off that grid line, was good exercise for my retired mind.

pkturner:
> I hope there is a set of nearly integer test parameters for the general case, or maybe you could precalculate an offset which gives integer line coordinates.

I defined `rot37 = atan(0.75) * 180 / M_PI` in order to get an angle which fits a 3,4,5 right triangle and gives line coordinates that have some exact decimals. With a different bounding box, all the line coordinates could be integers.

dg0yt:
> ith a different bounding box, all the line coordinates could be integers.

Feel free to make adjustments to that commit. I was glad that it was testable at all.

src/core/map_grid.cpp
pkturner: OK

src/util/util.h
pkturner: Done.

src/util/util.h
pkturner: Done.

pkturner:
@dg0yt Let me know if you'd like this rebased into 1 commit.

lpechacek:
> What I would do is git reset HEAD\~1, then update the unit test commit, then git stash, then go through git rebase -i HEAD~2 for editing the first commit (i.e. applying the stash).

Given that in this case changes are separated into different source code files, the following procedure is perhaps more straightforward:
```
$ git reset HEAD~1
Unstaged changes after reset:
M	src/core/map_grid.cpp
M	src/util/util.h
M	test/grid_t.cpp
$ git commit --fixup HEAD~1 src/core/map_grid.cpp src/util/util.h
[grid-align c97ace3e8e08] fixup! MapGrid: Fix offset direction
 2 files changed, 8 insertions(+), 6 deletions(-)
$ git commit --fixup HEAD~1 test/grid_t.cpp
[grid-align 299ecdf34848] fixup! GridTest: Test hatching operations
 1 file changed, 50 insertions(+), 22 deletions(-)
$ git rebase -i --autosquash HEAD~4
Successfully rebased and updated refs/heads/grid-align.
```

Just my 0.02€.

pkturner:
Thanks @lpechacek and @dg0yt. That's an eye opener. The ability of git rebase to reorder changes is something I didn't expect.

Issue #1305, opened 2019-08-01
Crash related to Android toasts


dg0yt:
### Steps to reproduce
On a device running Android 8 or 9, either:
- Trigger actions which cause quick changes of the Android toast messages, or
- insert a `toast.cancel()` before `toast.show()` in `onTick()` of MapperActivity.java.

### Actual behaviour
The app sooner or later crashes:
~~~
D Surface : Surface::disconnect(this=0x90499000,api=1)
D GraphicBuffer: unregister, handle(0x8f764a40) (w:52 h:87 s:64 f:0x1 u:f02)
D View    : [Warning] assignParent to null: this = android.widget.LinearLayout{b29ac2e V.E...... ......ID 0,0-87,52}
D WindowClient: Remove from mViews: android.widget.LinearLayout{b29ac2e V.E...... ......ID 0,0-87,52}, this = android.view.WindowManagerGlobal@4ec0af7
D WindowClient: Add to mViews: android.widget.LinearLayout{b29ac2e V.E...... ......ID 0,0-87,52}, this = android.view.WindowManagerGlobal@4ec0af7
D ViewRootImpl[Toast]: hardware acceleration = true , fakeHwAccelerated = false, sRendererDisabled = false, forceHwAccelerated = false, sSystemRendererDisabled = false
D AndroidRuntime: Shutting down VM
E AndroidRuntime: FATAL EXCEPTION: main
E AndroidRuntime: Process: org.openorienteering.mapper.dev, PID: 8060
E AndroidRuntime: java.lang.IllegalStateException: View android.widget.LinearLayout{b29ac2e V.E...... ......ID 0,0-87,52} has already been added to the window manager.
E AndroidRuntime: 	at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:328)
E AndroidRuntime: 	at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:93)
E AndroidRuntime: 	at android.widget.Toast$TN.handleShow(Toast.java:496)
E AndroidRuntime: 	at android.widget.Toast$TN$1.handleMessage(Toast.java:400)
E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:164)
E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6523)
E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #1318, opened 2019-08-08
Replace Android system Toast

dg0yt:
Fix GH-1305 by own implementation.

### DIFF
--- a/src/gui/widgets/toast.h
+++ b/src/gui/widgets/toast.h
@@ -1,0 +1,141 @@
+/*
+ *    Copyright 2012, 2013 Thomas Schöps
+ *    Copyright 2017, 2019 Kai Pastor
+ *
+ *    This file is part of OpenOrienteering.
+ *
+ *    OpenOrienteering is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


dg0yt:
@lpechacek  I asked for review, but I hope it is okay if I merge this tonight anyway. I want to quickly create a new dev preview release for Android which fixes the two biggest show stoppers for completing v0.9.0.
Oh, and I have to solve merge conflict.

lpechacek:
Apologies for delayed response. I'm on vacation with limited Internet access and while I've noticed the patch set, I missed the review request.

I glanced at the patches and they look good to me. I've noticed only one nit but that's not a blocker for merge, which has happened already.

src/gui/widgets/toast.h
```diff
@@ -0,0 +1,141 @@
...
+public:
+	/**
+	 * Constructs a new Toast.
+	 */
+	Toast(QWidget* parent = nullptr);
+	
+	/**
+	 * Shows a Toast.
+	 * 
+	 * @param region   The global region of interest. The Toast will be placed
```
lpechacek: `showText()` does not have region parameter. But I think Doxygen can deal with the extra line.

dg0yt: Thanks!

Issue #1325, opened 2019-08-14
Minor issues with dev v20190813.2


mlerjen:
### Issue 1:
1. Opening a geo-referenced file in dev v20190813.2 the gps position is wrong. (Opening the same file in 0.8.4. the gps position is right.)
![2019-08-14-10-36-14](https://user-images.githubusercontent.com/15611653/63059757-8e47a800-bef0-11e9-994f-893b687f71ba.png)
### Issue 2: 
2. Opening a file which was - saved with the option "hide all templates" -  in dev v20190813.2 all files not checked as visible are "not found" and cannot be loaded later. 
![2019-08-14-09-42-53](https://user-images.githubusercontent.com/15611653/63059767-94d61f80-bef0-11e9-8a8f-0a7d96863f7b.png)
### Configuration
Mapper Version: dev v20190813.2
Operating System: Android 4

dg0yt:
> Opening a geo-referenced file in dev v20190813.2 the gps position is wrong. (Opening the same file in 0.8.4. the gps position is right.)

Which was the georeferencing and (approximate) actual position in this moment? Now using PROJ library version 6, things could get *more accurate*, but we might still be using the library wrongly with regard to some CRS.

> Opening a file which was - saved with the option "hide all templates" - in dev v20190813.2 all files not checked as visible are "not found" and cannot be loaded later.

I could not reproduce this on the desktop now. Are the "lost" templates stored in the same directory tree or type of storage? Mapper asks for permissions a little bit later now, and if it doesn't need to ask for the map, it won't have the permissions for the tempates later.

dg0yt:
> > Opening a file which was - saved with the option "hide all templates" - in dev v20190813.2 all files not checked as visible are "not found" and cannot be loaded later.
> 
> I could not reproduce this on the desktop now.

And also not on Android 4.4 where permissions are always granted in advance

lpechacek:
> Issue 2:

>    Opening a file which was - saved with the option "hide all templates" - in dev v20190813.2 all 
> files not checked as visible are "not found" and cannot be loaded later.

I also face this bug. The workaround is to make all templates visible before moving the map file to Android device (Android 7). In reverse direction when I open file saved on Android tablet on the desktop version, templates cannot be found again. However, after saving the file on desktop and opening it again makes templates appear. I'm using the `OOMapper` location on internal storage for map files.

dg0yt:
I see. There is some tricky code around handling absolute and relative paths.

[...]

PR #1328, opened 2019-08-16
Update template lookup

dg0yt:


dg0yt:
The first commit is the minimal fix for #1325 subissue 2. 

The remaining commits are some more changes around this code. Since this area is not covered well by unit tests, I would be okay with leaving them out if considered too risky for v0.9.

dg0yt:
Couldn't resist to make a few more changes, but I will leave it alone now, if it passes CI.

dg0yt:
I wouldn't say that I have an easy time when I touch any of this legacy code, but obviously I have to do bug fixes.
I don't have a GH issue for refactoring the `Template` mess, but there is a [card in the Refactoring project](https://github.com/OpenOrienteering/mapper/projects/6#card-13986038). Feel free to add some points.

lpechacek:
I can see that I wrote too long comment and my main point drowned in noise. For the sake of future rewrite, I'd be happy to see improved commit message in d5243f0 explaining why template state becomes invalid and why it is safe to switch from `Invalid` to `Unloaded`. If this knowledge does not exist, the commit message should state it too. Thanks!

dg0yt:
The re-assessment is very much appreciated. While reviewing the documentation of what `tryToFindTemplateFile()` does and why, I was able to spot more subtleties in documenation and behaviours, so there is even one more commit now.

lpechacek:
The commit message nicely captures template state semantics. Thanks!

PR #1330, opened 2019-08-18
Basic GeoTIFF support

dg0yt:
Use GDAL raster support at least for finding georeferencing information. This could handle more than GeoTIFF, but only the latter was tested


lpechacek:
Quickly read through and looks good to me. Pretty cool patchset. Thanks!

PR #1334, opened 2019-08-21
Modernize print support

dg0yt:


src/core/map_printer.cpp
```diff
@@ -1401,3 +1401,5 @@ void MapPrinter::cancelPrintMap()
 
 
 }  // namespace OpenOrienteering
+
+#include "moc_map_printer.cpp"
```
lpechacek: Why is this change needed?

dg0yt: The effect is: The moc output for the class is compiled together with the class, instead of together with all other loose moc output files. So the compiler sees code which belong to the same class at the same time. This might allow some improvements in code generation or compilation speed. 

However, without evidence, it may belong to the field of premature optimization. I tried to find a reference before committing, but failed.

dg0yt: I will remove this. It is done in some Qt source files, but I can't find any recommendation to include moc_*.cpp. (This is different from *.moc files, which are generated for Q_OBJECTs defined in cpp files.)

dg0yt: Now I am able to cite the reasoning:
https://lists.qt-project.org/pipermail/interest/2019-December/034237.html

- "reduces the number of compiler runs" and thus the overall compile time "as the Qt headers need to be processed only once" - This is not relevant for our CMake builds because all automoc files are compiled in a single compiler run, but [at the expense](https://lists.qt-project.org/pipermail/interest/2019-December/034249.html) of:
- "produces slightly better code overall"

And there is [a tool for conversion to this style](https://lists.qt-project.org/pipermail/interest/2019-December/034251.html).

PR #1336, opened 2019-08-24
Sensors library, Windows location, touch mode

dg0yt:


dg0yt:
I found it necessary to reorganize the sensors part and the mobile mode handling. I hope commit and comments are clear enough. Future room for improvement:
- Improve unit test with fake data.
- Report missing requirements (Powershell, .NET Framework) to the user.
- Apply the name change from `mobile_mode` to `touch_mode_enabled` to MapEditorController and its components.

Issue #1345, opened 2019-09-04
Enhancement: corner point editing


lpechacek:
### Problem statement
It is currently possible to establish corner points on a smooth curve by moving curve handles while holding Shift key down. In mobile version Snap button serves the same purpose. The functionality looks appropriate at first glance. However, in practice editing of corner points involves following steps: choose point for editing - look at the point and **realize** that it is a corner point - press Shift key / activate Snap button - make the intended handle adjustment - release Shift key / deactivate Snap button. After a few hours of field work the editing operation feels somewhat cumbersome as forgetting to activate Shift/Snap results in unintended move of the other curve handle and additional operations to rectify the erroneous step.

### Proposal
Neighboring curve handles are independent at corner point as long as their angle is < 180°. When the angle between handles reaches value close to 180° during editing the handles "click in" and the point turns into normal (smooth) point. Then corner point editing operations would work as follows:
1) Create corner point - hold down Shift / activate Snap, move handle
2) Adjust one of the curves meeting at corner point - move handle
3) Turn inner (convex) corner into outer (concave) one and vice versa - move handle while holding Shift / Snap
4) Remove corner point - align handles in straight line during editing, they start moving together

lpechacek:
First implementation illustrating what I mean with the above proposal. Feedback on the new functionality is welcome.

jmacura:
Now when I can see it on the video, I think it's quite cool and elegant! Thumb up.

lpechacek:
As an afterthought... on a dashed line, do we want to make the newly created corner point a dash point?

PR #1349, opened 2019-09-10
Corner point "click in" editing mode

lpechacek:
Implementation of GH-1345 enhancement.
out-1_hMSkB5v9_knQo ogv

### DIFF
--- a/src/core/objects/object_mover.cpp
+++ b/src/core/objects/object_mover.cpp
@@ -34,7 +34,7 @@
 namespace OpenOrienteering {
 
 ObjectMover::ObjectMover(Map* map, const MapCoordF& start_pos)
- : start_position(start_pos), corner_tolerance(0), prev_drag_x(0), prev_drag_y(0), constraints_calculated(true)
+ : start_position(start_pos)
 {
 	Q_UNUSED(map);
 }
--- a/src/core/objects/object_mover.cpp
+++ b/src/core/objects/object_mover.cpp
@@ -268,6 +268,25 @@ void ObjectMover::calculateConstraints()
 		// Determine opposite handle constraints.
 		for (auto index : handles)
 		{
+			auto create_constraint = [this](const PathObject* path,
+			                         const MapCoordVector::size_type moved_index,
+			                         const MapCoordVector::size_type anchor_index,
+			                         const MapCoordVector::size_type opposite_index)
+			{
+				OppositeHandleConstraint constraint;
+				constraint.object = const_cast<PathObject*>(path);  // = item.first
+				constraint.moved_handle_index = moved_index;
+				constraint.curve_anchor_index = anchor_index;
+				constraint.opposite_handle_index = opposite_index;
+				const auto handle1 = path->getCoordinate(constraint.moved_handle_index);
+				const auto handle2 = path->getCoordinate(constraint.opposite_handle_index);
+				const auto anchor_point = path->getCoordinate(constraint.curve_anchor_index);
+				constraint.anchor_is_corner = Util::pointsFormCorner(handle1, anchor_point, handle2, corner_tolerance);
+				constraint.opposite_handle_original_position = handle2;
+				constraint.opposite_handle_dist = handle2.distanceTo(anchor_point);		
+				return constraint;
+			};
+			
 			Q_ASSERT(path->isCurveHandle(index));
 			auto& part = *path->findPartForIndex(index);
 			auto end_index = part.last_index;
@@ -284,17 +303,8 @@ void ObjectMover::calculateConstraints()
 				if (curve_anchor_index != part.first_index &&
 				    path->getCoordinate(part.prevCoordIndex(curve_anchor_index)).isCurveStart())
 				{
-					OppositeHandleConstraint constraint;
-					constraint.object = const_cast<PathObject*>(path);  // = item.first
-					constraint.moved_handle_index = index;
-					constraint.curve_anchor_index = index - 1;
-					constraint.opposite_handle_index = curve_anchor_index - 1;
-					const auto handle1 = path->getCoordinate(constraint.moved_handle_index);
-					const auto handle2 = path->getCoordinate(constraint.opposite_handle_index);
-					const auto anchor_point = path->getCoordinate(constraint.curve_anchor_index);
-					constraint.anchor_is_corner = Util::pointsFormCorner(handle1, anchor_point, handle2, corner_tolerance);
-					constraint.opposite_handle_original_position = handle2;
-					constraint.opposite_handle_dist = handle2.distanceTo(anchor_point);
+					auto constraint = create_constraint(path, index, index - 1,
+					                                    curve_anchor_index - 1);
 					handle_constraints.push_back(constraint);
 				}
 			}
@@ -308,17 +318,9 @@ void ObjectMover::calculateConstraints()
 				if (curve_anchor_index != end_index &&
 				    path->getCoordinate(curve_anchor_index).isCurveStart())
 				{
-					OppositeHandleConstraint constraint;
-					constraint.object = const_cast<PathObject*>(path);  // = item.first
-					constraint.moved_handle_index = index;
-					constraint.curve_anchor_index = curve_anchor_index;
-					constraint.opposite_handle_index = curve_anchor_index + 1;
-					const auto handle1 = path->getCoordinate(constraint.moved_handle_index);
-					const auto handle2 = path->getCoordinate(constraint.opposite_handle_index);
-					const auto anchor_point = path->getCoordinate(constraint.curve_anchor_index);
-					constraint.anchor_is_corner = Util::pointsFormCorner(handle1, anchor_point, handle2, corner_tolerance);
-					constraint.opposite_handle_original_position = handle2;
-					constraint.opposite_handle_dist = handle2.distanceTo(anchor_point);
+					auto constraint = create_constraint(path, index,
+					                                    curve_anchor_index,
+					                                    curve_anchor_index + 1);
 					handle_constraints.push_back(constraint);
 				}
 			}
--- a/src/core/objects/object_mover.cpp
+++ b/src/core/objects/object_mover.cpp
@@ -28,12 +28,13 @@
 #include "core/objects/text_object.h"
 #include "core/symbols/symbol.h"
 #include "core/symbols/text_symbol.h"
+#include "util/util.h"
 
 
 namespace OpenOrienteering {
 
 ObjectMover::ObjectMover(Map* map, const MapCoordF& start_pos)
- : start_position(start_pos), prev_drag_x(0), prev_drag_y(0), constraints_calculated(true)
+ : start_position(start_pos), corner_tolerance(0), prev_drag_x(0), prev_drag_y(0), constraints_calculated(true)
 {
 	Q_UNUSED(map);
 }
@@ -43,6 +44,12 @@ ObjectMover::ObjectMover(Map* map, const MapCoordF& start_pos)
 void ObjectMover::setStartPos(const MapCoordF& start_pos)
 {
 	this->start_position = start_pos;
+}
+
+
+void ObjectMover::setCornerTolerance(const qreal corner_tolerance)
+{
+	this->corner_tolerance = corner_tolerance;
 }
 
 
@@ -86,7 +93,7 @@ void ObjectMover::addTextHandle(TextObject* text, MapCoordVector::size_type hand
 }
 
 
-void ObjectMover::move(const MapCoordF& cursor_pos, bool move_opposite_handles, qint32* out_dx, qint32* out_dy)
+void ObjectMover::move(const MapCoordF& cursor_pos, HandleOpMode move_opposite_handles, qint32* out_dx, qint32* out_dy)
 {
 	auto delta_x = qRound(1000 * (cursor_pos.x() - start_position.x())) - prev_drag_x;
 	auto delta_y = qRound(1000 * (cursor_pos.y() - start_position.y())) - prev_drag_y;
@@ -102,7 +109,7 @@ void ObjectMover::move(const MapCoordF& cursor_pos, bool move_opposite_handles, 
 }
 
 
-void ObjectMover::move(qint32 dx, qint32 dy, bool move_opposite_handles)
+void ObjectMover::move(qint32 dx, qint32 dy, HandleOpMode move_opposite_handles)
 {
 	calculateConstraints();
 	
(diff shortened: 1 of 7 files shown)


src/core/objects/object_mover.cpp
```diff
@@ -33,7 +33,11 @@
 namespace OpenOrienteering {
 
 ObjectMover::ObjectMover(Map* map, const MapCoordF& start_pos)
- : start_position(start_pos), prev_drag_x(0), prev_drag_y(0), constraints_calculated(true)
+ : start_position(start_pos)
+ , corner_tolerance(0)
+ , prev_drag_x(0)
+ , prev_drag_y(0)
+ , constraints_calculated(true)
```
dg0yt: I would prefer to move the constant initializations to the member variable declaration.

src/core/objects/object_mover.cpp
```diff
@@ -102,7 +111,30 @@ void ObjectMover::move(const MapCoordF& cursor_pos, bool move_opposite_handles,
 }
 
 
-void ObjectMover::move(qint32 dx, qint32 dy, bool move_opposite_handles)
+namespace {
+bool forms_a_corner(const MapCoord anchor_point, const MapCoord moved_handle,
+                    const MapCoord opposite_handle, const qreal quantum_size)
```
dg0yt: `forms_a_corner` does not conform to our spelling of function names.

I'm not so happy with having this function between the ObjectMover member function definitions, but I wonder what is the best place/namespace to put this function to. It might be a candidate for a static function in MapCoord, or at least a free function in map_coord.h. I realized that functions in the anonymous namespace are unaccessible to unit tests, too...

src/core/objects/object_mover.cpp
```diff
@@ -124,19 +156,35 @@ void ObjectMover::move(qint32 dx, qint32 dy, bool move_opposite_handles)
...
+		MapCoord anchor_point = constraint.object->getCoordinate(constraint.curve_anchor_index);
+		MapCoord moved_handle = constraint.object->getCoordinate(constraint.moved_handle_index);
+		MapCoordF to_hover_point = MapCoordF(moved_handle - anchor_point);
+		to_hover_point.normalize();
+		
+		MapCoord control = constraint.object->getCoordinate(constraint.opposite_handle_index);
+
+		// Check conditions for "corner point click-in" operation
+		if (move_opposite_handles == HandleOpMode::Click
+			&& constraint.anchor_is_corner)
```
dg0yt: Please align `&&` with the `move_...` in the line above.

src/core/objects/object_mover.cpp
```diff
@@ -124,19 +156,35 @@ void ObjectMover::move(qint32 dx, qint32 dy, bool move_opposite_handles)
...
-			MapCoordF to_hover_point = MapCoordF(constraint.object->getCoordinate(constraint.moved_handle_index) - anchor_point);
-			to_hover_point.normalize();
-			
-			MapCoord control = constraint.object->getCoordinate(constraint.opposite_handle_index);
-			control.setX(anchor_point.x() - constraint.opposite_handle_dist * to_hover_point.x());
-			control.setY(anchor_point.y() - constraint.opposite_handle_dist * to_hover_point.y());
-			constraint.object->setCoordinate(constraint.opposite_handle_index, control);
+			if (forms_a_corner(anchor_point, moved_handle, control, corner_tolerance))
+				break; // still is a corner point -> refrain from moving opposite handle
+			else
```
dg0yt: `else` after `break` is redundant.

src/core/objects/object_mover.h
```diff
@@ -52,6 +52,15 @@ class ObjectMover
...
 	void setStartPos(const MapCoordF& start_pos);
 	
+	/** Sets corner point tolerance. A curve anchor point is considered a corner
+	 * point during editing when the moving curve handle is more than
+	 * corner_tolerance away from the direction line set by the opposite handle.
+	 * 
+	 * @param corner_tolerance Maximum difference in vector directions
+	 *                         in millimeters.
+	 */
+	void setCornerTolerance(const qreal corner_tolerance);
```
dg0yt: `const` is fine in the definition, but has no effect and shall be omitted in the declaration.

src/core/objects/object_mover.h
```diff
@@ -64,16 +73,24 @@ class ObjectMover
 	/** Adds a text handle to the set of elements to move. */
 	void addTextHandle(TextObject* text, MapCoordVector::size_type handle);
 	
+	enum HandleOpMode {
+		Never,
+		Click,
+		Always
+	};
+
```
dg0yt: I would prefer to have the enum at the beginning of the class definition.

dg0yt: I think that this enum and its members need documentation.

src/core/objects/object_mover.cpp
lpechacek: Me too. Assuming that you are talking about all the data members, I'll rather move the refactoring to a separate patch so that I don't mix unrelated changes into the feature implementation.

src/core/objects/object_mover.cpp
lpechacek: Ack. I originally thought I'll reuse the function in OCD export code for corner point generation but later my thoughts went in different direction. I'll take another at the function naming and placement.

src/core/objects/object_mover.cpp
lpechacek: Well, they are aligned with 4-space-sized tab. :)

However, looking again at this part I saw that the mode handling is somewhat confusing in its current form so I rewrote this part into a switch with fallthroughs, which is structuring the code in a more readable manner.

src/core/objects/object_mover.cpp
lpechacek: Ack.

src/core/objects/object_mover.h
lpechacek: Ack.

src/core/objects/object_mover.h
lpechacek: Indeed. The `enum` placement is somewhat random. Will fix.

lpechacek:
Thanks for the review comments, Kai. I'll post updated patch set ASAP.

dg0yt:
:clock1130: Kindly pinging ;-)

lpechacek:
😀 yes, this pull request is not forgotten. It was just a conference, sickness and next SUSE Linux release planning which got into the way. Apologies for the slow progress and thanks for the ping.

src/core/objects/object_mover.cpp
lpechacek: This part has been reimplemented as a `switch`.

src/core/objects/object_mover.cpp
lpechacek: This is now `Util::pointsFormCorner()`.

lpechacek:
Changes in this iteration:
* Rebased on top of a newer `master`.
* The `if-else` maze was rewritten as a switch.
* Following [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), `HandleOpMode::Always` was dropped.
* `HandleOpMode` values have been documented.
* Corner point test function has been renamed, moved to `Util` class and documented.
* All refactoring changes are separated from the function change.

Items for discussion:
* Point coordinates are passed to `Util::pointsFormCorner()` by reference, following `isPointOnSegment()` example. While this approach saves compile time, it perhaps induces pointer access in run time. Wouldn't pass-by-value be more efficient at the expense of including `map_coord.h` in `util.h`?

lpechacek:
> Point coordinates are passed to Util::pointsFormCorner() by reference, ...

So I looked at function assembly code and compared it with the case when coordinates are passed by value. Without analyzing the very details, I would say the machine code is very similar in both cases on x86_64 and I don't expect significant impact from the reference passing. Of course, the ultimate answer can be provided only through performance measurements but we are not designing a latency critical inner loop here, are we? That said, I consider this point a non-issue from performance point of view and I'm fine if it is ignored.

dg0yt:
> > Point coordinates are passed to Util::pointsFormCorner() by reference, ...
> 
> So I looked at function assembly code and compared it with the case when coordinates are passed by value. Without analyzing the very details, I would say the machine code is very similar in both cases on x86_64 and I don't expect significant impact from the reference passing. Of course, the ultimate answer can be provided only through performance measurements but we are not designing a latency critical inner loop here, are we? That said, I consider this point a non-issue from performance point of view and I'm fine if it is ignored.

Qt passes QPoint by const reference, e.g. [https://doc.qt.io/qt-5/qpainter.html#drawEllipse-4](in QPainter::drawEllipse], and I made Mapper follow this standard for MapCoord and friends a year ago, https://github.com/OpenOrienteering/mapper/compare/35526b9...37bedd84

src/util/util.h
```diff
@@ -201,6 +202,25 @@ void gridOperation(const QRectF& extent, double horz_spacing, double vert_spacin
...
+ * @param point1 First anchor_point neighbor.
+ * @param point2 Second anchor_point neighbor.
+ * @param quantum_size How far can point2 lie from the line defined by
+ *        point1--anchor to be considered unaligned with the two other points.
+ * @return True if point2 is further than quantum_size from the line defined 
+ *         by point1--anchor_point or point2 is on the same side of 
+ *         anchor_point as point1.
+ */
+bool pointsFormCorner(const MapCoord& anchor_point, const MapCoord& point1,
+                      const MapCoord& point2, const qreal quantum_size);
```
dg0yt: - There is the `const qreal` in the declaration again ([readability-avoid-const-params-in-decls](https://releases.llvm.org/8.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-avoid-const-params-in-decls.html)).
- I would appreciate a short unit test for this function at least, in `util_t.cpp`.
- I'm concerned that the MapCoord parameter order is not so obvious when looking at a calling line. How about this signature:
~~~
bool pointsFormCorner(const MapCoord& first_point, const MapCoord& inner_point,
                      const MapCoord& last_point, const qreal quantum_size);
~~~
where `inner_point` takes the role of the former `anchor_point`?

src/core/objects/object_mover.cpp
```diff
@@ -124,18 +131,31 @@ void ObjectMover::move(qint32 dx, qint32 dy, bool move_opposite_handles)
...
 	
 	// Apply handle constraints
-	if (move_opposite_handles)
+	for (auto& constraint : handle_constraints)
 	{
-		for (auto& constraint : handle_constraints)
+		MapCoord anchor_point = constraint.object->getCoordinate(constraint.curve_anchor_index);
+		MapCoord moved_handle = constraint.object->getCoordinate(constraint.moved_handle_index);
+		MapCoordF to_hover_point = MapCoordF(moved_handle - anchor_point);
+		MapCoord control = constraint.object->getCoordinate(constraint.opposite_handle_index);
```
dg0yt: Now looking at this again, I wonder if we can't use a better name than `control` here, e.g. `opposite_handle`.

src/util/util.h
lpechacek: Thanks, Kai, for the quick review. I can see you started using clang-tidy. I looked into making the clang helpers work with openSUSE yesterday but it looks like there is something broken in the tidy plugin handling. Will try harder to make it work.
I'm all fine with `const` removal from the declaration. I'll watch for this error next time. Regarding the parameter ordering for `Util::pointsFormCorner()`, the rename brought a very slight shift in the meaning which escaped my attention. With the current function name it makes sense to order them as "first, middle, last". The original meaning was "(test that a point) forms a corner" and the order was "point to be tested, the other two points".
Ad the test for `Util::pointsFormCorner()`, I'll add a patch to the series.

src/core/objects/object_mover.cpp
lpechacek: Well, from my point of view the rename is out of scope of this change. The task is to implement corner click-in editing mode and this is the original code that gets rearranged towards the desired goal. By default I don't change the original code more than necessary. That said, I don't have strong opinion about the naming and I'm fine with either leaving it as is, which is expressed by this patch set, or renaming it to obviously more logical `opposite_handle`.

lpechacek:
> Let me know if I you want to leave this to me.

I don't object to maintainer edits and I believe they will serve the purpose of smooth cooperation. After the years I'm still unable to get all the coding details right and related pull request conversations tend to be lengthy. Feel free to implement small adjustments on your own. I'd appreciate if the changes were mentioned in the commit message like in [this example](https://kernel.suse.com/cgit/kernel-livepatch/commit/?h=SLE15_Update_15&id=f842fd53a604965a864d3827ab90c691b44cb9a6) (note the remarks in square brackets) but it is not an absolute requirement from my side.

src/util/util.h
dg0yt: I was regularly using iwyu and clang-tidy, for the files white listed in code-check-wrapper. But unfortunately I didn't realize that the '.clang-tidy' was ignored due to wrong syntax.

src/core/objects/object_mover.cpp
dg0yt: I know that the name comes from the original code. I just can't skip taking such a note when the unit is touched anyway. It can go to a separate commit, but I wouldn't want to enforce a separate PR, with all the CI cycles.

dg0yt:
> > Let me know if I you want to leave this to me.
> 
> I don't object to maintainer edits and I believe they will serve the purpose of smooth cooperation. After the years I'm still unable to get all the coding details right and related pull request conversations tend to be lengthy. Feel free to implement small adjustments on your own. I'd appreciate if the changes were mentioned in the commit message like in [this example](https://kernel.suse.com/cgit/kernel-livepatch/commit/?h=SLE15_Update_15&id=f842fd53a604965a864d3827ab90c691b44cb9a6) (note the remarks in square brackets) but it is not an absolute requirement from my side.

I try to learn from the feedback.
However, we will always need some signaling for handing over, to avoid duplicate work, and to not block other work in the same units.

lpechacek:
`Util::pointsFormCorner()` test added. I used the data driven test format. Let me know if that is fine with you or you have a different idea about the test procedure. Please note that I haven't yet checked the test coverage.
I'll respond to your comments later throughout the day.

src/core/objects/object_mover.cpp
lpechacek: Fine with me and I don't object incremental refactoring. IMO all these changes can easily go into the last patch, now labeled "Use in-class member initializers ".

Talking about these exact numbers, why stop at the rename? Mapper Coding conventions encourage use of AAA. Guidelines add ES.23 and Con.4. Opinions?
```suggestion
		const auto anchor_point = constraint.object->getCoordinate(constraint.curve_anchor_index);
		const auto moved_handle = constraint.object->getCoordinate(constraint.moved_handle_index);
		auto opposite_handle = constraint.object->getCoordinate(constraint.opposite_handle_index);
		auto to_hover_point = MapCoordF { moved_handle - anchor_point };
```

dg0yt:
> `Util::pointsFormCorner()` test added. I used the data driven test format. Let me know if that is fine with you or you have a different idea about the test procedure.

:+1: Looks very good to me.

src/util/util.h
lpechacek: So let's make a move on this one. I've incorporated the drop of `const` on function prototype and parameter reordering for `Util::pointsFormCorner()`.
I've also taken another look at `code-check-wrapper.sh` and maybe I've guessed how to use it. In any case, I'd be happy to see instructions for use in developer wiki if the check output is important part of review.

src/core/objects/object_mover.cpp
lpechacek: Accepting the suggestion would break the code as it renames only the variable declaration but not the subsequent use. :)
Thanks for clarification on the expected extent of refactoring. For now I've just made the rename as part of the feature implementation.

lpechacek:
Changes in this iteration:
* Drop unnecessary `const` from function declarations.
* Reorder `Util::pointsFormCorner()` parameters.
* Add `Util::pointsFormCorner()` test.

PR #1356, opened 2019-09-16
Preparing v0.9.0: Merge dev into master

dg0yt:


dg0yt:
Default branch changed to master.

lpechacek:
> Default branch changed to master.

Not sure about meaning of the remark. Does it mean that `dev` is no longer target for pull requests or something else?

dg0yt:
> > Default branch changed to master.
> 
> Not sure about meaning of the remark. Does it mean that `dev` is no longer target for pull requests or something else?

Yes, the default branch is the default for pull requests, and also for checkouts IIUC.  
https://help.github.com/en/articles/setting-the-default-branch

PR #1363, opened 2019-09-19
ISOM 2017-2

dg0yt:
Attempt to eventually merge the work of @Symbian9 and @yevhenmazur, but documenting major changes and decisions.

Task:  GH-1184


dg0yt:
## Color overview
- Knockout/overprinting might need a second look, but is deprecated anyway.
- Color naming might need a second look.

ISOM-2017-2 colors

dg0yt:
Symbol set finished. Direct download for review: https://raw.githubusercontent.com/OpenOrienteering/mapper/84afdd69ad8933133a345698cf29cb4c6b145944/symbol%20sets/src/ISOM%202017-2_15000.xmap

dg0yt:
Note: For simplicity, I ignored the traditional, now-deprecated order of spot colors. I assume it will no longer have any use.

Zerbembasqwibo:
Stairways white doesn't cut the blue colors. Intentionally? (Railway-whit cut them)

When cheating with too short impassable fence or linear, it might appear two dots/tags and a short line. Wouldnt it be better the other way: the symbols appear first when line is longer then the distance between the "double symbols"? (Same with those symmbols but no "doubes, ordinary fence for example, but the problem is less.

dg0yt:
> Stairways white doesn't cut the blue colors. Intentionally? (Railway-whit cut them)

I wonder if there should be white at all. The ISOM only mentions black.
Apart from that, I don't expect stairways (up/down) to cut water features (flat). However, we probably don't want the white to cover magnetic north lines, paved areas, contour lines, track symbols.

> When cheating with too short impassable fence or linear, it might appear two dots/tags and a short line. Wouldnt it be better the other way: the symbols appear first when line is longer then the distance between the "double symbols"?

The question is if "cheating" should be impossible, hard, or convenient. Without the "decorations", the symbol is incomplete and can be misunderstood. 
The same for a narrow ride which is too short. It doesn't become unambiguous by drawing a gap between tiny dashes.

PR #1382, opened 2019-10-16
Minor updates

dg0yt:


dg0yt:
This now also integrates code-check-wrapper, clang-tidy and iwyu into the clazy CI job. 
Despite using the same Ubuntu version (18.04 LTS) as my main system, CI generates different false warnings...

PR #1412, opened 2019-11-08
TemplateTest: Initialize OSM_CONFIG_FILE

dg0yt:
Ensure that the tests use the file from the Mapper source code.


dg0yt:
Not needed. But we need GDAL data to be present at build time in order to run the `template_t` test.

PR #1417, opened 2019-11-11
Mobile UI: Add baseline and hatched areas view

lpechacek:
Baseline view fix for Android was originally implemented by 0230da0 but was reverted later by b48b4f2. Neither b48b4f2 not the merge commit provide understandable explanation why the change was reverted.

In this series patch titled "Android: Fix baseline view rendering", which was originally superseded by 0230da0, at least survived a few hundred hours of fieldwork on my devices.


dg0yt:
It is mentioned somewhere that 0230da0 went into to master too quickly at that time, together with other changes. I guess this why it was reverted. Not sure why I didn't bring it back to dev later. It still seems attractive to me, eliminating the hard coded variant for Android.

Apart from that, I'm fine with this PR. However, if ask for review, I would come up with stylistic changes such as `auto*`.

lpechacek:
Thanks, Kai, for the quick response. I also like 0230da0 more than my simpler fix. If you don't mind I'll replace my patch with yours and refresh the patch set. I'll also include the stylistic changes and possibly also add patch polishing surrounding code lines in the same manner.

PR #1428, opened 2019-11-18
Support for Gentoo Linux (licensing)

pkturner:
Adding a gentoo-licensing.cmake to the source package would make it easier to build Mapper releases on my Gentoo-based desktop machine.

Gentoo Linux has support for all the third party dependencies except Polyclipping. I have my own "ebuild" files for Polyclipping and OpenOrienteering Mapper, which I may see about submitting for inclusion in the Gentoo distribution. These are fairly routine.

Gentoo's package database lists licenses, and for GDAL the list is  just "BSD Info-ZIP MIT". Text for each of these licenses is stored separately. It is meager in comparison to what Mapper shows for GDAL licensing on other platforms. I have set it up so that GDAL appears on Mapper's licensing page in a bare bones line just like the "easy" dependencies mentioned in `linux-distribution.cmake`.

Guidance/help/suggestions are welcome.

### DIFF
--- a/doc/licensing/gentoo-licensing.cmake
+++ b/doc/licensing/gentoo-licensing.cmake
@@ -1,0 +1,29 @@
+#
+#    Copyright 2019 Kai Pastor
+#    
+#    This file is part of OpenOrienteering.
+# 
+#    OpenOrienteering is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+# 
+#    OpenOrienteering is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


dg0yt:
Thanks for the PR!

The default branch (for PRs) is `master` now. Either you edit your PR to contain a single commit against master, or we need to change the PR to target `dev`.

Regarding GDAL: Pristine GDAL installs a LICENSE.TXT to its data folder (`/usr/share/gdal` or similar). If this file is available on Gentoo, it might be just what you are looking for.

(The default licensing provider is OFF for Windows, macOS and Android. We might simply use the same for Linux, instead of NOTFOUND which is really meant to stop configuration.)

pkturner:
Thanks for the pointer to the GDAL license file. That _is_ what I have been looking for.
I changed this PR to have 1 commit based on master.

dg0yt:
Since the file became quite simple now, may I ask you to try a clean build (from an empty build dir) with this branch? If you did that already, a confirmation is enough.

pkturner:
Yes, I've built that revised branch. I've also used the same file in a patch to build v0.9.0 and v0.9.1_pre20191118.

PR #1436, opened 2019-12-02
Update Translations

dg0yt:


dg0yt:
These changes will be merged via GH-1437, to catch changes made to the source code in that PR.

PR #1438, opened 2019-12-03
Create release v0.9.1

dg0yt:
Actually identical to the current master which can be regarded as 'release candidate'.


dg0yt:
Fixed the ISSkiOM issues found in the latest master build, and updated translations. Will go on with release creation tonight if nobody objects.

PR #1455, opened 2019-12-12
Minor refactoring of Map and Symbol object interfaces

lpechacek:
This is part of code from PR#1251 which is making very slow progress. The three patches clean up minor troubles in `Map` API and `Symbol` id handling.

The remaining work for the symbol id undo is:
* Decoupling of symbol id from its position in symbol selector widget. I've hit several cul-de-sac's and I'm exploring a "sparse vector" implementation now.
* The actual symbol undo implementation, which does exist in https://github.com/lpechacek/mapper/tree/issue-1200-symbol-delete-undo and hit the roadblock of volatile symbol id assignment, needs stable symbol id for undo buffer serialization.

Please consider merging this preparatory cleanup work.

### DIFF
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -892,7 +892,7 @@ std::size_t Map::deleteIrregularObjects()
 	{
 		for (auto part : parts)
 		{
-			if (part->deleteObject(object, false))
+			if (part->deleteObject(object))
 			{
 				++result;
 				goto next_object;
@@ -945,7 +945,7 @@ void Map::deleteSelectedObjects()
 			if (index >= 0)
 			{
 				undo_step->addObject(index, *obj);
-				part->deleteObject(index, true);
+				part->releaseObject(index);
 			}
 			else
 			{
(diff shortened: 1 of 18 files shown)


dg0yt:
Thanks for the PR. I have a question about the last commit (Restricting symbol IDs to integers). Do you aim for a particular benefit? Or is loading generic IDs (too) broken?

lpechacek:
Thanks for the excellent question, Kai. I acknowledge that the patch alone makes limited sense.

`Symbol: Enforce numeric symbol id's` is a preparation for the stable symbol id implemented as an integer. We've agreed on that approach in PR#1251 (https://github.com/OpenOrienteering/mapper/pull/1251#issuecomment-501285897). While working on the implementation I found that strictly integer id's are beneficial for the current state of the codebase. Specifically, I can easily find number sequence gaps for reuse and I can use simpler data structures for symbol<->id association instead of hash maps and trees.

Seeing no harm in enforcing the integer-only id, I've packed the patch in this submission as a generic cleanup. Hope this clarifies my intent.

dg0yt:
Allright. I would suggest you go ahead with merging these and the following changes when you think they are ready, or explicitly add me as a reviewer when I shall take a closer look (which will take some time).

lpechacek:
Thanks, Kai, for the review and offer of help for the upcoming changes. Second pair of eyes will definitely useful for the symbol undo changes.
I'm merging this pull request.

PR #1468, opened 2019-12-20
Resolve ambiguous file name extensions in GDAL/OGR

dg0yt:


lpechacek:
Possibly unrelated to this patch set, but I thought that the symptom I'm seeing could be eradicated by the driver shadowing.

While loading a JPEG template for a map I'm working on, I can see following message on console:
```
GdalTemplate: Using GDAL driver 'JPEG'
GdalTemplate raster bands: RedByte, GreenByte, BlueByte
ERROR 6: Reading this image would require libjpeg to allocate at least 124949574 bytes. This is disabled since above the 104857600 threshold. You may override this restriction by defining the GDAL_ALLOW_LARGE_LIBJPEG_MEM_ALLOC environment variable, or recompile GDAL by defining the GDAL_LIBJPEG_LARGEST_MEM_ALLOC macro to a value greater than 104857600
ERROR 1: /home/lpechacek/Documents/OB/mapy/<anonymized>.jpg, band 3: IReadBlock failed at X offset 0, Y offset 0: Reading this image would require libjpeg to allocate at least 124949574 bytes. This is disabled since above the 104857600 threshold. You may override this restriction by defining the GDAL_ALLOW_LARGE_LIBJPEG_MEM_ALLOC environment variable, or recompile GDAL by defining the GDAL_LIBJPEG_LARGEST_MEM_ALLOC macro to a value greater than 104857600
GdalTemplate: Falling back to QImageReader, reason: Failed to read image data: /home/lpechacek/Documents/OB/mapy/<anonymized>.jpg, band 3: IReadBlock failed at X offset 0, Y offset 0: Reading this image would require libjpeg to allocate at least 124949574 bytes. This is disabled since above the 104857600 threshold. You may override this restriction by defining the GDAL_ALLOW_LARGE_LIBJPEG_MEM_ALLOC environment variable, or recompile GDAL by defining the GDAL_LIBJPEG_LARGEST_MEM_ALLOC macro to a value greater than 104857600
```
Needless to say that, the template loads fine in the end.

Is the preference for `QImageReader` in JPEG case something to be addressed by this patch set or shall I file a new bug report, please?

dg0yt:
> Is the preference for QImageReader in JPEG case something to be addressed by this patch set or shall I file a new bug report, please?

I actually wanted to have as much formats as possible exposed to GDAL, for feedback, and it just turns out to be useful ;-) I left only PNG and BMP for direct use with `TemplateImage` /`QImageReader`, because this is relevant for Paint-on-Template.

For JPEG, the preference is `GdalImageReader` since the addition of `GdalTemplate`, but that template class falls back to QImageReader if GDAL fails/refuses to load a file - this is what you observed due to a size issue. Falling back also happens/works for unsupported color formats. I tested with the TIFF test file collection from Qt ImageFormats, which has 16 bits per color files which `GDALImageReader` cannot handle at the moment.

If the fallback works reliable enough, we can leave it as is (i.e. expose most formats to GDAL). If we prefer stability in v0.9, we can give preference to Qt for all (or most) formats it supports. What do you think?

lpechacek:
> If the fallback works reliable enough, we can leave it as is (i.e. expose most formats to GDAL). If we prefer stability in v0.9, we can give preference to Qt for all (or most) formats it supports. What do you think?

Thanks for the explanation, Kai. I'm not in a position to make an expert decision on the benefits of full switch to GDAL for template image loading. I do have some faint idea about what the benefits could be but I haven't tested them or checked for them in the code. On a more general level, however, I'm fine with making the switch to the new loader, assuming that end-users are made aware of the change. And thanks for pushing the GDAL integration further!

dg0yt:
> `Reading this image would require libjpeg to allocate at least 124949574 bytes.`

At least, this seems to indicate that (large blocks of) memory would be allocated more than once for loading a single image. I guess it is better to use `QImageReader` directly for JPEG.

Issue #1467, opened 2019-12-19
"Import and remove" changes OMAP template position


pinnabilli:
### Steps to reproduce
The following is the simplest scenario I found in which the issue occurs (no georeferencing needed).
1. Add an OMAP template to a map.
2. If necessary, change the template's rotation attribute to a nonzero value.
3. Apply the "Import and remove" function.

### Actual behaviour
The template's objects are placed in wrong position; it looks like an unwanted shear transformation occurs.

### Expected behaviour
The template's objects should be placed as displayed before the "Import and remove" function is applied.

### Configuration
Mapper Version: 0.9.1
Operating System: Manjaro Linux 18.1.4

dg0yt:
Thanks for reporting the issue. I observed something similar a few days ago.

pkturner:
The bug is in Object::rotate(qreal angle). In one place where sin_angle is used, it needs to be subtracted.

dg0yt:
@pkturner Do you want to prepare a PR?

PR #1472, opened 2019-12-20
Object: Fix rotate() math.

pkturner:
This fixes the problem reported in #1467 with "import and remove" messing up the template position.

### DIFF
--- a/src/core/objects/object.cpp
+++ b/src/core/objects/object.cpp
@@ -587,8 +587,8 @@ void Object::rotate(qreal angle)
 	for (auto& coord : coords)
 	{
 		auto const old_coord = MapCoordF(coord);
-		coord.setX(cos_angle * old_coord.x() + sin_angle * old_coord.y());
-		coord.setY(sin_angle * old_coord.x() + cos_angle * old_coord.y());
+		coord.setX(+ cos_angle * old_coord.x() + sin_angle * old_coord.y());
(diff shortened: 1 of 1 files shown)


src/core/objects/object.cpp
```diff
@@ -588,7 +588,7 @@ void Object::rotate(qreal angle)
 	{
 		auto const old_coord = MapCoordF(coord);
 		coord.setX(cos_angle * old_coord.x() + sin_angle * old_coord.y());
-		coord.setY(sin_angle * old_coord.x() + cos_angle * old_coord.y());
+		coord.setY(cos_angle * old_coord.y() - sin_angle * old_coord.x());
 	}
```
dg0yt: I see, this hits the nail. This function is only used from TemplateTransform. However, for symmetry with `Obejct::rotateAround(const MapCoordF& center, qreal angle)` just above these lines, I would prefer to leae the x/y order and rather write:

~~~
		coord.setX(+ cos_angle * old_coord.x() + sin_angle * old_coord.y());
		coord.setY(- sin_angle * old_coord.x() - cos_angle * old_coord.y());
~~~

pkturner: Now that I've made the change, I see how it reflects the common layout of a rotation matrix.

dg0yt:
This also fixes the bug which I observed. Thank you very much!

Issue #1470, opened 2019-12-20
Sporadic crashes when using vectoriser


mpickering:
I am using an artifact from the most recent master commit to test the vectoriser code. (MacOS - https://dev.azure.com/OpenOrienteering/Mapper/_build/results?buildId=1310&view=artifacts&type=publishedArtifacts)

From time to time, mapper crashes without any error message, I haven't managed to work out how to reproduce the crash yet. All map state is completely lost. 

Any tips on how to provide some more information so we can debug this issue?

lpechacek:
Thanks for the bug report, Matthew! Of course, the best thing to get from a crash is a core dump or at least a stack trace. However, I know nothing about Mac OS and debugging possibilities on that platform for I cannot give guidance at the moment.
On a more generic level, any piece of information might be important. If you can describe if Mapper crashes in the CoVe dialog, after closing CoVe or even without starting CoVe at all, these are all important data points. Also if you can share data files with which Mapper crashes, I can try playing with the same setup to see if I spot something.

dg0yt:
Assumed to be fixed with GH-1487. Re-open otherwise.

Issue #1243, opened 2019-05-06
Enhancement: Add progress bar for operations taking long time


lpechacek:
### Steps to reproduce
1. Use simplify path or cutout tool on large number of objects.

### Actual behaviour
Mapper freezes until operation finishes. There is also no feedback about operation start.

### Expected behaviour
Progress bar should appear. As a bonus many of these operation can also be parallelized with help of `QtConcurrent`.

### Configuration
Mapper Version: origin/dev
Operating System: openSUSE Tumbleweed

PR #1475, opened 2019-12-22
CoVe review

dg0yt:


dg0yt:
These commits add CoVe to Mapper's code check (iwyu, clang-tidy), resolve all warnings, and apply some additional modernizations and modifications. I need to draw a first cut at this point...

... however, there are many more things which I would like to touch:
- Default initialization / constructor modernization.
- Simplifying the `OrganizableElement` hierarchy, possibly flattening it completely.
- Apply morpological operations in a separate thread.
- Unify spelling style of class names and of file names.
- Do not define class names starting with 'Q'.
- Create a left/right split UI instead of a top/bottom split UI (cf. symbol editor).
- Find a way to remove the dependency of `mainform.cpp` on Mapper.
- ...

And of course, integrating the translation is on the top of my todo list.

dg0yt:
`PolygonTest::testJoins(simple joins)` creates the same number of polygons and nodes on all platforms. But on Windows x32, the nodes have x/y deviations which are distributed as follows:

| Deviation |  0.0 | >0.0...1.0 |  >1.0...2.0 |  >2.0...3.0 | > 3.0 |
|--------------|------|---------------|---------------|---------------|--------|
| Occurences | 41864 | 55898 |     4        |         2       |    0    |

I will do one more CI run, with deviation interval bounds at 0.5, and then choose deviation interval bounds and max number of occurences.

dg0yt:
> I will do one more CI run, with deviation interval bounds at 0.5, and then choose deviation interval bounds and max number of occurences.

| Deviation |  <=0.5 | >0.5...1.5 | >1.5...2.5 | >2.5...3.5 | > 3.5 |
|--------------|----------|---------------|--------------|--------------|--------|
| Occurences | 97748  |     15    |        4      |       1        |     0   |

I will go with this intervals, and a a little margin on top of  15/4/1,

3rd-party/cove/app/Thread.cpp
```diff
@@ -21,29 +21,27 @@
...
   \param[in] v Vectorizer on which to run performClassification.
   \param[in] p UIProgressDialog to be passed to performClassification. Can be 0.
   */
-ClassificationThread::ClassificationThread(Vectorizer& v, UIProgressDialog* const p, QObject* parent)
-    : v(v)
+ClassificationThread::ClassificationThread(Vectorizer& v, ProgressObserver* const p, QObject* parent)
+    : QThread(parent)
+    , v(v)
     , p(p)
-    , QThread(parent)
```
lpechacek: I've stopped here for a moment and yes, this is completely correct. This code was written long before Guidelines came into existence. Hence the "wrong" initialization order. Thanks for the fixes!

3rd-party/cove/app/UIProgressDialog.cpp
```diff
@@ -43,8 +43,9 @@ UIProgressDialog::UIProgressDialog(const QString& labelText,
 {
 	pDialog.setMinimumDuration(0);
 	pDialog.setWindowModality(Qt::WindowModal);
-	connect(this, SIGNAL(percentageUpdated(int)), &pDialog,
-			SLOT(setValue(int)));
+	connect(this, &UIProgressDialog::percentageUpdated,
+	        &pDialog, &QProgressDialog::setValue,
+	        Qt::QueuedConnection);
```
lpechacek: This could be the reason for crashes reported in issue GH-1470. I recall fixing the inter-thread signaling and I'm surprised that it was wrong at this spot.
...
And this is where I broke it: https://github.com/lpechacek/mapper/commit/9bac3fd2519a110a754fbb28cf4cd869d682462b
Funnily, the commit message talks about queued connections. :)

Thanks for catching this!

lpechacek:
> ... things which I would like to touch:
>
>    Default initialization / constructor modernization.

Ack.

>    Simplifying the OrganizableElement hierarchy, possibly flattening it completely.

The hierarch was architected under the influence of 1990's approach to OO programming. No objection from me on getting rid of archaisms.

>    Apply morpological operations in a separate thread.

I had the impression that `QtConcurrent` will split workload into separate threads by default. Is the main thread occupied by part of the workload?

>    Unify spelling style of class names and of file names.
>    Do not define class names starting with 'Q'.
>    Create a left/right split UI instead of a top/bottom split UI (cf. symbol editor).

Ack on all three.

>    Find a way to remove the dependency of mainform.cpp on Mapper.

Which involves designing a data exchange interface. Any plan for the task or is the design still an open question, please?

lpechacek:
> I will go with these intervals, and a little margin on top of 15/4/1,

Another approach would be creating a "more controlled test". This version of `PolygonTest` is essentially a snapshot of a real map and was created as a support for a larger rework. This test is a quick hack which, until now, served well.

dg0yt:
> > Apply morpological operations in a separate thread.
> 
> I had the impression that `QtConcurrent` will split workload into separate threads by default. Is the main thread occupied by part of the workload?

I have looked into the thinning operation.The situation is that
- it is more or less mandatory,
- it cannot be easily parallelized (at least not with the image part approach),
- it is not run via QtConcurrent,
- progress is hard to estimate.
So it ends up heavily occupying the main thread indeed, with poor and lagging progress reporting.

dg0yt:
> > Find a way to remove the dependency of mainform.cpp on Mapper.
> 
> Which involves designing a data exchange interface. Any plan for the task or is the design still an open question, please?

The geometric data exchange interface might be the easier part (Qt, GDAL, etc.). But when I want to let the user choose a symbol from the map's symbol set, I cannot see how there would be an easy way to do that without using Mapper API... and maybe an argument to abandon this idea.

dg0yt:
> > I will go with these intervals, and a little margin on top of 15/4/1,
> 
> Another approach would be creating a "more controlled test". This version of `PolygonTest` is essentially a snapshot of a real map and was created as a support for a larger rework. This test is a quick hack which, until now, served well.

Ack, I see the current test more as a starting point, but it is useful to have the test and the coverage reporting in CI, to guide future work.

3rd-party/cove/app/UIProgressDialog.cpp
dg0yt: Hm, maybe QueuedConnection shoots back. Now the progress update is put into the main thread's event queue, but running the event loop is no longer triggered...

lpechacek:
> > > Apply morpological operations in a separate thread.
> > 
> > 
> > I had the impression that `QtConcurrent` will split workload into separate threads by default. Is the main thread occupied by part of the workload?
> 
>  I have looked into the thinning operation.The situation is that ... with poor and lagging progress reporting.

I've also taken a closer look at the morphological operations and I can see that they haven't been converted to `QtConcurrent` despite my impression that I did the work. Sorry for the noise, I merely lost track of what has been done.

I also looked at the sloppy progress reporting. Indeed the queued connections approach does not perform that well as the original custom-event based one. A patch returning the old implementation back to CoVe is in https://github.com/lpechacek/mapper/tree/cove-smooth-progress-reporting. This is a quick shot at the revert that needs further review. Yet, feel free to share any high-level ideas you have for this message passing mechanism. In the future, I'd like to apply the same approach for the core Mapper routines for issue https://github.com/OpenOrienteering/mapper/issues/1243.

lpechacek:
> > > Find a way to remove the dependency of mainform.cpp on Mapper.
> > 
> > 
> > Which involves designing a data exchange interface. Any plan for the task or is the design still an open question, please?
> 
> The geometric data exchange interface might be the easier part (Qt, GDAL, etc.). But when I want to let the user choose a symbol from the map's symbol set, I cannot see how there would be an easy way to do that without using Mapper API... and maybe an argument to abandon this idea.

Thanks for the clarification, Kai. I don't have a good solution either. :(

dg0yt:
> Yet, feel free to share any high-level ideas you have for this message passing mechanism. In the future, I'd like to apply the same approach for the core Mapper routines for issue #1243.

Indeed, I'm currently studying how to refactor this, and I will post a PR soon.

PR #1493, opened 2020-01-11
XMLFileImporter: Validate declination/grivation

dg0yt:
Modify declination when declination, grivation and convergence are
not consistent. Resolves GH-1235 (errors on editing declination).


dg0yt:
@pkturner That's my take on the file loading part of #1235.

Issue #1497, opened 2020-01-15
Occasional Mapper crash after drawing a circle object


lpechacek:
### Steps to reproduce
1. Select a line symbol
2. Draw a circle
3. ??? some magic movement, cannot tell for sure, perhaps select edit mode

### Actual behaviour
Debug build crashed with
```
/usr/include/c++/9/bits/stl_iterator.h:820:17: runtime error: reference binding to null pointer of type 'struct PathPart'
/usr/include/c++/9/bits/stl_vector.h:1121:16: runtime error: reference binding to null pointer of type 'struct value_type'
/data/src/oomapper/src/gui/map/map_editor.cpp:2544:177: runtime error: member access within null pointer of type 'struct value_type'
/data/src/oomapper/src/gui/map/map_editor.cpp:2544:177: runtime error: member call on null pointer of type 'struct VirtualPath'
/data/src/oomapper/src/core/virtual_path.cpp:339:2: runtime error: member access within null pointer of type 'const struct VirtualPath'
```

### Expected behaviour
n/a

### Configuration
Mapper Version: master-v20191221.17-62-gb4466b076612
Operating System: openSUSE Tumbleweed

Possibly related, this is a backtrace of a similar crash I observed on Android with another custom build. I was unable to replicate that one either. I'm attaching this data bit just in case it helps to determine the root cause of this irregular crash.

```
--------- beginning of crash
08-15 10:56:52.818  3953  3967 F libc    : Fatal signal 11 (SIGSEGV), code 1, fault addr 0x0 in tid 3967 (qtMainLoopThrea)
08-15 10:56:52.913  7609  7609 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
08-15 10:56:52.914  7609  7609 F DEBUG   : Build fingerprint: 'nvidia/wx_na_wf/shieldtablet:7.0/NRD90M/1928188_1006.1489:user/release-keys'
08-15 10:56:52.914  7609  7609 F DEBUG   : Revision: '0'
08-15 10:56:52.914  7609  7609 F DEBUG   : ABI: 'arm'
08-15 10:56:52.914  7609  7609 F DEBUG   : pid: 3953, tid: 3967, name: qtMainLoopThrea  >>> org.openorienteering.mapper.dev <<<
08-15 10:56:52.915  7609  7609 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
[...]

PR #1498, opened 2020-01-17
DrawCircleTool: Don't create empty objects

dg0yt:
For creating a circle, we need more than just the position of the
initial mouse press. That's why this tool leaves the preview object
empty after startDrawing(). However, the empty preview object must
not be added to the map on finishDrawing.
Fixes GH-1497 (crash after drawing with circle tool).

### DIFF
--- a/src/tools/draw_circle_tool.cpp
+++ b/src/tools/draw_circle_tool.cpp
@@ -31,6 +31,7 @@
 #include <QMouseEvent>
 #include <QLatin1String>
 #include <QPixmap>
+#include <QPointF>
 #include <QRectF>
 #include <QString>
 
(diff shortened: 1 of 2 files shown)


src/tools/draw_circle_tool.cpp
```diff
@@ -223,7 +223,10 @@ void DrawCircleTool::finishDrawing()
 	second_point_set = false;
 	updateStatusText();
 	
-	DrawLineAndAreaTool::finishDrawing();
+	if (preview_path->parts().empty())
```
dg0yt: `|| preview_path.parts().back().empty()`

lpechacek:
> On an afterthought, ...

The thing is that it is not safe to call the following sequence: `startDrawing(); finishDrawing();`  I'd expect a sensible API to allow this.

lpechacek:
https://github.com/lpechacek/mapper/commits/issue-1497-draw-circle-crash-fix contains my proposal of the fix. It embodies all the above ideas, as far as I can tell.

dg0yt:
> The thing is that it is not safe to call the following sequence: startDrawing(); finishDrawing();  I'd expect a sensible API to allow this.

We agree on this point.

> https://github.com/lpechacek/mapper/commits/issue-1497-draw-circle-crash-fix contains my proposal of the fix. It embodies all the above ideas, as far as I can tell.

That's one change to fix DrawCircleTool::finishDrawing (the special case) and another one to change DrawLineAndAreaTool::finishDrawing (the general case), right? This reflects my afterthought after reviewing my own change, and the chosen conditions look good to me.

dg0yt:
Picked your DrawCircleTool change (but edited commit message).
Rewrote the DrawLineAndAreaTool change.

Issue #1503, opened 2020-01-24
CoVe: Corrupted B/W image


lpechacek:
Please note that this misbehavior has been observed on a development build.

### Steps to reproduce
1. Attempt to vectorize attached example template ([fb-orienteering-mappers-mom-2020.zip](https://github.com/OpenOrienteering/mapper/files/4108436/fb-orienteering-mappers-mom-2020.zip))
2. Classify colors into 20 buckets, choose the brown ones
3. Switch to Thinning tab

### Actual behaviour
Corrupted B/W image. The machine is a two-core notebook with (Intel) hyperthreading off. Processing was split into two chunks.

![image](https://user-images.githubusercontent.com/5584406/73068954-cd6d2280-3eac-11ea-940f-a2c5b701bf29.png)

Going back to the color classification tab and retrying classification gave another interesting result.

![image](https://user-images.githubusercontent.com/5584406/73069421-fe9a2280-3ead-11ea-9318-b3a1a6899b46.png)

It seems that data for the bottom part of the picture come from uninitialized memory. Interestingly enough, the address sanitizer didn't complain so my guess is that the allocated memory was merely not populated with new data.

### Configuration
Mapper Version: master-v20191221.17-77-g43c0cb2b056a (debug build)
Operating System: openSUSE Tumbleweed 20200121

dg0yt:
I saw similar issues with some stripes being weird, but not in all the manual testing I did in the end.

> Mapper Version: master-v20191221.17-77-g43c0cb2b056a (debug build)

So this is the master v20191221.17 source code? Or a later build which includes my concurrency changes?

lpechacek:
> So this is the master v20191221.17 source code? Or a later build which includes my concurrency changes?

It is a build that contains the concurrency changes. The string in Mapper version field is an output of `git describe --tags`. It says that `HEAD` is v20191221.17 plus 77 additional commits. Behind the final `-g` there is SHA of the actual `HEAD`, i.e. 43c0cb2b056a. Git can parse the string as is and check out the corresponding tree state. I hope that this way of identifying development builds is fine for you.

PR #1505, opened 2020-01-25
Fix concurrent image processing

dg0yt:
Assumed to (help) resolve GH-1503.
However, not yet tested in an environment which reproduces the issue.

### DIFF
--- a/3rd-party/cove/libvectorizer/Concurrency.h
+++ b/3rd-party/cove/libvectorizer/Concurrency.h
@@ -131,7 +131,7 @@ struct Job
  * adds this object to the return value.
  */
 template <typename ResultType, typename FunctorType, typename ... Input>
-Job<ResultType> run(const FunctorType& functor, Input ... args)
+Job<ResultType> run(const FunctorType& functor, Input&& ... args)
 {
 	Progress progress;
 	auto future = QtConcurrent::run(functor, &FunctorType::operator(), std::forward<Input>(args)..., progress);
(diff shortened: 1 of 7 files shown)


lpechacek:
FWIW, images are generated correctly now in my test case. I'll dive into the patch set later on. Thanks!

dg0yt:
I'm still working on extending unit test coverage around my QImage handling, it is quite telling...

dg0yt:
So in the end, I learned that I don't need InplaceImage. It was just hiding the failure to implement perfect forwarding correctly. OTOH there was a problem with the const char*-taking constructor of QImage when used with index image formats which require a color table. I will create a Qt bug report.

dg0yt:
Reported https://bugreports.qt.io/browse/QTBUG-81674

dg0yt:
I must admit that I just see GH-1503 happening again, although not always, when switching between Colors and Thinning tab. But not a single crash... Maybe it is still some unexpected detaching in QImage.

dg0yt:
> Maybe it is still some unexpected detaching in QImage.

Indeed, there is occassional detaching during processing. But why only sometimes?

dg0yt:
So, finally making sure that the worker functions get a target QImage which is not implicitly shared with another instance, by creating this target QImage in the worker thread.

@lpechacek Ready for review, to be merged for the next master snapshot.

dg0yt:
... took out unrelated changes (unused slots, BW image history) - will move them to the refactoring PR.
Anyway, I will probably merge this soon, approved or not. I face #1503 w/o this PR, but not with this PR.

lpechacek:
Thanks for the improvements, Kai! I'll review the patches tomorrow morning.

PR #1508, opened 2020-01-26
Calculate grid scale factor compatibly 

pkturner:
### a new approach  
This PR is a descendant of #1277 (Determine grid scale factor automatically) showing an approach that I like. It eliminates the matrix that compensates for shear in the projection. As a consequence, this version very compatible with previous releases of Mapper. It computes grid scale factor, while also permitting the user to make adjustments. It has a user setting to tailor the Georeferencing dialog to the user's knowledge.

### the parallel  
These enhancements to both user interface and code can be understood more easily, if certain parallels are kept in mind between parameters indicating orientation of the grid, and factors governing scale of the grid.

| | | orientation | scale |
| --- | --- | --- | --- |
| **grid to model** | computable from CRS | convergence | grid scale factor |
| **model to terrain** | user-specified | declination | auxiliary scale factor |
| **grid to terrain** | used in transformation between projected and map coordinates | grivation | combined scale factor |
| **relationship** |  | declination = grivation + convergence | combined scale factor = grid scale factor * auxiliary scale factor |

As the table shows, convergence is analogous to grid scale factor in that both can be computed from the CRS and reference point. Similarly, declination is analogous to auxiliary scale factor as both are user-supplied.

Note, the “grid scale factor” as it appeared previously in Mapper, really meant “combined scale factor” and the more accurate term is used in both the new implementation and user interface.

The user-specified scale factor is called “auxiliary scale factor”, rather than “elevation scale factor” so as not to limit what users might do with it. 

Earlier versions of Mapper provided for editing the "grid scale factor" which was effectively the combined scale factor, compensating for both the grid and any other needed adjustments. With this change, the grid compensation is automatic, and the user no longer sets the combined scale factor. Now the user sets the auxiliary scale factor. That relieves the user of determining and dealing with the grid scale factor.

### the enhancement to Georeferencing  
The enhancement to the Georeferencing class is modest.
* It computes the convergence and grid scale factor together.
* The methods get/setGridScaleFactor are renamed to get/setCombinedScaleFactor.
* get/setAuxiliaryScaleFactor methods are added.
* The auxiliary_scale_factor is saved/loaded as an attribute in .omap files.
* It adds an auxiliaryScaleFactorChanged signal.

### compatibility  
The new Georeferencing class maintains compatibility by using 'grid_scale_factor' in .omap files to save and restore the _combined_ scale factor. It introduces an 'auxiliary_scale_factor' attribute into .omap files. This new attribute, if encountered by an earlier version of Mapper, is ignored with no ill effect.
When the new Georeferencing class loading from a file, the auxiliary scale factor is calculated if absent. I suppose it should always be re-calculated, as is done for declination (#1493).

### use cases
Considering that there is no button to click that sets the scale factor to the calculated value, how does the user benefit from the changes?

**new map**  
The primary case that benefits from this PR is a new map, because the grid scale factor is computed, and the auxiliary scale factor is simply 1.0. It is automatic, without the user explicitly choosing "automatic".

**changed CRS**  
Another interesting case is when the CRS (or reference point) of an existing map is changed. The grid scale factor is re-computed, but the auxiliary scale factor is kept. This is obviously a good result if the auxiliary scale factor has been set to account for elevation. Thus, the combined scale factor is again automatic.

**existing map**  
If the grid scale factor was set using an earlier Mapper, the value will appear unchanged in the Georeferencing dialog as “Combined scale factor”. (An Auxiliary scale factor will also be displayed, having the value which is consistent with the combined scale factor and the calculated grid scale factor.)

If a user wishes to take advantage of calculated grid scale factor for such a map, he/she may set the auxiliary scale factor to 1.0, using the Georeferencing dialog.

The effect of setting auxiliary scale factor is not trivial, as indicated when a dialog pops up, asking “Do you want to stretch/shrink the map content accordingly?” If “stretch/shrink” is declined, then the global coordinates of all map objects will be affected. This new dialog is analogous to the “Rotate map” dialog which appears if the user changes the declination.

### taming the Georeferencing dialog  
It can be understood from the description of this enhancement, that the benefit will occur gradually, as new maps are created. Note that the new feature does not depend on a feature flag being turned on in the Georeferencing object, or in the way it is saved to file.

At the same time, we are not content with having made the Georeferencing dialog bigger. We would like to give the user an option to avoid yet more subtlety around scale factors.

Noticing that the new Georeferencing does not require any user interaction to achieve its benefits, and gives all users less need to attend to scale factor, this PR proposes to remove the scale factor from the dialog, and give the user a _setting_ to make the scale factors visible and editable.


dg0yt:
:+1:

Regarding the georeferencing dialog, I'm fine with hiding the auxiliary scale factor field by default. However, wouldn't it be better to always show the auxiliary factor field when the current map has a value different from 1.0? After all, the user might be looking into the dialog to find an explanation for some observations, or for what he knew to be there as "Grid scale factor".
Or we could turn the grid compensation section into something collapsible/expandable, instead of using the settings dialog.

pkturner:
@dg0yt 
> However, wouldn't it be better to always show the auxiliary factor field when the current map has a value different from 1.0? 

One reason against that is that most existing maps will have an auxiliary scale factor different from 1.0, while at the same time the user will be doing well to not bother with it. Here's a typical case I expect.

| factor | value |
| ---- | ---- |
| combined scale factor | 1.0 |
| grid scale factor | 0.9999 |
| auxiliary scale factor | 1.0001 |

That's because the mapper has until now left the scale factor at 1.0. The new Georeferencing keeps this as the combined scale factor value, in order to retain both the scale of objects on paper and their lat/long on the globe. It calculates the grid scale factor, typically not 1.0, and consequently the auxiliary scale factor also cannot equal 1.0.

In such a case, showing the auxiliary scale factor would lead to unhelpful discontent.

Showing the auxiliary scale factor could have a minimum **threshold** before it would appear. I was lazy and didn't want to choose a threshold arbitrarily. 

-------
A weaker reason is that it's more comfortable for users if the form of the dialog doesn't vary when switching between maps.

pkturner:
@dg0yt 
> Or we could turn the grid compensation section into something collapsible/expandable, instead of using the settings dialog.

That would be good, provided the collapsed/expanded state were kept as a user preference rather than as a property of the map.

I don't have a specific vision of how that should work in the Georeferencing dialog. If you would point me to what you have in mind, I'd be happy to implement it. Is there a case of another user preference in Mapper that is controlled apart from the settings dialog?

dg0yt:
> > Or we could turn the grid compensation section into something collapsible/expandable, instead of using the settings dialog.
> 
> That would be good, provided the collapsed/expanded state were kept as a user preference rather than as a property of the map.

Yes, this is what I meant.

> I don't have a specific vision of how that should work in the Georeferencing dialog. If you would point me to what you have in mind, I'd be happy to implement it. Is there a case of another user preference in Mapper that is controlled apart from the settings dialog?

We have some check boxes in the line symbol settings which also control the visibility of additional fields. The dependent widgets are kept in lists which are used when changing the state, e.g. for dashed lines:

https://github.com/OpenOrienteering/mapper/blob/e1155d841559963a0148f4b8a8d9d6c5aa6a51c0/src/gui/symbols/line_symbol_settings.cpp#L686-L715

For the start, use a "Grid compensation..." checkbox, a flat, checkable "Grid compensation..." push button or a label with a hyperlink.  Later, we can override the normal checkbox marks or prefix the button/label text with a small triangle (▶ / ▼), or plus/minus (`[+]`/`[-]`). We must see what works best cross-platform.

pkturner:
> For the start, use a "Grid compensation..." checkbox, a flat, checkable "Grid compensation..." push button or a label with a hyperlink. 

Thanks. I'll do that.

pkturner:
@dg0yt Thanks for all the guidance. It's refreshing to work with up-to-date coding standards.
Should I close this?

dg0yt:
> Thanks for all the guidance. It's refreshing to work with up-to-date coding standards.

Thanks for taking the time and for the explanations for your side.

> Should I close this?

It will be closed automatically when I click merge. I just wanted to make sure there were no more changes pending.

src/core/georeferencing.cpp
```diff
@@ -884,6 +1013,8 @@ bool Georeferencing::setProjectedCRS(const QString& id, QString spec, std::vecto
 			if (ok && state != Normal)
 				setState(Normal);
 		}
+		if (!ok)
+			updateGridCompensation();
```
dg0yt: Q: Why do we call updateGridCompensation() after an unsuccessfull change of the CRS, but not after a successful change?

pkturner: When the change is successful, the call to `setState(Normal)` calls updateGridCompensation, so a direct call is not needed. When the change is unsuccessful, updateGridCompensation is still needed to set convergence and grid_scale_factor consistent with the CRS, even if they are trivially 0 and 1.0 respectively.

I hope this helps. I am away from my office for a few days, so I do not have the opportunity just now to re-check all the paths through this section of code.

PR #1521, opened 2020-02-07
Changing a template's georeferenced state may trigger lossy rewriting of the template file

dg0yt:
Changing whether a template is used georeferenced or not does not mean
that the template file itself should be rewritten. In fact, the old
behaviour is dangerous: Rewriting may cause loss of quality (JPEG) or
loss of georeferencing information (GeoTIFF -> TIFF).


dg0yt:
I didn't noticed the bug when we enabled changing the georeferenced state. but now I was testing with GeoTIFF, using our test files in git, and only git helped me recover the GeoTIFF. 

I think this might make a new official release more urgent, with proper information.

Issue #1515, opened 2020-02-03
(Slightly) misplaced templates


jmacura:
### Steps to reproduce
1. open .tiff with World file as a template
2. open Geotiff as a template
3. see how much they align

### Actual behaviour
They are mutually shifted by some 16 m. It is kind of hard to say which one is correct, but I feel like the Geotiff is misplaced.
Below is a screenshot from Mapper. Transparent b/w background is shifted (GeoTIFF), while the ortophoto below (.tiff+.tfw) is correctly placed (I assume)
![image](https://user-images.githubusercontent.com/5598693/73699030-02664a00-46e3-11ea-9178-3d1112fdd2a7.png)

### Expected behaviour
All georeferenced templates should fit.
Below is a screenshot of those very rasters in QGIS 3.10.
![image](https://user-images.githubusercontent.com/5598693/73699216-89b3bd80-46e3-11ea-9152-824324cf7b6b.png)

### Configuration
Mapper Version: 0.9.1 & last master (20191221.17)
Operating System: Win 10 Pro v1903

### Notes
Sadly, I should not be sharing the template files ...
The CRS I use is EPSG:5514. It can be tricky sometimes :-(

dg0yt:
Thanks for reporting the issue. It will be difficult to find out where this particular issue comes from. The main difference to QGIS is our usage of more recent versions of PROJ and GDAL. That's why we already stumbled about various errors.

We do have test files for world file and GeoTIFF, and they are run as unit tests. During our builds, the unit test for these files does not indicate any issue. You may try yourself with the files from test/data/templates. 

If you don't mind sharing the geographic coordinates, than let us know the georeferencing:
- The contents of the world file, and also the width and heigth of the file, in pixels.
- The geoinformation of the TIFF file. On Linux, I would use the command `listgeo`. This command belongs to the geotiff package. Maybe there is something similar with QGIS. (And `listgeo` can be used to generate world files.)

For the test files, the output is like this:

`OpenOrienteering/mapper/test/data/templates$ cat world-file.pgw`

~~~
0.15
0
0
-0.15
649078
394159
~~~

`mapper/test/data/templates$ listgeo *.tiff`

~~~
Geotiff_Information:
   Version: 1
   Key_Revision: 1.0
   Tagged_Information:
      ModelTiepointTag (2,3):
         0                 0                 0                
         649077.925        394159.075        0                
      ModelPixelScaleTag (1,3):
         0.15              0.15              0                
[...]

PR #1522, opened 2020-02-12
Revise template georeferencing setup

dg0yt:
Increase flexibility without further complicating the code. Most of all,
- allows overriding of CRS information when it differs from the map, or when (re-)enabling georeferencing for an existing template,
- mitigates GH-1515 (slightly misplaced templates.


dg0yt:
@lpechacek I don't ask for a thorough review. I just would like to verify that the changes take a good direction, and we can let them into the pending master snapshot.

PR #1523, opened 2020-02-14
OCD: Silence false warnings on shifted rows pattern export

lpechacek:
I've been irritated by the warning about shifted rows pattern upon export to OCD for some time. While Mapper has a more flexible point pattern definition mechanism than the other program, there is one case where they match. The case typically occurs when and OCD is loaded to Mapper and then exported back to OCD. The import/export routines are at the level of completeness that this workflow can complete without excessive noise.

This patch detects when Mapper can successfully export shifted rows area pattern and silences the warning in this particular case. The code change does not look particularly beautiful and I'm considering it a "v1" with the possibility of rework.

### DIFF
--- a/src/fileformats/ocd_file_export.cpp
+++ b/src/fileformats/ocd_file_export.cpp
@@ -1368,12 +1368,43 @@ quint8 OcdFileExport::exportAreaSymbolCommon(const AreaSymbol* area_symbol, OcdA
 				// NOTE: This is only a heuristic which works for the
 				// orienteering symbol sets, not a general conversion.
 				// (Conversion is not generally possible.)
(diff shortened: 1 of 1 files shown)


lpechacek:
@dg0yt, I'd be thankful for your review and comments.

dg0yt:
Shall I pick this up? It is really worth pursueing.

lpechacek:
Feel free to do so. The patch rework is currently low on my TODO list. Thanks!

lpechacek:
And thanks for the review! I consider all the above valid points and I also share the long term ideas about export code improvements.

lpechacek:
Reimplemented in the above-referenced pull request.

PR #1531, opened 2020-03-04
Finishing touches to cove

dg0yt:
Adds a symbol selection combo box, and a few mor change to the UI.

A manual page with screenshots is coming in a separate PR.


3rd-party/cove/app/mainform.cpp
```diff
@@ -772,43 +773,45 @@ void mainForm::on_createVectorsButton_clicked()
...
 		{
-			// transform from template coordinates to map coordinates
-			OpenOrienteering::MapCoordF templCoords =
-				ooTempl->templateToMap(QPoint(point.x() + xOff, point.y() + yOff));
-			OpenOrienteering::MapCoord c(templCoords.x(), templCoords.y());
-
-			newOOPolygon->addCoordinate(c);
+			if (!coords.back().isPositionEqualTo(coords.front()))
+				coords.push_back(coords.front());
+			coords.back().setClosePoint(true);
```
lpechacek: I'm wondering what is the reason for exposing the inner workings of closed paths. Would some sort of delegation to `PathPart::setClosed()` be an option here?

src/gui/widgets/symbol_dropdown.h
```diff
@@ -81,7 +96,7 @@ protected slots:
 	// happen as long as a SymbolDropDown is shown, which is the case currently)
 	
 private:
-	int num_custom_items;
+	int num_custom_items = 0;
```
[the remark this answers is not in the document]
dg0yt: So far, I prefer the `=` syntax for direct member initialization, at least for `int` and friends. There should be no difference in compiled code, at least if some members are non-zero. Braces are used often enough, for enum, sub-types, direct function definitions.

3rd-party/cove/app/mainform.cpp
dg0yt: I noticed that this pulled more of the inner workings into this function, and that there may be a missing piece of API. I just didn't want to add new API too quickly.
But we are not dealing with `PathParts` here: It is `MapCoordVector` which I wanted to construct completely (i.e. closed) before constructing the `PathObject`.

dg0yt: Studied the code more deeply. I will not add new API, but I will revert to calling `PathObject::closeAllParts()`.

src/gui/widgets/symbol_dropdown.h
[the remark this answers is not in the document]
lpechacek: Ack.

PR #1535, opened 2020-03-07
cove: Warn when vectorizing thick lines

lpechacek:
It is very easy for a user to omit the thinning operation in line vectorization. The function then gives unsatisfactory results. This patchset makes CoVe emit a warning in case the user runs vectorization on B/W image which didn't see thinning. Attempt for proper integration with the undo/redo functionality resulted in the removal of the legacy code and adoption of `QUndoStack`. Although the rewrite is fairly straightforward, I'd appreciate a review. Thanks!

### DIFF
--- a/3rd-party/cove/app/mainform.cpp
+++ b/3rd-party/cove/app/mainform.cpp
@@ -187,6 +187,7 @@ mainForm::~mainForm()
 void mainForm::clearBWImageTab()
 {
 	bwBitmapUndo->clear();
+	bwBitmapVectorizable = false;
 	ui.bwImageView->setPolygons(PolygonList());
 	ui.saveVectorsButton->setEnabled(false);
 	ui.bwImageView->setImage(nullptr);
@@ -520,19 +521,25 @@ bool mainForm::performMorphologicalOperation(
 	Vectorizer::MorphologicalOperation mo)
 {
 	QString text;
+	bool transVectorizable = false;
+
 	switch (mo)
 	{
 	case Vectorizer::THINNING_ROSENFELD:
 		text = tr("Thinning B/W image");
+		transVectorizable = true;
 		break;
 	case Vectorizer::PRUNING:
 		text = tr("Pruning B/W image");
+		transVectorizable = bwBitmapVectorizable;
 		break;
 	case Vectorizer::EROSION:
 		text = tr("Eroding B/W image");
+		transVectorizable = bwBitmapVectorizable; // formally, does not change the status
 		break;
 	case Vectorizer::DILATION:
 		text = tr("Dilating B/W image");
+		transVectorizable = false;
 		break;
 	}
 	UIProgressDialog progressDialog(text, tr("Cancel"), this);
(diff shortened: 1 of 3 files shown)


3rd-party/cove/app/mainform.cpp
```diff
@@ -516,19 +521,25 @@ bool mainForm::performMorphologicalOperation(
 	Vectorizer::MorphologicalOperation mo)
 {
 	QString text;
+	bool transVectorizable {};
```
dg0yt: This another example of the recent initialization style question. Maybe my concerns are more obvious here:
- In initialization, this style uses `{}` to express `false`.
- In assignement, a few lines later, `false` is used to express `false`.

3rd-party/cove/app/mainform.cpp
```diff
@@ -892,6 +842,31 @@ void mainForm::on_applyFIRFilterPushButton_clicked()
...
+}
+
+void mainForm::BwBitmapUndoStep::redo()
+{
+	undo(); // the same data swap
+}
+
+void mainForm::BwBitmapUndoStep::undo()
+{
+	Q_ASSERT(formPtr);
```
dg0yt: While it is not quite Qt-ish, I try to use references when I mean "not null" in new code. (I know Mapper is full of non-zero pointers...)

3rd-party/cove/app/mainform.h
```diff
@@ -51,6 +53,18 @@ class mainForm : public QDialog
 {
 	Q_OBJECT
 
+	struct BwBitmapUndoStep : public QUndoCommand
+	{
+		BwBitmapUndoStep() = delete;
+		BwBitmapUndoStep(mainForm* form, QImage image, bool vectorizable);
```
dg0yt: I try to follow the Rule of five or Rule of zero, in new code. Rule of zero seems possible here, and it would also be safe at least with a mainForm reference instead of a pointer.

3rd-party/cove/app/mainform.cpp
```diff
@@ -892,6 +842,31 @@ void mainForm::on_applyFIRFilterPushButton_clicked()
...
+	std::swap(formPtr->bwBitmap, image);
+	std::swap(formPtr->bwBitmapVectorizable, suitableForVectorization);
```
dg0yt: Note that idiomatic use of swap is:
~~~
using std::swap;
swap(formPtr->bwBitmap, image);
swap(formPtr->bwBitmapVectorizable, suitableForVectorization);
~~~
in order to find optimized specializations via [ADL](https://en.cppreference.com/w/cpp/language/adl). However,  QImage doesn't provide such a specialization (free function). It has a member function for swap.

3rd-party/cove/app/mainform.cpp
lpechacek: I see your point. I'll reconsider the use of the brace initialization, including zero init.

3rd-party/cove/app/mainform.cpp
lpechacek: Good catch. I was pulled into the use of pointer under the API. A reference is indeed the right tool. Thanks!

3rd-party/cove/app/mainform.h
lpechacek: Likewise. I tried to apply the rule of zero here but failed to initialize the structure without a constructor. Then applying the full rule of five felt overly noisy to me. I'll revisit this part.

3rd-party/cove/app/mainform.cpp
lpechacek: Understood. I didn't know this idiom. Thanks!

PR #1548, opened 2020-03-20
Update OpenOrienteering_zh_CN.ts

xianchb:
Chinese translation


xianchb:
I am from China. I want to create a Chinese Manual. What should I do? @dg0yt

lpechacek:
Answered in PR#1549, which is a duplicate of this one.

Issue #1552, opened 2020-03-22
Unexpected declination adjustment


dl3sdo:
### Steps to reproduce
1. import ocd (OCAD10) file
2. save as omap
3. open omap

### Actual behaviour
Mapper shows a warning message:
"Inconsistent declination/grivation detected. Declination adjusted from 0° to 0,01°."
Note: although the declination gets changed the map is not marked as having unsaved changes.

### Expected behaviour
I'm not sure whether this adjustment really needs to be done.

### Configuration
Mapper Version: 0.92 (x64)
Operating System: Win7x64

dg0yt:
Well, most likely it needs to be done earlier in this context. 0.01° sounds like a rounding issue.

dg0yt:
That map file is close to the central meridian of UTM Zone 32 (9°), in SW Germany. At this location, grid convergence is 0.01° indeed. So the declination must be different from the grivation.

Assuming that the map is drawn with regard to the current alignment of the templates and that the templates have stable rotation against true north (= grivation), it is reasonable to leave the map's grivation unchanged, and adjust the value of the declination instead, based on the calculated grid convergence, to establish consistency.
(If the calculated declination does not match real declination, the user may adjust that from the georeferencing dialog, rotating the map as needed.)

However, this adjustment is implemented for loading .OMAP files only, not for loading .OCD files. This may need work. However, I would still not trigger the flag for unsaved changes. Either you do other changes, or we leave the file unmodified.

dl3sdo:
Thank you for the explanation Kai.

dg0yt:
Need to verify the OCD import/export case. (Don't unnecessarily change the OCD file, but always write a valid OMAP file.)

PR #1565, opened 2020-03-31
OcdFileImport: Always set grivation

dg0yt:
We need to set the grivation even if it is the default value of zero.
This ensures that the declination is adjusted such that grivation,
declination and convergence are consistent.
Fixes GH-1552 (inconsistent import from OCD leading to later warning).


dg0yt:
Unlike OMAP format, OCD format doesn't explicit save declination. That's why there is no warning here: we silently establish the consistent state on every import. This is what the user assumes anyway.

PR #1575, opened 2020-04-12
Update CRT files for OSM

dg0yt:


dg0yt:
I will accept more suggestions until merging this just before v0.9.3.

Symbian9:
Does we already use all suggestions from related page on *OSM Wiki*?

- https://wiki.openstreetmap.org/wiki/IOF_mapping#IOF-Codes_.28ISOM.2C_ISSOM.2C_ISMTBOM.29_and_their_tags

FTR, It would be cool if for testing purpose there would be some complex *OSM-file* (`high-mountain-village.osm`) sample/testcase of good mapped town located within the mountains, because actual OSM testcase is useless:

- https://github.com/OpenOrienteering/mapper/blob/master/test/data/templates/map.osm

dg0yt:
1. The `ìof` tags might be added in a separate CRT file.
    But IMO, this is a dead horse. I have studied that ~10 years ago. (I created a print plugin for JOSM at that time.) The COMO documents date back to 2012.  I can't see these tags are used more now than at that time. There are only 3338 objects tagged with `iof`, most of them buildings, roads, paths:
    https://taginfo.openstreetmap.org/keys/iof
2. The `map.osm` is specifically crafted for an automated test and serves its purpose well. You can export complex files from openstreetmap.org at any time you like.
   A tutorial page or video would be more useful.

Note that OSM data has a license which I consider incompatible with most (paid) orienteering map projects: It requires you to provide derived "databasese" under share-alike terms to the public.

mlerjen:
I checked out the new OSM-ISOM 2017-2.crt and noticed that some lines refer to symbols I do not find in the according Symbol Set. Turns out that they all are there in the .omap template file defined in the symbol sets folder, but will not be imported to a "new map". Likely because they are toggled hidden in the symbol set template.

This leads to three issues.

1) Is it sensible, that hidden symbols from the symbol-set-template are not used for the "new map" at all?  IMO they should at least be imported as hidden if not automatically set to  not hidden.
image

2) In the ISOM 2017-2 symbol set there are still symbols "from ISOM 2000" e.g. "201.9 Impassable cliff, minimum size (from ISOM 2000)". These should be removed, shouldn't they?

3) The OSM-ISOM 2017-2.crt still refers to such ISOM 2000 artifacts as 201.9, 203.9 502.1. Because they are set to hidden in the template file, Mapper makes no use of them on the import.
image fig.2. Landuse=quarry symbolized with the default area symbol instead of 709 (even though symbol exists in ISOM 2017-2 template.

Can you tell me if these are "issues" and maybe propose how to structure these issues? Thx.

dg0yt:
>  Is it sensible, that hidden symbols from the symbol-set-template are not used for the "new map" at all?  IMO they should at least be imported as hidden if not automatically set to  not hidden.

The hidden symbols are not meant to be used for new maps. That's why they are removed when create a new map. OTOH, when you upgrade an existing map, there is no simple replacement with a new symbol. In this case, the hidden symbol would be used. 
Of course, there is no point in having them used by CRT files for OSM. 

But nobody complained that roads were broken...

mlerjen:
Thus:
Errors in the ISOM2017-2 symbol-set template:
- 201.2 should not be hidden 
- 701-713 should not be hidden

Errors in OSM-ISOM 2017-2.crt
- referring to 203.9 instead of 203.2
- referring to 502.1 instead of 502 
- referring to 406 instead of 405
- reference to 713  ... nevermind.

PR #1577, opened 2020-04-12
NMEA positioning fix and new static plugin

dg0yt:


dg0yt:
Now here is the additonal positioning (static) plugin. Tested with a FIFO (`mkfifo`) for feeding @gardners NMEA data. 

Maybe the QProcess stuff could be simplifed to QFile if not dealing with (blocking) FIFOs, but it is quite flexible. It might even be extended to `netcat` TCP NMEA providers.

src/sensors/nmea_position_plugin.cpp
```diff
@@ -0,0 +1,147 @@
+/*
+ *    Copyright 2019 Kai Pastor
```
dg0yt, line 2: 2020

src/sensors/nmea_position_plugin.cpp
```diff
@@ -0,0 +1,147 @@
...
+	
+	QString defaultDevicePath() const
+	{
+		auto const ports = QSerialPortInfo::availablePorts();
+		return ports.empty() ? QLatin1String("/dev/null") : ports.front().systemLocation();
+	}
+	
+	QString devicePath() const
+	{
+		// Path in NMEA_POSITION_DEVICE overrides settings.
```
dg0yt, line 103: Obsolete comment.

src/sensors/sensors_settings_page.cpp
```diff
@@ -87,7 +86,10 @@ void SensorsSettingsPage::apply()
 	
 	settings.setPositionSource(position_source_box->currentData().toString());
 	if (nmea_serialport_box)
-		settings.setNmeaSerialPort(nmea_serialport_box->currentText());
+	{
+		auto const port = nmea_serialport_box->currentText().trimmed();
+		settings.setNmeaSerialPort(port == tr("Default") ? QString{} : port);
+	}
```
dg0yt, line 92: Might be worth moving to an independent commit.

gardners:
I'll to clone it down and built it, to see if it works for me.
If it were not hard to allow setting the baud rate for the serial port, that would be a good insurance policy.  I can dig up example code for that, if required.

gardners:
Any ideas for dealing with this error for building on Ubuntu 18.04?

```
$ cmake src
CMake Error at CMakeLists.txt:292 (mapper_translations_sources):
  Unknown CMake command "mapper_translations_sources".


CMake Warning (dev) in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 3.10)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring incomplete, errors occurred!
See also "/home/paul/Projects/Arkaroola/orienteering/mapper/CMakeFiles/CMakeOutput.log".

```

dg0yt:
> If it were not hard to allow setting the baud rate for the serial port, that would be a good insurance policy. I can dig up example code for that, if required.

With Qt's plugin, the baud rate is hard wired, really:
https://code.qt.io/cgit/qt/qtlocation.git/tree/src/plugins/position/serialnmea/qgeopositioninfosourcefactory_serialnmea.cpp#n69

The code that I proposed doesn't deal with serial ports in particular, except for finding their device file path. Either we would need to access them via QSerialPort instead of reading from QProcess/`cat`, or we would need to run `stty` before starting QProcess/`cat`.

dg0yt:
> Any ideas for dealing with this error for building on Ubuntu 18.04?

You need to build all of the Mapper source code. This plugin is not standalone but integrated. 

In this way you can get a full debug build. This is what I use most of the time, with an fairly recent Qt Creator.

gardners:
```
git clone git@github.com:OpenOrienteering/mapper.git
git checkout positioning
cd mapper
cmake src
```

should work for building, right?

gardners:
I've found the pre-built binary from the automated tests, but I don't have libproj.so.18 or libgdal.so.26. If you have easy access to packages containing those, I should be able to test the build fairly quickly.  I'll keep looking in the meantime for solutions to those, but GDAL doesn't look particularly straightforward to build so far.

dg0yt:
> ... should work for building, right?

The top-level CMakeLists.txt is in the root directory. So in the end, there should be `cmake .` for an in-source build. But I would recommend an out-of-source build:

~~~
git clone -b positioning https://github.com/OpenOrienteering/mapper.git
mkdir build
cd build
cmake ../mapper
~~~

Cf. https://github.com/OpenOrienteering/mapper/blob/master/INSTALL.md for more information. All dependencies can be used directly from Ubuntu 18.04, it is really straightforward.

> I've found the pre-built binary from the automated tests, but I don't have libproj.so.18 or libgdal.so.26. If you have easy access to packages containing those, I should be able to test the build fairly quickly. I'll keep looking in the meantime for solutions to those, but GDAL doesn't look particularly straightforward to build so far.

These Ubuntu builds are just for covering certain configurations in the limited CI setup. I would not recommend using them for "production" machines. (The dependencies are available from the Superbuild pipeline.)

gardners:
Ok, I managed to make my way through the usual dependency hell and get a build done of the positioning branch. It brings back the fond memories of being a sysadmin for 13 years :)

Anyway, I can indeed now select the "NMEA (openorienteering)"  sensor source, and it actually works, so well done, and a +1 from me in favour of incorporating it.

Now, as for the name of the option, and its relationship to the existing "serialport (NMEA)", this is certainly a bit confusing as it currently stands. I'd suggest the new one be:

"Internal/serial GNSS receiver"  

and rename the old serialport (NMEA) to

"QT5 Positioning"

or simply remove it or mark it as "QT5 Positioning (deprecated)", since I can't see how it would work for anyone.  Even after I set my GPS to 4800, it wouldn't see it.

Now to find a bit of spare time to go outside and try it out in earnest!

Danke nochmals, dass ihr es so schnell und bequem gelöst habt. Es soll mein Leben bei Kartenmalen wirklich einfacher machen.

LG
Paul

src/sensors/sensors_settings_page.cpp
```diff
@@ -113,6 +115,9 @@ void SensorsSettingsPage::updateWidgets()
 		else if (display_name == QLatin1String("serialnmea"))
 			//: Position source
 			display_name = tr("Serial port (NMEA)");
+		else if (display_name == QLatin1String("NMEA (OpenOrienteering)"))
+			//: Position source
+			display_name = tr("NMEA (OpenOrienteering)");
```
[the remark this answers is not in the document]
dg0yt: I thought that "NMEA (OpenOrienteering)" was distinct and clear enough.

- We don't know if it is external...
- We might suppress "Serial port (NMEA)" completely. It doesn't do much than opening a QSerialPort as device, where we are using QProcess. 
  - We could use QSerialPort as an alternative to QProcess if the device name identifies a serial port. Than we might add the usual serial port settings (baud rate etc.).
  - We could simply use the same plugin identifier as Qt, with a higher priority.
- Note that the "Serial port (NMEA)" label is also given by Mapper (or: chosen by me), i.e. we might change it as well, e.g. to "NMEA (Qt)". (All we get from the plugin is "serialnmea").

src/sensors/sensors_settings_page.cpp
```diff
@@ -138,19 +143,25 @@ void SensorsSettingsPage::updateWidgets()
 	if (nmea_serialport_box)
 	{
 		nmea_serialport_box->clear();
+		nmea_serialport_box->addItem(tr("Default"));
+		nmea_serialport_box->setCurrentIndex(0);
+		
 		auto const current_port_name = Settings::getInstance().nmeaSerialPort();
-		auto add_serialport = [this, &current_port_name](const QString& display_name, bool is_default = false) {
```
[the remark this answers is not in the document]
dg0yt: Ack. I thought of this, too.

dg0yt:
For now, I would suggest, we simply offer the choice between 
"NMEA (Qt)"
and 
"NMEA (OpenOrienteering)".

I wonder if/how we could do something similar on Windows/macOS/Android, at least for accessing NMEA via Bluetooth connections.

gardners:
Sounds good to me.
Yes, I suspect that something similar should be possible for those platforms, it will just be working out how you wrap the BT connection.

lpechacek:
> I wonder if/how we could do something similar on Windows/macOS/Android, at least for accessing NMEA via Bluetooth connections.

Apologies upfront for the long post. TL;DR Bluetooth connections can be made from Qt and I consider the current code structure good for future expansion.

For me, this is a much broader topic after messing with GNSS receiver design for about a year. I don't want to overwhelm you with my ideas but allow me to dump my ideas at this point. External devices (GNSS receiver, laser ranger, ...?) can be connected by:
1) Cable - simple, controlled, the user knows when it's on and when it's off, the cable leads only from here to there and consumes essentially no power. The downsides are potentially fragile connectors and the cable hooking on branches in thickets. It is also unclear how are serial ports handled in Android but SI-Droid and PPM Commander applications are using such cable connections so there is some way to making it work.
2) Bluetooth 2.0 RFCOMM - still reasonably simple configuration, energy-efficient, no problems with connectors and dangling cables. Bluetooth, however, becomes hell when it comes to debugging. Also, the RFCOMM implementation in Linux kernel is somewhat buggy these days. On Android RFCOMM works smooth, maybe thanks to a userspace implementation.
4) Bluetooth 4.0+ Low Energy - pros as in the previous point plus even lower energy consumption. The cons include the removal of RFCOMM (serial port over BT) protocol from the standard. As a result, software vendors do not feel obliged to provide "serial port over Bluetooth" implementation in their products and hardware vendors do not want to invest effort into providing OS-specific drivers for their devices. In the absence of a standard protocol, at least two proprietary (but documented) protocols emerged - [Texas Instruments](http://www.ti.com/lit/pdf/tidua63) and [u-blox](https://www.u-blox.com/en/docs/UBX-16011192). In any case, communication to BLE devices is now handled in userspace.
4) WiFi - I have no experience with this one but it also may work fine. I suspect configuration might be more difficult and I would expect higher energy consumption.

For me, all the above boils down to making the various connection methods a descendant of `QIODevice` (and `QBluetoothSocket` and `QTCPSocket` already are) and feeding data to a positioning plugin that will parse NMEA 0183 and possibly other protocols. For example, there is [`PUBX,00` message from u-blox receiver](https://www.u-blox.com/en/docs/UBX-19035940) indicating the receiver's idea of estimated position error and there is `GST` sentence indicating pseudorange errors. I guess other receivers are able of producing similar information and Mapper can use that for the fix accuracy indication instead of the current `HDOP*<magic number>` calculation. This is why I welcome the current structure in the code as it allows expansion into the areas drafted above.

Ad the laser rangefinder use case, I posses a [LTI TruPulse TP360B](http://wiki.awf.forst.uni-goettingen.de/wiki/images/7/71/TruPulse_360-B_Users_Manual_2nd_Edition_English.pdf) and would need `QBluetoothSocket` on Android. So far I figured out that it means adding `bluez` and `libqt5-qtconnectivity` to Superbuild. I've made one attempt already, QtConnectivity part went well but BlueZ still resists to building there. Then connecting to the laser rangefinder is a matter of:
```
	bt_socket = new QBluetoothSocket(QBluetoothServiceInfo::RfcommProtocol, this);
	bt_socket->connectToService(QBluetoothAddress("00:1B:35:00:50:E6"),
	                            QUuid("{00001101-0000-1000-8000-00805f9b34fb}"));
	
	connect(bt_socket, &QBluetoothSocket::readyRead, this, &MainWindow::readSocket);
	connect(bt_socket, &QBluetoothSocket::connected, this, QOverload<>::of(&MainWindow::connected));
	connect(bt_socket, &QBluetoothSocket::disconnected, this, &MainWindow::disconnected);
	connect(bt_socket, QOverload<QBluetoothSocket::SocketError>::of(&QBluetoothSocket::error),
	        this, &MainWindow::onSocketErrorOccurred);
	connect(bt_socket, &QBluetoothSocket::stateChanged, this, &MainWindow::socketStateChanged);
```
I've got a standalone demo working on Linux and I have an idea about the user interface in Mapper.

gardners:
Having worked on wireless mesh networks a lot (see servalproject.org), BT is a royal pain compared with Wifi.  The WiFi power consumption can be dealt with by making the sensor be the wifi master, and have a bigger battery (it will be hungier, anyway, due to the sensor package), and have the phone connect as an energy conserving wifi client.  This also gives you regular old TCP etc for convenient transport modes in the code side of things.  There are plenty of cheap WiFi APs with serial ports and GPIOs, e.g., GL-iNET AR-150, that could be used as a basis for this.  Happy to help out with this a bit, if that would be useful.

dg0yt:
As as a licensed radio amateur, I'm very interested in these GPS, BT, Wifi topics, but it is out of scope here.

Finished changes to the PR:
- split out independent changes
- added a test (with your NMEA data @gardners)
- added macOS support
- made QSerialPort dependency optional
- changed Qt serialport displayname to "NMEA (Qt)"

gardners:
Understood re off-topic. Maybe chuck me an email.

lpechacek:
> Having worked on wireless mesh networks a lot (see servalproject.org), BT is a royal pain compared with Wifi.

I intended to share what connection types are being used out there in the wild and not to spark an off-topic discussion. In the hindsight, I should have listed devices using the individual connection types rather than listing pros and cons. Sorry for the noise.

Issue #980, opened 2017-09-30
Rectangle around selected objects and nodes clash


ollesmaps:
### Steps to reproduce
In Edit objects mode you easily move object even if you want to move just a node.  Start, finish (and other nodes on rectangle) do problems.

### Expected behaviour
In Edit objects mode put the rectangle around selected objects a little bit further so you can better differentiate between moving the whole object and moving only a node

### Configuration
Android App 0.7.0

dg0yt:
There is a setting which mitigates the issue: "Editor" > "Click tolerance". If you increase that value, the area which will select the individual node instead of the object frame will be larger.
- Of course this affects other interactions, too.
- Please verify that the dimension is correctly used even on high-dpi screens.
- Once we start handling touch different from mouse input, we may use a different value for touch input.

I admit that there is no visual clue for touch input. For mouse input, there are hover effects.

Zerbembasqwibo:
But the same should be used for mouse interaction, why not?

ollesmaps:
Thank you for the advice @dg0yt. I settled on click tolerance 5mm and it works fine now. Thank you. Just for the record my original click tolerance was 3mm. I tried also higher values (7mm), however that was too much as it was making me select objects involuntarily.

ollesmaps:
@dg0yt The only downside of the higher click tolerace I notice is when adding/ deleting nodes. It is less sensible now.

ollesmaps:
This causes troubles also reversely. If you want to just move object, you can move node instead. I think the click tolerance adjustment works for move object / move node in opposite ways. So another solution shall be applied.

PR #1587, opened 2020-04-21
EditPointTool,EditLineTool: Enlarge object frame (GH-980)

lpechacek:
There is a heavy code duplication between edit point and edit line tools. The work done in the `EditTool` code aims at minimizing further code duplication.

### DIFF
--- a/src/tools/edit_line_tool.cpp
+++ b/src/tools/edit_line_tool.cpp
@@ -429,6 +429,11 @@ int EditLineTool::updateDirtyRectImpl(QRectF& rect)
 	
 	selection_extent = QRectF();
 	map()->includeSelectionRect(selection_extent);
+
+	const auto frame_extension = 0.001 * cur_map_widget->getMapView()
+	                             ->pixelToLength(clickTolerance()); // in mm
+	selection_extent.adjust(-frame_extension, -frame_extension,
+	                        frame_extension, frame_extension);
 	
 	rectInclude(rect, selection_extent);
 	int pixel_border = show_object_points ? pointHandles().displayRadius() : 1;
@@ -622,6 +627,13 @@ void EditLineTool::updateHoverState(const MapCoordF& cursor_pos)
 		effective_start_drag_distance = (hover_state == OverNothing) ? startDragDistance() : 0;
 		updateDirtyRect();
 	}
+}
+
+
+void EditLineTool::applyViewChanges(MapView::ChangeFlags change)
+{
+	if (change == MapView::ZoomChange)
+		updateDirtyRect(); // the bounding rectangle extent changes
 }
 
 
--- a/src/tools/edit_tool.h
+++ b/src/tools/edit_tool.h
@@ -34,6 +34,7 @@
 #include <QString>
 
 #include "core/map_coord.h"
+#include "core/map_view.h"
 #include "tools/tool.h"
 #include "tools/tool_base.h"
 
@@ -148,6 +149,16 @@ protected:
 	 * An utility implementing object selection logic.
 	 */
 	QScopedPointer<ObjectSelector> object_selector;
+
+public slots:
+	/**
+	 * Rendering of some of the editor graphic elements may depend on
+	 * MapView settings. This method is the place where to process
+	 * the changes.
+	 *
+	 * @param change type of change as emitted by MapView
+	 */
+	virtual void applyViewChanges(OpenOrienteering::MapView::ChangeFlags change) = 0;
 };
 
 
--- a/src/tools/edit_line_tool.h
+++ b/src/tools/edit_line_tool.h
@@ -29,6 +29,7 @@
 #include <QString>
 
 #include "core/map_coord.h"
+#include "core/map_view.h"
 #include "tools/edit_tool.h"
 
 class QAction;
@@ -68,6 +69,8 @@ public:
 	void dragMove() override;
 	void dragFinish() override;
 	void dragCanceled() override;
+	
+	void applyViewChanges(OpenOrienteering::MapView::ChangeFlags change) override;
 	
 protected:
 	bool keyPress(QKeyEvent* event) override;
--- a/src/tools/edit_point_tool.cpp
+++ b/src/tools/edit_point_tool.cpp
@@ -575,6 +575,11 @@ int EditPointTool::updateDirtyRectImpl(QRectF& rect)
 	
 	selection_extent = QRectF();
 	map()->includeSelectionRect(selection_extent);
+
+	const auto frame_extension = 0.001 * cur_map_widget->getMapView()
+	                             ->pixelToLength(clickTolerance()); // in mm
+	selection_extent.adjust(-frame_extension, -frame_extension,
+	                        frame_extension, frame_extension);
 	
 	rectInclude(rect, selection_extent);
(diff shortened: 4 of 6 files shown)


src/tools/edit_line_tool.cpp
```diff
@@ -429,6 +429,11 @@ int EditLineTool::updateDirtyRectImpl(QRectF& rect)
 	
 	selection_extent = QRectF();
 	map()->includeSelectionRect(selection_extent);
+
+	const auto frame_extension = .001 * cur_map_widget->getMapView()
```
dg0yt, line 433: I prefer to not drop the leading 0 in 0.001. (It is technically possible in C++ but uncommon in general.)

src/tools/edit_tool.h
```diff
@@ -144,10 +145,28 @@ Q_OBJECT
...
 	QScopedPointer<ObjectSelector> object_selector;
+
+public slots:
+	/**
+	 * This slot forwards viewChange notifications from MapView to the
+	 * sub-class specific implementation.
+	 * 
+	 * @param change 
+	 */
+	void applyViewChanges(OpenOrienteering::MapView::ChangeFlags change);
```
dg0yt, line 169: I'd prefer to make this slot a virtual method, with a default implementation which does nothing ATM, and with an explanation that overriding implementations shall call the parent's implementation.

Although common in Mapper, it is sometimes not really clear what problem is solved by the "impl" approach. I suppose the common real use case is when you want to have efficient variation in some part of an algorithm. This is where Qt uses virtual "event" methods (e.g. paintEvent). But having slots and (separate) virtual methods appears to be a lot of indirection.

src/tools/edit_line_tool.cpp
```diff
@@ -429,6 +429,11 @@ int EditLineTool::updateDirtyRectImpl(QRectF& rect)
...
+	                             ->pixelToLength(clickTolerance()); // in mm
+	selection_extent.adjust(-frame_extension, -frame_extension,
+	                        frame_extension, frame_extension);
 	
 	rectInclude(rect, selection_extent);
 	int pixel_border = show_object_points ? pointHandles().displayRadius() : 1;
```
dg0yt, line 439: Instead of adjusting the selection rectangle, we may simple increase the returned pixel border - this uses the same unit clickTolerance(). The common rationale is: `selection_extent` is from the data, `pixel_border` is from the tool and view.
~~~
int pixel_border = show_object_points ? std::max(qRound(clickTolerance()), pointHandles().displayRadius()) : 1;
~~~

Additional remark:
If you adopt this change, you might 
either eliminate this variable and move the expression to the return statement, 
or start with `int pixel_border = 1;`, and put another assignment into the `if (show_objects_points)` block.

dg0yt: Bonus: Probably the whole view change slot machinery becomes obsolete by moving to the pixel_border approach.

lpechacek: As far as I can tell, it is the `selection_extent` rectangle which defines the purple rectangle around the object at `/data/src/oomapper/src/tools/edit_point_tool.cpp:623`. But there is a valuable point in your remark. The purple rectangle should be perhaps extended by pixel_border as well to give even more breathing room to the outer vertices.
image

lpechacek: The code would be like this:
```
	const auto pixel_border = show_object_points ? pointHandles().displayRadius() : 1;
	const auto frame_extension = 0.001 * cur_map_widget->getMapView()
	                             ->pixelToLength(pixel_border + clickTolerance()); // in mm
	selection_extent.adjust(-frame_extension, -frame_extension,
	                        frame_extension, frame_extension);

	rectInclude(rect, selection_extent);
<snip>
	return 0; // selection_extent covers the border points
```
OK?

src/tools/edit_line_tool.cpp
lpechacek: Ack. Likely another bad habit from my C practice. :)

src/tools/edit_tool.h
lpechacek: Right. A virtual method was one of my first thoughts but for some reason, I thought it cannot be used as a slot. It turns out that it was just my careless reading of the Signals & Slots chapter in Qt documentation, which explicitly mentions this option. I will fix this part.

src/tools/edit_line_tool.cpp
lpechacek: The unconditional inclusion of `pixel_border` is not perfect either as in most cases the room expansion for point handles is unnecessary because sufficient space is ensured by the object renderables themselves. IOW, most objects are massive enough to cover the point handles as in the below example rendered *without* the patch.
image

At this point, however, I think the analysis is going into absurd dimensions and that perfect is the enemy of good. So, the question now is whether or not to include `pixel_border` into the purple frame. I'd leave it out as `clickTolerance()` already gives a sufficient margin and flat line endpoints are essentially a corner case.

dg0yt: I had only a few minutes to test and check, and I seem to have failed in two ways: get the broad picture of the surrounding code, and actually testing the modifications which I meant to test. I'm sorry for the confusion.

lpechacek: No p
No problem. On my side, it took me half a day to verify that the conversion of click tolerance value is correct. This code is quite complex.

lpechacek:
The CI checks are failing. I'll have a look at it tomorrow.

src/tools/edit_line_tool.cpp
dg0yt: > This code is quite complex.

Yes. It is relying too much on side effects. It didn't really expect updateDirtyRect to update the box size, but I think I see why it is done. We need to mix two domains, map coordinates (mm) and screen coordinates (px). I would like to try another implementation approach soon (and I wouldn't know where to start without your PR). I just cannot switch branches two often while I'm working on 1561 (vector data georeferencing).

dg0yt:
I re-run the failed jobs.

src/tools/edit_line_tool.cpp
lpechacek: FWIW, the API is a pain to work with. I would have suggestions for rework and I'd be willing to implement them. I'm, however, unsure if such intrusive changes would be welcome.

> I just cannot switch branches two often...

`git worktree` or `git stash` might help with work on multiple branches. I'll wait for your remarks on the alternatives.

dg0yt: > FWIW, the API is a pain to work with. I would have suggestions for rework and I'd be willing to implement them. I'm, however, unsure if such intrusive changes would be welcome.

In general, yes. However, we should dicuss the suggestions first. (My hesitating with this PR comes from the fact that it adds more signals/slots, and I feel like it should be doable with regular interfaces for these classes because they are designed to work together closely.)

> > I just cannot switch branches two often...
> 
> `git worktree` or `git stash` might help with work on multiple branches. 

The problem are the build directories, not the source directories. I build and test multiple configurations.

dg0yt:
I'm going to merge this as is now. The benefit of getting away from the extra signals/slots isn't big enough to justify delaying this work from reaching a public build.

PR #1222, opened 2019-03-26
Update georeferencing dialog

dg0yt:

PR #1593, opened 2020-04-26
Georeferencing: Fix state switching

dg0yt:
Fixes the change of mapping between paper and ground when changing from a CRS to a local georeferencing and back, by retaining (fake) convergence and grid scale factor.

### DIFF
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -489,6 +489,8 @@ void Georeferencing::load(QXmlStreamReader& xml, bool load_scale_only)
 			if (auxiliary_scale_factor <= 0.0)
 				throw FileFormatException(tr("Invalid auxiliary scale factor: %1").arg(QString::number(auxiliary_scale_factor)));
 		}
+		grid_scale_factor = combined_scale_factor / auxiliary_scale_factor;
+		
 		if (georef_element.hasAttribute(literal::declination))
 			declination = roundDeclination(georef_element.attribute<double>(literal::declination));
 		if (georef_element.hasAttribute(literal::grivation))
@@ -496,6 +498,7 @@ void Georeferencing::load(QXmlStreamReader& xml, bool load_scale_only)
 			grivation = roundDeclination(georef_element.attribute<double>(literal::grivation));
 			grivation_error = georef_element.attribute<double>(literal::grivation) - grivation;
 		}
+		convergence = declination - grivation;
 		
 		while (xml.readNextStartElement())
 		{
@@ -686,7 +689,6 @@ void Georeferencing::setState(Georeferencing::State value)
 		if (state != Normal)
 		{
 			setProjectedCRS(QStringLiteral("Local"), {});
-			updateGridCompensation();
 		}
 		
 		emit stateChanged();
@@ -1022,7 +1024,7 @@ bool Georeferencing::setProjectedCRS(const QString& id, QString spec, std::vecto
 			if (ok && state != Normal)
 				setState(Normal);
 		}
-		if (!ok)
+		if (isValid() && !isLocal())
 			updateGridCompensation();
 		
 		emit projectionChanged();
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -891,13 +891,8 @@ void Georeferencing::updateGridCompensation()
 	if (determinant < 0.00000000001)
 		return;
 
-	// This is the angle between true azimuth and grid azimuth.
-	// In case of deformation, the convergence varies with direction and this is an average.
 	convergence = qRadiansToDegrees(atan2(d_northing_dx - d_easting_dy,
 	                                      d_easting_dx + d_northing_dy));
-
-	// This is the scale factor from true distance to grid distance.
-	// In case of deformation, the scale factor varies with direction and this is an average.
 	grid_scale_factor = sqrt(determinant);
 }
 
--- a/src/core/georeferencing.h
+++ b/src/core/georeferencing.h
@@ -637,12 +637,29 @@ private:
 	unsigned int scale_denominator;
 	double combined_scale_factor;
 	double auxiliary_scale_factor;
+	
+	/**
+	 * This is the scale factor from true distance to grid distance.
+	 * 
+	 * In case of deformation, the scale factor varies with direction and this is an average.
+	 * 
+	 * \see updateGridCompensation()
+	 */
 	double grid_scale_factor;
 
 	double declination;
 	double grivation;
 	double grivation_error;
+	
+	/**
+	 * This is the angle between true azimuth and grid azimuth.
+	 * 
+	 * In case of deformation, the convergence varies with direction and this is an average.
+	 * 
+	 * \see updateGridCompensation()
+	 */
 	double convergence;
+	
 	MapCoord map_ref_point;
 	
 	QPointF projected_ref_point;
--- a/test/georeferencing_t.cpp
+++ b/test/georeferencing_t.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2012-2015, 2019 Kai Pastor
+ *    Copyright 2012-2020 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -228,9 +228,8 @@ void GeoreferencingTest::testGridScaleFactor()
 	{
 		QCOMPARE(geod_distance, elevation_scale_factor * ground_distance);
 	}
-
-	// Finally, see that the auxiliary scale factor is preserved when
-	// the CRS changes.
+	
+	// See that the auxiliary scale factor is preserved when the CRS changes.
 	georef.setProjectedCRS(QString::fromLatin1(QTest::currentDataTag()), utm32_spec);
 	QVERIFY2(georef.isValid(), georef.getErrorText().toLatin1());
 	QCOMPARE(georef.getAuxiliaryScaleFactor(), elevation_scale_factor);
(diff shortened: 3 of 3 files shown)


dg0yt:
@pkturner This is touching grid compensation. May I kindly ask you to review this change? Testing instructions are in the last commit.

pkturner:
@dg0yt I'll review it today (UTC-4).

pkturner:
@dg0yt I'm not seeing that a fix is needed. I followed the manual test instructions in the 9e729446 commit with v0.9.2, and after changing back to EPSG 2180 the template positioning looked good and the projected coordinates of a corner point were the same as originally. I also ran the georeferencing_t with the "Verify state switching" change on a Mapper built without the "Fix state switching" commit, and it passed.

Perhaps I'm getting different results from you as I'm on Linux with proj-6.3.0 and gdal-3.0.4.

However, I also think it may not be appropriate to carefully save and restore the original grid scale factor. When you switch to Local, Mapper still saves and respects the auxiliary_scale_factor. Then  when EPSG 2180 is reinstated, the grid scale factor is re-calculated with the same value as before, and Mapper calculates the combined_scale_factor getting the same result as before.

This is a good way to find the combined_scale_factor, particularly because the CRS might be some other projection than EPSG 2180. In that case the grid scale factor would be different, but the auxiliary_scale_factor is still valid and probably accurate enough. So the combined_scale_factor needs to be calculated from the retained auxiliary_scale_factor and the newly calculated grid_scale_factor to handle the general case of a new CRS.

pkturner:
Somewhat off topic, the case I just mentioned has been on my mind a bit. When switching back to Normal state from Local, if the CRS is now different, can Mapper help the user get back the former geographic coordinates of the reference point? My inexperienced expectation is that it would usually work out poorly to set the geographic coordinates from the projected coordinates, if the latter came from a different CRS. It would make sense to me to have preserved some geographic coordinates and use them to update the projected coordinates.

dg0yt:
Maybe I should have been a little clearer: The fix is meant to improve the situation immediately after switching georeferencing to local: I would neither want to rotate/scale the map contents nor would I expect any changed projected coordinates, but the latter is what happens with released version of Mapper.

The geographic coordinates are an interesting hint. Maybe we should just disallow switching to non-georeferenced local, and use a local stereographic setup instead.

dg0yt:
Original SE corner point coordinates:


After setting georeferencing to "Local", Mapper v0.9.2:


-> Coordinates of SE corner point changed (paper coordinates unchaged). Note that the georeferenced template followed the coordinate system change, in scale and rotation.

This PR is meant to retain coordiates as in the first picture. (And of course the same for switching back to the original CRS.)

pkturner:
Sorry I misunderstood. That's a cool improvement.

Regarding the code, it's not necessary to save the grid scale factor, because it can be recovered from the combined scale factor and auxiliary scale factor, just as the convergence is recovered from the declination and grivation.

I uploaded a suggested patch to my GitHub pkturner/mapper repository, on 'georeferencing' branch with commit c3564c31. It is simpler, and also preserves the parallel between
declination - convergence - grivation, and
auxiliary scale factor - grid scale factor - combined scale factor

dg0yt:
Thanks for the scale factor suggestion, which I adopted with a modification: Do it unconditionally, like we do with the rotation parameters, for full analogy.

Now I managed to isolate the relevant behaviour from `GeoreferencingDialog` which is fixed by the change to `Georeferencing`, and put it into `GeoreferencingTest`: 
The change to georeferencing state doesn't immediately lead to a wrong transformation. But rotation and scale parameters are put into an inconsistent state which breaks the transformation as soon as you attempt a noop on e.g. declination or auxiliary scale factor.

So now the modified test fails before the fix, and succeeds after.

:+1:

pkturner:
It looks good.

pkturner:
Today I was looking at this again. I believe that for the purposes of this change, the call to `updateGridCompensation` can be removed from `setProjectedCRS`. Back when I was working on grid compensation I examined that odd-looking `if(!ok)`, tracing various paths through the code, and found that normally the caller of `setProjectedCRS` ensures that `updateGridCompensation` gets called afterwards.

dg0yt:
>  I examined that odd-looking if(!ok), tracing various paths through the code, and found that normally the caller of setProjectedCRS ensures that updateGridCompensation gets called afterwards.

I tried to verify that claim, and I found that I should take some notes:

| Called from | ensures updateGridCompensation afterwards by |
|----|----|
| `Georeferencing::setState` | *Actually relying on `setProjectedRefPoint()`.* `set...RefPoint() after some calls to `setState()` |
| `ocd_georef_fields.cpp`:`applyGridAndZone` | `setProjectedRefPoint()` after call to `applyGridAndZone()` |
| `OgrTemplate::importGeoreferencing` x2 | `setProjectedRefPoint()` |
| `ogr_template.cpp`:preserveRefPoints` | `setProjectedRefPoint()` |
| `OgrTemplate::loadTemplateFileImpl` | `setProjectedRefPoint()` |
| `GeoreferencingDialog::crsEdited` |  `setState()`, `set...RefPoint()` |
| `SelectCRSDialog::updateWidgets` |   pointless: no spatial location |
| `TemplateImage::postLoadConfiguration` |   `setProjectedRefPoint()`, dialog |
| `TemplateImage::findAvailableGeoreferencing` |   pointless: no spatial location |
| `TemplateImage::calculateGeoreferencing` |  pointless: `setTransformationDirectly()` |
| `TemplateTrack::applyProjectedCrsSpec` | `setGeographicReferencePoint()` |
| test/*_t.cpp | not relevant |

So the claim seems valid for most spots, but there are some corner cases around `setState` which might get tricky.

- I would suggest to leave the statement as is for now. It doesn't do any harm.
- As part of a Georeferencing redesign, we shall look for a better way. 
  (One idea I have is to put the explicit property changes (`setXy()`) in some kind of transaction object, and trigger automatic updates and signal emissions at the end of the transaction.)

pkturner:
@dg0yt 
The thought of retaining the geographic reference point came back, and I realized that it provides an interesting alternative approach to addressing this problem. For what it's worth, I tried out some code, now uploaded to my pkturner/mapper repository on the "switch-state" branch, commit 73c20c44.

This approach does not retain alignment of the map with template while the state is local, but uses the saved geographic reference point to restore the alignment when the CRS is restored and the state is again normal. Your addition to `georeferencing_t` is trimmed back to only test the properties after the state is restored.

For some of the behavior, the code needs to test whether the `Georeferencing::geographic_ref_point` holds a meaningful value. Rather than use `std::optional` or a boolean flag, this experimental code simply treats `LatLon{0,0}` as the indication that there is no meaningful geographic reference point.

dg0yt:
> This approach does not retain alignment of the map with template while the state is local, but uses the saved geographic reference point to restore the alignment when the CRS is restored and the state is again normal.

I feel like the alignment should be preserved while the state is local - this is the least surprise to the user. 

The geographic reference point could be kept in addition to that - it still gives some spatial reference. On the other hand I'm not sure if I want to add so much spatial reference to the "local georeferencing". If I do have Lat/Lon, I could immediately use a local stereographic projection.

pkturner:
> I feel like the alignment should be preserved while the state is local - this is the least surprise to the user.

OK with me. I mentioned the difference to make it clear how the code that retains the geographic reference point behaves differently.

pkturner:
Yes, given Lat/Lon and declination, the local state could be replaced with a more capable local stereographic projection. Without experience of a use case for switching to local state, I cannot judge that option one way or the other.

pkturner:
@dg0yt
I've been looking at this change further, with a view to understanding "Grivation locked" better.
The primary sense of "grivation locked" may be that when georeferencing is in the local state, grivation takes precedence over declination, and changes to convergence should have no impact on the map.

So it may be that a better fix for state switching would undo part of the change way back in "Update georeferencing dialog" [GH-1222](https://github.com/OpenOrienteering/mapper/pull/1222). There we changed `GeoreferencingDialog::accept` to force the grivation to fit the declination.

pkturner:
I was mulling over the thought of providing a local stereographic projection, perhaps automatically, if the user has a valid CRS and then chooses -local-. If there's a scale, a declination, and a geographic reference point, then you really do have as good as a projection. So why would the user choose -local- when they've got a projection?

I went ahead with a presumption that the user chooses -local- for the purpose of dropping some of that baggage, in particular the declination, scale factor, or the tie to a geographic reference point. If that were the case, what business does Mapper have, when loading a map that's in Local state, with calculating a _convergence_? In Local state, declination has no part in the transformation which places objects and templates on the paper -- it rests directly on grivation. And yet prior to the GH-1593 fix, grivation was being modified in _Local_ state. In the above-mentioned use case involving `world-file.xmap`, it is in `GeoreferencingDialog::accept` that the grivation is modified.

This idea that `GeoreferencingDialog::accept` ought not to modify the grivation (in Local state) provides a different approach to taking care of this use case. Simply restrict that method to not call `updateGrivation` in that situation. So I developed that, and have uploaded another patch to the GitHub pkturner/mapper repository, on 'grivation-locked' branch with commit [af5753ed](https://github.com/pkturner/mapper/commit/af5753ed8e25f25398cce7cec7009ffc6a595d5b).

Beyond the code change, I've modified the GH-1593 addition to `georeferencing_t` in a way which is consistent with this approach. I've tested by hand the above-mentioned scenario of `world-file.xmap`. I've tested by hand that this patch preserves the fix provided by GH-1222 (which is the PR that caused this behavior with `world-file.xmap` to regress).

I'm embarrassed to submit yet another offer of an alternative fix, at the late date when the existing fix is reasonable and is about to be released. I do believe that this presents an insightful and worthwhile point of view.

dg0yt:
I'm fine with tuning this further. However I feel like the principle of least surprise should not only apply for the dialog (user experience) but also for the programming interface (developer experience): I came to look at this code while dealing with georefererencing of templates. The test code explains what I assume as a developer.

Note that you may also look at #1604.

PR #1598, opened 2020-05-01
Georeferencing: Use direct pipeline between CRS

dg0yt:
@pkturner This is how using modern PROJ, i.e. direct transformation between CRS, would look like, eliminating the need to commit to a particular geographic CRS for coordinate transformations, and even considern the area of interest.

This is just the principle. Creating the pipeline and area object for each single transformation is too expensive. But probably it is good enough for verification and experimentation. 

In the end, we might turn `Georeferencing` into something which doesn't bind the projected CRS to WGS84 at all. We would instead have all secondary georeferencings (template, GPS, Lat/Lon display) provide a pipeline to the map's primary georeferencing.


dg0yt:
On `proj_create_crs_to_crs`:
- https://proj.org/development/reference/functions.html#c.proj_create_crs_to_crs
- https://proj.org/operations/operations_computation.html

src/core/georeferencing.cpp
```diff
@@ -325,6 +325,26 @@ ProjTransform& ProjTransform::operator=(ProjTransform&& other) noexcept
...
+	auto target_latin1 = target.toLatin1();
+#ifdef PROJ_ISSUE_1573
+	// Cf. https://github.com/OSGeo/PROJ/pull/1573
+	source_latin1.replace("+datum=potsdam", "+ellps=bessel +nadgrids=@BETA2007.gsb");
+	target_latin1.replace("+datum=potsdam", "+ellps=bessel +nadgrids=@BETA2007.gsb");
+#endif
+	auto* area = proj_area_create();
+	proj_area_set_bbox(area,
+	                   reference.longitude()-0.1, reference.latitude()-0.1,
+	                   reference.longitude()+0.1, reference.latitude()+0.1);
```
dg0yt, line 342: TODO: Handle bounds.

pkturner:
This code change would actually have acceptable efficiency, because the affected variant of toMapCoordF is only called when there's a _change_ to georeferencing, such as when a map is imported.

pkturner:
> This code change would actually have acceptable efficiency, because the affected variant of toMapCoordF is only called when there's a _change_ to georeferencing, such as when a map is imported.

This is true for _some_ template types, but for others where `toMapCoordF` is called for every map object, it is unacceptable to incur the overhead of constructing the direct CRS-to-CRS transform every time.

pkturner:
> we might turn `Georeferencing` into something which doesn't bind the projected CRS to WGS84 at all. We would instead have all secondary georeferencings (template, GPS, Lat/Lon display) provide a pipeline to the map's primary georeferencing.

The changes underway in #1789 and #1814 bring the prospect of Mapper supporting geographic CRSes beyond WGS84. Ideally, Mapper would handle various geographic CRSes on an equal footing. Using the direct CRS-to-CRS transform efficiently is the essential step toward that goal.

@dg0yt I expect you already have an implementation structure in mind, and a code contribution from myself would not be much help.

dg0yt:
> This is true for _some_ template types, but for others where `toMapCoordF` is called for every map object, it is unacceptable to incur the overhead of constructing the direct CRS-to-CRS transform every time.

Yes. This is why it did't want to propose it for inclusion as is (cf. first comment).

> The changes underway in #1789 and #1814 bring the prospect of Mapper supporting geographic CRSes beyond WGS84. Ideally, Mapper would handle various geographic CRSes on an equal footing. Using the direct CRS-to-CRS transform efficiently is the essential step toward that goal.

#1814 is somewhat independent, fixing inconsistencies and improving testing of existing code.

I'm close to updating #1789, making it ready for inclusion. This shall lead to direct transformations on loading data, independent of this PR here, and also independent of #1814 at least for `OgrTemplate`. I will try to include similar capabilities for `TemplateTrack`. This must be tested.

PR #1604, opened 2020-05-10
Keep geographic reference point

dg0yt:
Edited cherry-pick of https://github.com/OpenOrienteering/mapper/commit/73c20c44d335db6496255652756d2de36e4365b6 on top of PR GH-1593.

### DIFF
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -654,7 +654,7 @@ void Georeferencing::save(QXmlStreamWriter& xml) const
 		}
 	}
 	
-	if (state == Normal)
+	if (state == Normal || geographic_ref_point != LatLon())
 	{
 		XmlElementWriter crs_element(xml, literal::geographic_crs);
 		crs_element.writeAttribute(literal::id, literal::geographic_coordinates);
@@ -836,7 +836,7 @@ void Georeferencing::updateGridCompensation()
 	convergence = 0.0;
 	grid_scale_factor = 1.0;
 
-	if (state != Normal || !isValid())
+	if (state != Normal || !isValid() || geographic_ref_point == LatLon())
 		return;
 
 	const double delta = 1000.0; // meters
@@ -909,7 +909,7 @@ void Georeferencing::setGeographicRefPoint(LatLon lat_lon, bool update_grivation
 		
 		bool ok = {};
 		QPointF new_projected_ref = toProjectedCoords(lat_lon, &ok);
-		if (ok && new_projected_ref != projected_ref_point)
+		if (ok)
 		{
 			projected_ref_point = new_projected_ref;
 			updateGridCompensation();
@@ -918,9 +918,8 @@ void Georeferencing::setGeographicRefPoint(LatLon lat_lon, bool update_grivation
 			if (update_scale_factor)
 				updateCombinedScaleFactor();
 			updateTransformation();
-			emit projectionChanged();
 		}
-		else if (geo_ref_point_changed)
+		if (geo_ref_point_changed)
 		{
 			emit projectionChanged();
 		}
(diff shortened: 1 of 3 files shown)


dg0yt:
@pkturner This is "geographic reference point preservation" on top of the local state switching PR.

pkturner:
I rebased the 2 existing commits on top of #1621 in master. Then I added two more for what they're worth, as drafts.

1. Georeferencing: Retain geographic reference point
2. GeoreferencingDialog: Unify geographic reference point updates
3. Georeferencing: Eliminate use of convergence when state is local
4. Georeferencing: Use enum for update args in calls to setXxxxRefPoint

These can be found in my pkturner/mapper GitHub repository on the reference-point branch, with the 4th commit being c2969f16.

Number 4 is my take on @lpechacek 's suggestion to pass the `update_grivation` argument using an enumerator. It is enhanced so that the caller can specify not just updating grivation to match declination, but also the other way around: updating declination to match grivation. This turns out to apply in several locations.

Number 3 applies the idea that convergence is meaningless when the state is local, and uncomputable when the CRS is invalid. So the implementation of setDeclination is modified so that it does not update grivation in those conditions. And similarly for setGrivation, setAuxiliaryScaleFactor, etc. This change has a relatively wide impact.

While some of the code changes for number 3 are for situations we might prefer not to bother with, a few are occasions for helpful thought about what the user experiences. For example, this helped me realize that the "Fix state switching" commit had altered the effect of editing the Declination when the state is -local-. It used to force grivation to equal declination, whereas with v0.9.3 the two values are coupled at a separation equal to the former convergence. I think the former behavior is more useful, since it's the _grivation_ which matters to a local CRS, and the user would like to be able to control it directly.

lpechacek:
> Number 4 is my take on @lpechacek 's suggestion to pass the update_grivation argument using an enumerator. It is enhanced so that the caller can specify not just updating grivation to match declination, but also the other way around: updating declination to match grivation. This turns out to apply in several locations.

Thanks, @pkturner, for listening to my hints. My idea was slightly different. I meant packing both the update control bits into one parameter and test. A draft implementation is at https://github.com/lpechacek/mapper/commit/d49a96811715d88c2b178d7a3d14f9762e181893. The outcome is not satisfactory for me either. My idea is definitely to make the action obvious from the method call itself. `setXXXRefPoint(point, false, false)` says nothing about what is going to happen and `setXXXRefPoint(point, Georeferencing::NoUpdates)` is only marginally better. This failed attempt leads me to further questions.

* Can the various method name + update flag combinations receive expressive names that would describe what is actually going to happen? Example: `georef.setXXXRefPoint(point, Georeferencing::UpdateScale)` would become `setXXXRefPointAndUpdateScale(point)` - yes, it's overly long name but says what going to happen.
* Following up on the previous, another question comes: Do the grid and scale updates have to happen in `setXXXRefPoint()`? The method has accumulated too much diverse functionality. Would a sequence `georef.setXXXRefPoint(point); georef.updateCombinedScaleFactor();` be a good alternative to `georef.setXXXRefPoint(point, Georeferencing::UpdateScale);`?

While the API cleanup is not an urgent issue, I think we should tackle the design principles as similar hindrances affect the whole codebase. At the same time, we don't have to worry about multiple rework cycles as we are the only user of Mapper internal interfaces and all the code is under our control.

pkturner:
@lpechacek 
Combining the two parameters into one that controls updating has merit. 

> Do the grid and scale updates have to happen in setXXXRefPoint()? The method has accumulated too much diverse functionality.

There are so many constraints on the properties of Georeferencing that it's typical that setting one property either affects others or leaves the Georeferencing object inconsistent. It also seems typical that when one property must change, there's a choice of how to make the other properties consistent. Good support for getting/keeping the Georeferencing consistent is essential.

The mention of what setXXXRefPoint should do prompted me to post a related problem that's been nagging at me. #1637

Issue #1608, opened 2020-05-11
Some area symbol objects trigger a crash


valdisj:
### Steps to reproduce
1. Opening OCAD12 file in Mapper

This happens with this OCAD file in attach. Other OCAD files are opening normal. But its worth to check, if something special is making MAPPER to totally crash, since this could be indication of wider problem.

### Actual behaviour
Mapper totally crashes. 

### Expected behaviour
OCAD file is opened.

### Configuration
Mapper Version: 0.9.2
Operating System: MacOS
[Bargupite_SIkompass-2019.ocd.zip](https://github.com/OpenOrienteering/mapper/files/4609190/Bargupite_SIkompass-2019.ocd.zip)

lpechacek:
Possible fix.

```
diff --git a/src/core/symbols/line_symbol.cpp b/src/core/symbols/line_symbol.cpp
index d0c8ed207a8a..e28f99198972 100644
--- a/src/core/symbols/line_symbol.cpp
+++ b/src/core/symbols/line_symbol.cpp
@@ -403,7 +403,7 @@ void LineSymbol::createSinglePathRenderables(const VirtualPath& path, bool path_
        if ((create_line || create_border) && processed_coords.size() > 1)
        {
                VirtualPath processed_path = { processed_flags, processed_coords };
-               processed_path.path_coords.update(path.first_index);
+               processed_path.path_coords.update(0);
                if (create_line)
                {
                        output.insertRenderable(new LineRenderable(this, processed_path, path_closed));
```

dg0yt:
@lpechacek Yes, this argument should always be zero, or rather `processed_path.first_index`. Broken since 7c26d51e, when I renamed the local variable `path` (shadowing a parameter with same name) to `processed_path`.

lpechacek:
Minimal reproducer. The key points are - combined symbol (area+dashed line), line dash longer than the hole in the area, and the hole must have at least one Bézier segment.

[crasher.omap.gz](https://github.com/OpenOrienteering/mapper/files/4611581/crasher.omap.gz)

PR #1609, opened 2020-05-11
LineSymbol: Fix crash in createSinglePathRenderables

dg0yt:
Amends 7c26d51.
Fixes GH-1608 (Crash with particular OCD file).


dg0yt:
I think I should create a new release soon, postponing a pile of template work to the next release.

lpechacek:
> I think I should create a new release soon, postponing a pile of template work to the next release.

I don't think this bug is severe enough to warrant an immediate update. The conditions for triggering it are very special and it is only one object in the example map that fulfills the condition. While there is no workaround, maybe the reporter can live with a development build until there is enough material for an update.

PR #1618, opened 2020-05-13
Set version to 0.9.3

dg0yt:


dg0yt:
Oh, I had to enable explicit PR validation for the `release` branch. Never noticed that when the source branch was `master`.

PR #1621, opened 2020-05-14
Georeferencing: Set reference point correctly when loading templates

### LINKED ISSUE #1620: In 0.9.2 addition of georeferenced Ocad template places it incorrectly.
### Steps to reproduce
1. Open Map.omap in 0.9.2 from attached archive. It shows georeferenced map with one TIF template in background.
2. Press CTRL+SHIFT+9 to open Template setup window
3. Add new template 4311-44-34_Contours.ocd (georeferenced Ocad file) from attached archive above TIF template.

### Actual behaviour
Loaded Ocad template does not match TIF template. There is some offset. Both are for the same tile, the same coordinate system. Version 0.9.1 does not have this behaviour and works correctly.

### Expected behaviour
Both templates should match.

### Configuration
Mapper Version: 0.9.2
Operating System: Microsoft Windows 10 Pro 10.0.18362
[OcadTemplateBug.zip](https://github.com/OpenOrienteering/mapper/files/4620472/OcadTemplateBug.zip)

pkturner:
When loading a template from an OCAD file, etc. and setting the projected reference point, hold on to the received combined scale factor rather than update it based on the auxiliary scale factor and grid scale factor.

This appears to fix #1620.

### DIFF
--- a/src/fileformats/ocd_georef_fields.cpp
+++ b/src/fileformats/ocd_georef_fields.cpp
@@ -852,7 +852,7 @@ void OcdGeorefFields::setupGeoref(Georeferencing& georef,
 
 	QPointF proj_ref_point(x, y);
 	georef.setProjectedRefPoint(proj_ref_point, false, false);
-	
+	georef.setCombinedScaleFactor(1.0);
 	georef.setGrivation(qIsFinite(a) ? a : 0);
 }
 
--- a/src/fileformats/ocd_georef_fields.cpp
+++ b/src/fileformats/ocd_georef_fields.cpp
@@ -851,7 +851,7 @@ void OcdGeorefFields::setupGeoref(Georeferencing& georef,
 		applyGridAndZone(georef, i, warning_handler);
 
 	QPointF proj_ref_point(x, y);
-	georef.setProjectedRefPoint(proj_ref_point, false);
+	georef.setProjectedRefPoint(proj_ref_point, false, false);
 	
 	georef.setGrivation(qIsFinite(a) ? a : 0);
 }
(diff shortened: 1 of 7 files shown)


dg0yt:
I see the pattern. I will try to test all changed spots.

dg0yt:
This change also fixes a bug which I observe with a local DXF file template (`OgrTemplate`) in one of my maps.

dg0yt:
I added another change which is a necessary complement of the original change.
(The spot I left out is `GeoreferencingDialog`, but there is a different pattern of `Georeferencing` usage.

So this works for me AFAICT now.

dg0yt:
Actually I want to redirect this PR to the the release branch first (to reach v0.9.3), and merge this branch or release to master afterwards.

lpechacek:
Well, I have very little insight into this part of code. Two remarks though.

First, this code change does get the contour lines from issue 1620 back into the right position. That is good. On the other hand, the boolean-only parameters are hard to read. In general, `setProjectedRefPoint(point, false, false)` is understood only by the author of the code and very determined outsiders who take the time to memorize the method signatures. I'd prefer flag setting approach here (e.g. `setProjectedRefPoint(point, Georeferencing::NoUpdates)`).

dg0yt:
> On the other hand, the boolean-only parameters are hard to read. In general, `setProjectedRefPoint(point, false, false)` is understood only by the author of the code and very determined outsiders who take the time to memorize the method signatures. I'd prefer flag setting approach here (e.g. `setProjectedRefPoint(point, Georeferencing::NoUpdates)`).

Thanks. The parameters are a valid point but cannot be done changed as a simple bug fix. And for the calls from `GeoreferencingDialog`, they may differ. We can record a `\todo` next time.

PR #1622, opened 2020-05-14
GeoreferencingTest: Use "WGS 84" as CRS ID

dg0yt:
Before PROJ 7.0.0, "WGS84" used to identify a geographic CRS (by
approximate match on "WGS 84"). Since 7.0.0, it identifies a datum
instead (by exact match).
Cf. https://github.com/OSGeo/PROJ/issues/2216.

### DIFF
--- a/test/georeferencing_t.cpp
+++ b/test/georeferencing_t.cpp
@@ -281,7 +281,7 @@ void GeoreferencingTest::testCRS_data()
 	        << QStringLiteral("+init=epsg:4258")
 	        << true;
 	QTest::newRow("WGS 84")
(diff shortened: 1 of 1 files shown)


dg0yt:
Asking for approval to put this into the v0.9.3 release.

lpechacek:
I doubt I have any formal authority to provide *approval*. :-)
I'll sanity-check the change tomorrow morning latest.

dg0yt:
In the first place, I want to know if you agree to include this (and the other change) into v0.9.3, now after the regular merge of master to release.

Issue #1302, opened 2019-07-31
OCD import - misplaced georeferenced OCD templates


lpechacek:
### Steps to reproduce
1. Open OCAD file with vector (previous maps in .ocd) templates.

### Actual behaviour
Templates are places at map origin instead of being aligned with other templates. Closing the templates and reopening them ensures correct placement.

### Expected behaviour
Templates should appear at the right position on the first attempt.

### Root cause
The .ocd templates are listed in background map (type 8) parameter strings without any rotation or positioning:
```
...
 8 	oldmap_ANONYMIZED_ocd12.ocd	s0	r1
...
```
However, Mapper does fill in some defaults for position, scale and rotation instead of reading these values from the georeferenced map. Another point is that this is happening without any warning about missing parameters.

### Configuration
Mapper Version: current `dev`
Operating System: openSUSE Tumbleweed

dg0yt:
How to proceed with this issue? I'm feeling unhappy with this bug being assigned, but without reproducible example or progress, for such a long time.

lpechacek:
Understood. I've got some hack in a topic branch which I can share in case you want to take the issue over. I can also share a reproducer.

dg0yt:
At least I would be willing to look into it. I'm already working on templates, OCD, and georeferencing.

lpechacek:
https://github.com/OpenOrienteering/mapper/commit/291c0ef67de416285e03d6ae0e361d37bc8770b5 is my shot at the problem. I started with tracking uninitialized data but I didn't like the the development direction and haven't returned to the topic. As far as I remember, the idea was to prevent filling in "sane defaults" into places which weren't specified in the file.

I've sent data for bug replication via e-mail.

dg0yt:
- We need to calculate the positioning in this case. That's what we already do when the user adds a new `MapTemplate` (not georefernced), but not when loading from OCD.
- We probably also want to save the OCD as before, i.e. with no explicit positioning instead of the calculated transformation. This should be easy if we remember that we have a georeferenced template.
  (Side note: Which one is the minimum version of OCD format which supports georeferenced templates?)

To get this resolved, we need to restore the support for `is_georefenced == true` in `TemplateMap` first. It is useful, and it shouldn't be too hard: I already do this in `GdalTemplate` which inherits from `TemplateMap`, and I know some cases which I shall test.

Will you find some time to look at GH-1645? I will probably add these commits on top of it.

lpechacek:
[...]

PR #1648, opened 2020-05-30
Fix positioning of georeferenced OCD templates

dg0yt:


dg0yt:
This is the POC with regard to GH-1302. Before reverting the draft state, we need
- add complementary OCD export capabilites
- extend TemplateMap's is_georeferenced support, including switching between georeferenced and non-georeferenced state.

dg0yt:
Maybe I start with a basic fix: Import georeferenced OCD templates as non-georeferenced, but with proper positioning. And add a \todo reminder to eventually add import in georeferenced state once `TemplateMap` is ready for this. The latter is within the scope of the other template georeferencing changes which pile up in local branches here (and sometimes uncover other template and georeferencing bugs).

dg0yt:
From https://github.com/OpenOrienteering/mapper/pull/1645#discussion_r433160889

OCD Version | OCD | GeoTIFF | World file
-------------|-----------|----------|-----------|
OCD v12 | :heavy_check_mark: |      ?     |      ?     |
OCD v11 | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
OCD v10 | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
OCD   v9 | :heavy_check_mark: |     ?      |      ?     |
OCD   v8 |      :x:      |     :x:       |      :x:      |

lpechacek:
Another point about OCAD implementation. v9 can load certain types of GeoTIFF but fails to display them afterward. I'm collecting these information bits here for a future reference as I expect complaints along the lines of "Mapper displays template XYZ just fine, but OCAD fails to show it."

lpechacek:
FWIW, I've just received a batch of .ocd files where templates get misplaced with Mapper `master` version. With this patchset, all templates land where they belong. From a practical point of view, this code change adds a lot of value.

dg0yt:
I suggest we just merge this as is now, even if it it only deals with import so far.

dg0yt:
Note that transforming map data on import is not exactly the same as transforming on drawing: The former applies an average scaling to symbols while the latter scales the rendered result in a more complex way. This may lead to small artifacts, but I assume it is neglectible in the context of orienteering maps.

lpechacek:
For some reason I get failed assertion in `Template::getTemplateExtent()`. I don't know why it didn't trigger before but now it takes the debug build down consistently. I stared into the code a bit but I was unable to figure out why is the test in place. I'll send you a test file for replication.

Quite naturally, the production build does not fail but the template does not render correctly in the initial view. Only after a window refresh, the template displays in full. I'd be glad if you found time to take a look. Thanks!

dg0yt:
Working on it, with good progress.

> I'll send you a test file for replication.

The test file has a quite interesting configuration: Both main map and OCD template have local georeferencing only. The current PR leads to the OCD template being handled as georefenced. But this is not what we are normally offering with locally georeferenced data (map or template), and it leads to follow-up issues (at least for changing the georeferenced state, not yet pushed). So I really need to handle this as a third case in `OcdFileImport` (besides explicit positioning and georeferencing).

dg0yt:
> For some reason I get failed assertion in `Template::getTemplateExtent()`. I don't know why it didn't trigger before but now it takes the debug build down consistently. I stared into the code a bit but I was unable to figure out why is the test in place. I'll send you a test file for replication.

I left the assertion in place. It just uncovered missing overrides in `TemplatePlaceholder`, added in https://github.com/OpenOrienteering/mapper/pull/1648/commits/a15a003378a7dd3f5041f853eb9b8890fe519437 now.

> Quite naturally, the production build does not fail but the template does not render correctly in the initial view. Only after a window refresh, the template displays in full. I'd be glad if you found time to take a look. Thanks!

The change to `TemplateMap::loadTemplateFileImpl()` was flawed, too. For the template extend (and calculation the dirty region), the transformation was applied twice.

Note that now the right transformation is determined, but the georeferenced state is not kept, for cases where the OCD files are not georeferenced.

https://github.com/OpenOrienteering/mapper/pull/1648/commits/36462a81f1fe9aad30cf00cfc9a2ad1eb6ade596

---

Following to the basic fix in this PR, the following is missing and/or in my pipeline:
- changing the georeferenced state of `MapTemplate` (similar to `OgrTemplate`, I will force-reload the data in order to get rid of the transformation).
- look into georeferenced state for OMAP templates (which we get as soon as we save to, and reload from, OMAP)

lpechacek:
Thanks, Kai for the fixes and explanation. The sample maps open fine with the latest code. I'll start patch review soon.

dg0yt:
> Note that transforming map data on import is not exactly the same as transforming on drawing: The former applies an average scaling to symbols while the latter scales the rendered result in a more complex way. This may lead to small artifacts, but I assume it is neglectible in the context of orienteering maps.

There is another difference which is not a "small artifact": Rotation is applied differently.
- Transforming on drawing means that data appears as if rotating a sheet of paper or a raster image.
- Transforming map data on import means that objects oriented (relative) to template north (point symbols, fill patterns, text) will have a new rotation relative the owning map's north. This is quite okay for geospatial data as in `OgrTemplate`, but it may not be what is desired for template orienteering maps / `MapTemplate`.

The current work-in-progress for switching the georeferenced state of `TemplateMap` is in https://github.com/OpenOrienteering/mapper/pull/new/template-map .

dg0yt:
Now looking at our export, it seems that the OCAD (Viewer) software ignores not only map scale, but *all explicit template positioning parameters* for OCD templates, and also the templates' georeferencing, e.g. UTM zone. It uses only the maps' Easting/Northing parameters. OCD templates must have
- the same scale
- the same grid (explicit or unset)
as the owning map.

dg0yt:
I added another commit to the template-map branch which prevents/detects saving of custom positioning of OCD templates in `OcdFileExport`: 51d9e027a7c204aaecb33ab28414ff4e562fe9cb
Open tasks:
- Error message for user, instead of `qDebug`
- Review this patch for changes which would better fit into earlier changes (starting with this PR).

dg0yt:
Replaced by GH-1697.

dg0yt:
> Replaced by [GH-1697](https://github.com/OpenOrienteering/mapper/pull/1697).

... which only deals with OCD templates, not graphics.

dg0yt:
This PR is now reduced to the originial changes which could handle absence of template positioning information for formats other than OCD. However, I'm not sure if things work like that.

Issue #1242, opened 2019-05-02
Enhancement: Fill areas in scribble mode


mlerjen:
### Situation
![hybrid](https://user-images.githubusercontent.com/15611653/57070657-684b1e80-6cd8-11e9-8226-65c84ebb4365.JPG)
The resolution of the scribble layer must be fine enough that you can sketch details with appropriate accuracy. On the other hand such fine grain makes it time consuming to fill the areas according their classification. In my sample above you can see: The grain is fine enough for the details but fills are represented in a sleazy manner.

### Proposed enhancement
Therefore I would like to propose the following. Enhance the sketch drawing by "auto-filling closed strokes"
![image](https://user-images.githubusercontent.com/15611653/57071277-320e9e80-6cda-11e9-88f3-ad541fa976dd.png)
Start.
![image](https://user-images.githubusercontent.com/15611653/57071292-42bf1480-6cda-11e9-8674-06f5d2a3989b.png)
Drawing a closed stroke/loop
![image](https://user-images.githubusercontent.com/15611653/57071372-90d41800-6cda-11e9-8fd7-e30b8f4c4c04.png)
Isolated stroke
![image](https://user-images.githubusercontent.com/15611653/57071316-58ccd500-6cda-11e9-8764-1effb764bb45.png)
Resulting filled area.

### Configuration
Mapper Version: 0.8.4.
Operating System: Android 5

ollesmaps:
I am not sure if I like it. The nature of scribbling might trigger accidental filling of areas ...

I was thinking something like "**spray**" with wider perimeter might do the job better. If I remember the MS Paint correctly the spray did not cover everything entirely. And that might be a good thing, you would be able to see stuff beneath, unless you spray longer.

jmacura:
Although I do not use the scribble layer myself, based on the description provided, I guess that the very basic "brush size" property might be a good progress in this.

lpechacek:
FWIW, the "pen width" option makes much more sense to me than flood fill.

mlerjen:
@lpechacek, @jmacura @ollesmaps:  
I would expect these basic benefits of my proposal:
1) You preserve the precision of the fine grained pen to define the borders of the so "flooded" area (surely an advantage to "clumsy brush" or "spray" :-)
2) You skip to excessively switch pen width. You can actually draw precisely and fill with the same tool.

lpechacek:
[...]

Issue #947, opened 2017-09-02
Possibility to hide color icons for scribbling and still be able to scribble


ollesmaps:
Please add possibility to hide color icons for scribbling if color is selecteted and is going to be used for a longer period. This action might do the same button as Paint-on-template. In fact the button hides color icons already now, however you can not scribble after that.

This will save space on small devices so you can draw on bigger screen.

Mapper Version: 0.7.0 (Android App)

dg0yt:
Side note: I never used this paint-on-template stuff - I use the normal drawing on the map. You could create your own map template with the desired colors and symbols, and then either use distinct field work maps (with the original map as a template), or add these symbols to the original map and work in a distinct map part during field-work. The recent changes on the freehand tool will make this work even better.

ollesmaps:
I draw all points and most of the lines in vector when fieldworking. I use scribbling for areas. 
Scribbling  is at the moment faster than vectorizing. The main reason is that you can have a selection of color icons available "on top" and you do not have to "fish" for it to a hidden screen as you have to for map symbols.

Issue #945, opened 2017-09-02
Color choice in Paint-on-template


ollesmaps:
The color selection on Paint-on-template could be bigger. 

Plus pink, orange, light green, violet, light blue, dark brown. It can be optional which colors and how many to add.

More colors mean more readable scribbles. For example now I have to draw all greens in one type of green and add letters inside them to give it meaning. That makes it less legible. In "mylar and pencil" mapping I used for example green for fight, orange for walk and pink for runnable thickets.

Mapper Version: 0.7.0 (Android App)

dg0yt:
How many colors will you need in total?

ollesmaps:
I would say 12 would be optimal. Maximum 14. 

I counted colors I used on mylar and that was 9. However if I applied some pressure I got different shades. So in total it was 12 colors.

ollesmaps:
I gave it some more thinking: The colors shall have different shades than map symbols (ISOM). For example now green has the same shade as ISOM green, so it is hard to distinguish between green on template and vector green (to distinguish between what has been vectorised and what not). 
Of course that relates to ISOM and if there is a non-ISOM map the clashes might be different. So a custum made palette would serve best.

lpechacek:
A test build is available from http://liborpechacek.wz.cz/oomapper/OpenOrienteering-Mapper-refs_pull_1738_merge_v20200925.1-Android-armeabi-v7a.apk

ollesmaps:
It works! I am excited :100: 

![palette](https://user-images.githubusercontent.com/20535899/100085153-fb642000-2e4b-11eb-94b8-b9a35d3409fc.jpg)

ollesmaps:
I experienced a peculiar crash with the above mentioned test build.

- All information in worked upon template _Draft @ 46...png_  got lost.
- This png file got corrupt.
- You could still see it in template list, however OOM could not open it.
- This oom file was saved short time before the crash - vectors, gpx were all fine, the rest of the templates were fine as well
-  This png was saved even before this mapping session, previous day.
- The trigger was perhaps a multitouch  - I was receiving a phone call and clumsily handled the tablet (but could be also something else).
- I later inspected this file and it was 0 kB, when opened in Notepad - no information inside

dg0yt:
> The trigger was perhaps a multitouch - I was receiving a phone call and clumsily handled the tablet (but could be also something else).

More likely: Android decided to terminate the app forcefully. It probably asked to quit before, but saving changed data took too long. 

[...]

PR #1652, opened 2020-06-07
PaintOnTemplateTool: Add filled area mode

lpechacek:
The `ScribbleOptions` part may feel like YAGNI but I have another idea for an enhancement. So at this point we can go with a boolean parameter to the method or leave the more descriptive and extensible `QFlags` in place. Please review, Kai.


src/templates/template.h
```diff
@@ -141,6 +141,7 @@ Q_OBJECT
 	 */
 	enum ScribbleOption
 	{
+		NoScribbleOptions = 0,
```
lpechacek, line 144: Yes, I had this line in the code but then I removed it. The "no options" is equivalent to newly initialized `QFlags` or just plain `{}`. That is why I removed the "no bits set" before submission.

Your proposal made me look at `QFlags` usage in Qt (`fgrep -B15 -r Q_DECLARE_FLAGS /usr/include/qt5/`). I can see that zero appears in the enums in about half of the cases. That said, I don't mind re-introduction of the value. This comment is a side note explaining why it was "missing".

src/templates/template.h
```diff
@@ -744,9 +745,10 @@ Q_OBJECT
 	Matrix template_to_map_other;
 };
 
-Q_DECLARE_OPERATORS_FOR_FLAGS(Template::ScribbleOptions)
-
 
 }  // namespace OpenOrienteering
 
+Q_DECLARE_OPERATORS_FOR_FLAGS(OpenOrienteering::Template::ScribbleOptions)
```
lpechacek, line 751: Ack this one.

lpechacek:
Thanks for the review, Kai.

Ad the pen widths, I believe that the filled areas mode provides a more capable alternative. As M. Lerjen [mentioned in the associated issue](https://github.com/OpenOrienteering/mapper/issues/1242#issuecomment-489554310), the fill areas option allows creating colored patches with precisely defined edges. Thicker pens do not provide this benefit.

My idea for enhancement was "overprint" mode. Recently I needed to overlay a color area over a few hand-sketched symbols. Combining these elements turned out to be quite tedious. I also think that the "overprint" scribbling mode could help people migrate from Mylar to a tablet.

Just my two cents, I'll be happy to read alternative opinions. :)

P.S. Plus a richer color palette would be required for migration from Mylar. Even I used about 9 pencils. Common number of pencils is 10-15 ([according to The Czech Mapper's Tutorial](http://tvorbamap.shocart.cz/vybaveni/soucasti.htm#tuzky)).

src/templates/template.h
dg0yt: Yes, this is is just a name for `ScribbleOptions{}`. However, a single-valued enum feels strange to mee, at least outside of C++ meta programming.

mlerjen:
image
Before. Tendency to incomplete area work.
image
With Fill areas in scribble mode.

The new filll area mode is a nice  and appreciated enhancement leading to much better field notes. Areas now can easily be represented by entirely filled areas instead of  their outlines  and a scribbled (time consuming) indication of the filling. A fine pen with and a efficient area filling are no longer contradicting.

image
So why not use the fill areas mode here?

In the logic of the workflow though, you would first draw point and line features making use of the template(s) before concluding a section with the specification of the area.  As the filling areas mode covers/erases all scribble it thus turns unhandy. In the example above the tool could only be applied to to the various sections of the central brown area each confined by either area outline or contours. (actually that was done for the upper brown area).

I would though consider to make the fill area mode "filling" instead of "covering" (replacing the only the transparent parts of a pixel with the filling color?? Except for white still voiding everything).

image
Point and short line features drawn directly. Contours as vectors sketched with a helper symbol and drawn with the freehand tool. Areas with out-lines on the scribble layer.

As a work-around to preserve the strength of the tool as it is now ("covering"). you can use the "draw freehand tool" with a helper line symbol to draw line features. These will not be covered by the concluding "fill area" scribble. This would btw. also be a way to approximate the full colorful potential of the Czech Map Painting tradition ;-)

dg0yt:
> I would though consider to make the fill area mode "filling" instead of "covering" (replacing the only the transparent parts of a pixel with the filling color?? Except for white still voiding everything).

Point taken. However, this would mean that you cannot re-color any area with the fill tool, unless erasing first. Or unless we find a way to handle at least two "layers".

lpechacek:
> > I would though consider to make the fill area mode "filling" instead of "covering" (replacing the only the transparent parts of a pixel with the filling color?? Except for white still voiding everything).
>
> Point taken. However, this would mean that you cannot re-color any area with the fill tool, unless erasing first. Or unless we find a way to handle at least two "layers".

The implementation is quite straightforward (https://github.com/lpechacek/mapper/commit/0aedd4be91f3b78c47da4044fb84deb9318de010) and I think the restriction @dg0yt mentions is far less hassle than trying to fit new drawing between existing lines.

I'd try producing a test build but my Pipelines instance is not yet configured. For a while I thought about creating a new branch in the official OO repo but then I realized that a (draft) pull request will also trigger Pipelines build with the possibility to download the signature-less test APK.

lpechacek:
A (self-)signed test APK is at http://liborpechacek.wz.cz/oomapper/OpenOrienteering-Mapper-refs_pull_1660_merge_v20200619.1-Android-armeabi-v7a.apk, however, I'm far from recommending installation of the package, given the previous experience with test builds.

The new tool behavior feels slightly counterintuitive to me. Making changes to existing drawing just does not "stick". The pen leaves a trace only in transparent areas.

dg0yt:
> The new tool behavior feels slightly counterintuitive to me. Making changes to existing drawing just does not "stick". The pen leaves a trace only in transparent areas.

I think this is what I meant by

> you cannot re-color any area with the fill tool, unless erasing first.

but extended to the pen!

Apart from the layers approach I mentioned, options are
- a pattern (like during drawing)
- multiplying blend mode for all drawing actions (like in overprinting simulation).

mlerjen:
Screenshot_20200619-180953_OpenOrienteering Mapper refspull1660merge v202006191

Great fun from my point of view. Last things to fine-tune maybe: 
- while the line pen has a certain opacity, the area pens are opaque and do blend with the drawn lines. (e.g. red and green result into a rather blackish line). Features scribbled with the line features thus tend to be "dissolved" by the areas.
- the button for the "fill area"... hm ... ?? :-)

Thanks again!

dg0yt:
> * while the line pen has a certain opacity, the area pens are opaque and do blend with the drawn lines.

This is the "multiplying blend mode" I mentioned.

lpechacek:
I've kept silent because I have no smart idea of how to solve the insufficiencies at the moment.

Let me at least draft the options:
1) The overprint/"multiplying blend" mode was my first idea. It is supposed to match pencil drawing but results in hue changes of the existing drawing. We haven't tried it so far, but I'm more and more persuaded that we should do give a shot to this one.

2) "Destination over" mode - current implementation where existing drawing projects through the new additions. There is a change in hue at the edges, where the drawing (alpha-) blends into the background. For lines, it means that they are less visible but the centerline still keeps its original hue. We could solve it by switching antialiasing off, but I'm not sure what would be the end-user response. The drawing would give an uglier appearance. Alternatively, thicker line would solve this as well.

Based on the increasing demand, I'll also look into the fill tool. :) My only concern is the increasing number of buttons in the scribble toolbar.

ollesmaps:
> My only concern is the increasing number of buttons in the scribble toolbar.  

What about a collapsible toolbar as suggested in issue #947. This isue was raised 3 years ago, when I was using small phone screen. Now I have a bigger one and it is not so pressing, however some might find it handy.

This also relates to color choice in scribble toolbar issue #945.

dg0yt:
> What about a collapsible toolbar as suggested in issue #947.

There is much non-standard code around the scribbling tools. I collected all related issues in a ["project"](https://github.com/OpenOrienteering/mapper/projects/7). (This PR is not added to the project, but the original issue is.) Now we still need someone to work on that.

lpechacek:
> There is much non-standard code around the scribbling tools. I collected all related issues in a "project". (This PR is not added to the project, but the original issue is.) Now we still need someone to work on that.

Ack. I'm aware of the project and read through it today again. I've got some idea about how to implement the requests, and I'd consider doing it for the pure joy of enhancing Mapper functionality. I'm, however, unsure what you mean with non-standard code and if you expect future contributions to move the code (style) in a certain direction.

dg0yt:
This is the result of painting with [`Qt::Dense4Pattern`](https://doc.qt.io/qt-5/qt.html#BrushStyle-enum), leaving some pixels untouched: 
dense-pattern

- It might make sense to turn off the outline (Qt: "pen") when drawing with a pattern, because repetitions will leave covering lines.
- On the same image, the pattern is always applied to the same pixels. Repeated painting with the same pattern will always change the same pixels (to the current color). No darkening, no mixing - only pixels created by pens or solid fills will shine through.
- 1 pixel wide pen strokes at 45° may get covered completely. (Cf. green line below red area.)

PR #1653, opened 2020-06-08
Major rework of Map's template API and of TemplateListWidget

dg0yt:
Move responsibilites from client code to `Map` where reasonable.
Factor our a proper `TemplateTableModel` from `TemplateListWidget`.

### DIFF
--- a/src/core/map.h
+++ b/src/core/map.h
@@ -631,8 +631,9 @@ public:
 	/**
 	 * Insert a new template at the given index.
 	 * 
-	 * To place a template immediately below the map, adjust first_front_template
-	 * manually with setFirstFrontTemplate()!
+	 * If the index is lower then the first front template index, the function
+	 * increments this index. To place a template immediately below the map,
+	 * pass index -1.
 	 */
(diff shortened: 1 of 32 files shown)


src/core/map.h
```diff
@@ -1469,7 +1471,7 @@ protected slots:
 private:
 	typedef std::vector<MapColor*> ColorVector;
 	typedef std::vector<Symbol*> SymbolVector;
-	typedef std::vector<Template*> TemplateVector;
+	typedef std::vector<std::unique_ptr<Template>> TemplateVector;
```
lpechacek, line 1506: The use of `unique_ptr` obsoletes `delete placeholder;` in `Importer::validate()` (file_import_export.cpp). In fact, the `delete` leads to a crash.

dg0yt: Okay. Will look into this tonight.

dg0yt: Release notes updated.
Side note: I meant to test the binaries once more before publishing the draft release, but accidenty hit the wrong button.

PR #1659, opened 2020-06-14
OgrTemplate: Double rotation while template is not visible

### LINKED ISSUE #1591: Rotation error with non-georeferenced DXF template
### Steps to reproduce
1. Map with open but invisible, non-georeferenced DXF template
2. Open map in Mapper.
3. Use the Georeferencing dialog to change declination.
4. Accept offer to rotate the map.
5. Make the template visible.

### Actual behaviour
The template appears, rotated twice as much as needed.

### Expected behaviour
The template should appear aligned with the map.

### Configuration
Mapper Version:0.9.2
Operating System:Linux

This is a case in which the template has or will have an `explicit_georef`. Perhaps it needs to be handled the same as a georeferenced template by `Map::rotateMap`.

pkturner:
Here is one way to fix #1591, along with an addition to the template_t test that exposes the problem if it is present.

In my current understanding, one of the two rotations applied to this template is more straightforward. Because the template is non-georeferenced, it has a TemplateTransform applied before it is displayed with the map. When the map is rotated, this TemplateTransform has the rotation added to it, to retain the template's relationship with the map objects.

The other rotation is more obscure (and this is the one which this fix removes). At the time the template data is loaded from its file, OgrTemplate provides a georeferencing for the imported data. In the problematic case, the chosen georeferencing is that of the main Map. Later, the template is hidden and the map is saved to a file.

After the map is reopened, the map is rotated, updating declination, before the template is loaded. Then when the template is loaded, Mapper again provides the georeferencing of the main Map, but this time its georeferencing is rotated from the original. It is not the georeferencing on which the template's TemplateTransform is based, and so it provides an additional, undesired rotation to the template data.

This fix adds a step to interpreting template XML which holds on to the map georeferencing which was present when the template was saved. When the template is loaded, this retained, unrotated georeferencing is used when OgrTemplate imports the data.

I have second thoughts about this change. Can't the problem arise any time the template is unloaded, not just from opening a map file containing an invisible template? I suspect the change is not in the right place.


dg0yt:
I'm working on this area of code, too. Some consolidation is needed with `TemplateMap` which doesn't really support `is_georeferenced` at the moment, but should do so (also needed for GH-1302). So I'm reluctant to deal with pull requests which touch these classes now. But I'm willing to cherry-pick automated tests for the current issues, even it means we need to mark some tests as expected fails for now.

pkturner:
Understood that it's best to hold off on this while you deal with `TemplateMap`. I found the name `is_georeferenced` a drag while working on `OgrTemplate`. In the relevant use cases, the issue revolves around the `Georeferencing` explicitly assigned to a DXF template, and all the while, the `is_georeferenced` property must be `false`. Perhaps you will be improving that aspect.

I'll be happy to wait and look at this again whenever appropriate.

dg0yt:
DXF and OSM are the formats which used to be hacked into `TemplateTrack`. So there are old maps which need "template track compatibility" for loading DXF/OSM data. OSM is close enough to GPX (using WGS 84), but for DXF, the legacy configuration is strange. I'm trying to document that. We don't have to care about this for new DXF/OSM templates, but even for GPX, there is the open issue for handling different realizations of WGS 84.

pkturner:
Pushed 3 more small commits to record my notions of code changes that are relevant to the issue, FWIW. Added a test for the effect of map rotation while the template is unloaded.

dg0yt:
Integrated into #1665. Many thanks!

PR #1660, opened 2020-06-19
PaintOnTemplateTool: Add drawing preservation mode

lpechacek:
Extends GH-1242 (Fill areas in scribble mode) - follow up to https://github.com/OpenOrienteering/mapper/pull/1652#issuecomment-645804058.

### DIFF
--- a/src/templates/paint_on_template_tool.cpp
+++ b/src/templates/paint_on_template_tool.cpp
@@ -32,6 +32,7 @@
 #include <QCursor>
 #include <QFlags>
 #include <QIcon>
+#include <QMenu>
 #include <QMouseEvent>
 #include <QPainter>
 #include <QPalette>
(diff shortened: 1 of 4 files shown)


dg0yt:
This would need to be (and benefit from being) reimplemented on top of GH-1675.

lpechacek:
Rebased and added the "overprint" mode for experimenting. I'd love to keep just one possibility for combining new drawing with the existing one but I'm undecided which of the "keep existing drawing unchanged"  and "overprint" is the right one. The switch between "keep" and "overprint" is hidden in the button menu.

The patch itself needs cleanup.

dg0yt:
For the record:
- As an alternative drawing mode, I wonder if we could fill the area twice: first solid with `CompositionMode_DestinationAtop` (filling all transparent areas), second by some brush pattern (partially covering existing colors).
- For the tool status, I would prefer to see the same locality for restoring (now scattered) as for saving (now single function).

lpechacek:
> * As an alternative drawing mode, I wonder if we could fill the area twice: first solid with `CompositionMode_DestinationAtop` (filling all transparent areas), second by some brush pattern (partially covering existing colors).
> * For the tool status, I would prefer to see the same locality for restoring (now scattered) as for saving (now single function).

Ack on both points. Regarding the combination with the existing drawing, I am more and more inclined to `DestinationAtop` without additional combinations. The only problem I register at the moment is the combination with the edge pixels which are partially transparent due to antialiasing (https://github.com/OpenOrienteering/mapper/pull/1652#issuecomment-646733547: _... the area pens are opaque and do blend with the drawn lines. (e.g. red and green result into a rather blackish line)._). It could be overcome by increasing the sketch template resolution and increasing the pen width so that the ratio between edge pixels and "core" pixels leans towards the core ones. The increase of resolution and pen width would also help in the pattern fill case (https://github.com/OpenOrienteering/mapper/pull/1652#issuecomment-656514808: _...painting with `Qt::Dense4Pattern`, leaving some pixels untouched:_).

mlerjen:
A new release including an android version with this feature would be very much appreciated.

dg0yt:
> A new release including an android version with this feature would be very much appreciated.

Sure. I try to finish another master release soon, including updates on third-party components which are meant to be tested before v0.9.4

dg0yt:
Picked up with significant modifications in GH-1706.

PR #1665, opened 2020-06-25
Template configuration state, and rotation fix

dg0yt:
This is a broader revision of the setup and configuration of templates. It includes an adaptation of the OgrTemplate rotation fix submitted by @pkturner in GH-1659.


dg0yt:
@pkturner I did following adaptations:
- Add basic testing of TemplateTrack and OgrTemplate, based on your test files and code.
- Consolidate the rotation test, have the unloaded case, the loaded cases, and the loaded-then-unloaded case for TemplateTrack and OgrTemplate all in one file (with the option to extent this to other templates classes.)
- The finish-configuration hook returns `bool`, as do most of our related virtual functions.

src/templates/template.h
```diff
@@ -646,6 +646,22 @@ Q_OBJECT
...
 	virtual bool loadTypeSpecificTemplateConfiguration(QXmlStreamReader& xml);
 	
+	/**
+	 * Hook which is called at the end of template configuration.
+	 * 
+	 * Derived classes may override this function if they need to do any post-
+	 * processing on the configuration. Unlike loadTypeSpecificTemplateConfiguration(),
+	 * this function is always called, even when there is no type-specific data.
+	 * 
+	 * The implementation must not do expensive calculations because is called
```
dg0yt, line 658: "it is called"

src/templates/template.h
```diff
@@ -646,6 +646,22 @@ Q_OBJECT
 	 */
 	virtual bool loadTypeSpecificTemplateConfiguration(QXmlStreamReader& xml);
 	
+	/**
+	 * Hook which is called at the end of template configuration.
```
dg0yt, line 652: "template configuration loading."

dg0yt:
> Add basic testing of TemplateTrack and OgrTemplate, based on your test files and code.

.. and in the CI build without GDAL, this immediately turns out a bug.

pkturner:
It looks good. I'm glad #1659 was useful.

PR #1666, opened 2020-06-26
XMLFileImporter: Adjust declination warning

dg0yt:
The warning shall not confuse users, but still give an indication
that manually checking the declination might be needed.

Task: GH-1235


dg0yt:
To be squashed on merge.

Issue #1673, opened 2020-07-10
Task: Refactor custom toolbars


dg0yt:
Qt provides a general toolbar class, `QToolBar`, which Mapper uses for most toolbars in the desktop UI.

Howevert, there is a number special-purpose toolbar implementations:
- [`ActionGridBar`](https://github.com/OpenOrienteering/mapper/blob/v0.9.3/src/gui/widgets/action_grid_bar.h#L46)
- [`KeyButtonBar`](https://github.com/OpenOrienteering/mapper/blob/v0.9.3/src/gui/widgets/key_button_bar.h#L53`)
- [`PaintOnTemplatePaletteWidget`](https://github.com/OpenOrienteering/mapper/blob/master/src/templates/paint_on_template_tool.h#L104)
- [`TextObjectAlignmentDockWidget`](https://github.com/OpenOrienteering/mapper/blob/v0.9.3/src/gui/widgets/text_alignment_widget.h#L46)

These individual classes seem to mix generic UI aspects with specific application aspects. They are not equally mature.

In particular, `ActionGridBar` is most mature and most generic. It is focused on UI, especially on usability in the touch UI. Like `QToolBar`, it is builds on `QToolButton` and `QAction`. It can handle multiple rows and an "overflowing" number of actions.

PR #1675, opened 2020-07-11
PaintOnTemplateTool: Use ActionGridBar for "palette"

dg0yt:
Replace the former proprietary `PaintOnTemplatePaletteWidget` by usage of the generic `ActionGridBar` (cf. GH-1673). This change facilitates development of scribbling features.

### DIFF
--- a/src/gui/widgets/mapper_proxystyle.cpp
+++ b/src/gui/widgets/mapper_proxystyle.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2013-2016, 2019 Kai Pastor
+ *    Copyright 2013-2020 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -32,6 +32,7 @@
 #include <QVariant>
 #include <QWidget>
 
+#include "settings.h"
 #include "gui/scaling_icon_engine.h"
 #include "gui/widgets/segmented_button_layout.h"
 
@@ -64,12 +65,19 @@ bool Q_DECL_UNUSED isDockWidgetRelated(const QWidget* widget)
 MapperProxyStyle::MapperProxyStyle(QStyle* base_style)
  : QProxyStyle(base_style)
 {
-	; // Nothing
+	auto& settings = Settings::getInstance();
+	onSettingsChanged(settings);
+	connect(&settings, &Settings::settingsChanged, this, [this]() {
+		onSettingsChanged(*qobject_cast<Settings*>(sender()));
+	});
 }
 
-MapperProxyStyle::~MapperProxyStyle()
+MapperProxyStyle::~MapperProxyStyle() = default;
+
+
+void MapperProxyStyle::onSettingsChanged(const Settings& settings)
 {
-	; // Nothing, not inlined
+	touch_mode = settings.touchModeEnabled();
 }
 
 void MapperProxyStyle::drawPrimitive(QStyle::PrimitiveElement element, const QStyleOption* option, QPainter* painter, const QWidget* widget) const
--- a/src/gui/widgets/mapper_proxystyle.cpp
+++ b/src/gui/widgets/mapper_proxystyle.cpp
@@ -96,6 +96,18 @@ void MapperProxyStyle::drawPrimitive(QStyle::PrimitiveElement element, const QSt
 			drawSegmentedButton(segment, overridden_element, option, painter, widget);
 			return;
 		}
+		if (touch_mode
+		    && element == PE_PanelButtonTool
+		    && option
+		    && option->state & State_On)
+		{
+			auto const& window_color = option->palette.window().color();
+			auto const fill = QBrush(qGray(window_color.rgb()) > 127 ? window_color.darker(125) : window_color.lighter(125));
+			painter->setPen(Qt::NoPen);
+			painter->setBrush(fill);
+			painter->drawRoundedRect(option->rect, 5.0, 5.0, Qt::AbsoluteSize);
+			return;
+		}
 		break;
 		
 #ifdef Q_OS_ANDROID
--- a/src/templates/paint_on_template_tool.cpp
+++ b/src/templates/paint_on_template_tool.cpp
@@ -77,7 +77,6 @@ void PaintOnTemplateTool::init()
 	
 	widget = new PaintOnTemplatePaletteWidget(false);
 	editor->showPopupWidget(widget, tr("Color selection"));
-	connect(editor->getMainWidget(), &QObject::destroyed, widget, [this]() { editor->deletePopupWidget(widget); });
 	connect(widget, &PaintOnTemplatePaletteWidget::colorSelected, this, &PaintOnTemplateTool::colorSelected);
 	connect(widget, &PaintOnTemplatePaletteWidget::undoSelected, this, &PaintOnTemplateTool::undoSelected);
 	connect(widget, &PaintOnTemplatePaletteWidget::redoSelected, this, &PaintOnTemplateTool::redoSelected);
--- a/src/templates/paint_on_template_tool.cpp
+++ b/src/templates/paint_on_template_tool.cpp
@@ -114,7 +114,7 @@ ActionGridBar* PaintOnTemplateTool::makeToolBar()
 			paint_color = color;
 			action->setChecked(true);
 		}
-		connect(action, &QAction::triggered, this, [this, color](bool checked) { if (checked) colorSelected(color); });
+		connect(action, &QAction::triggered, this, [this, color](bool checked) { if (checked) setColor(color); });
 		++count;
 	}
 	
@@ -164,7 +164,7 @@ void PaintOnTemplateTool::setFillAreas(bool enabled)
 }
 
 // slot
-void PaintOnTemplateTool::colorSelected(const QColor& color)
+void PaintOnTemplateTool::setColor(const QColor& color)
 {
 	paint_color = color;
 	
--- a/src/templates/paint_on_template_tool.cpp
+++ b/src/templates/paint_on_template_tool.cpp
@@ -23,33 +23,33 @@
 
 #include <Qt>
 #include <QtMath>
+#include <QAction>
+#include <QButtonGroup>
 #include <QBrush>
 #include <QCursor>
 #include <QFlags>
 #include <QIcon>
 #include <QMouseEvent>
 #include <QPainter>
-#include <QPaintEvent>
 #include <QPen>
 #include <QPixmap>
 #include <QPolygonF>
-#include <QRect>
 #include <QRgb>
 #include <QSettings>
+#include <QToolButton>
 #include <QVariant>
 
 #include "core/map.h"
 #include "core/map_view.h"
-#include "gui/util_gui.h"
 #include "gui/map/map_editor.h"
 #include "gui/map/map_widget.h"
+#include "gui/widgets/action_grid_bar.h"
 #include "templates/template.h"
 #include "util/util.h"
 
 
 namespace OpenOrienteering {
 
-// ### PaintOnTemplateTool ###
 
 int PaintOnTemplateTool::erase_width = 4;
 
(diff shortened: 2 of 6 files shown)


dg0yt:
This includes a visual change. These screenshots are from a desktop computer, in touch mode.

| Before: | After: |
|-----------|---------|
|before|after|

The button size in the "before" context was fixed, 13 mm on the PC, but it happens to be much smaller on Android [!]. The button size in the "after" context is the user-defined "Action button size", also set to 13 mm for these screenshots.

And obviously, I changed the distribution of the color buttons. We can easily restore the old order, but it is not so important if someone starts working on changing colors.

ollesmaps:
There is a vacant place for one extra colour :) I suggest pink.

dg0yt:
Pushed changes: Restored original sequence of color buttons.

dg0yt:
Pushed changes: Move `setAutoFillBackground` to `MapEditorController::showPopupWidget`.

dg0yt:
For this widget as much as for text alignment and `KeyButtonBar`, a valid alternative is to use a proper single-row `QToolBar`. I'm quite far with that for text alignment (scaling buttons to action button size, but not yet scaling icons to button size).

dg0yt:
> The marker frame was prominent enough to quickly identify which color will be used for scribbling. This change, however, makes the marker significantly more subtle

When switching to `QToolButton` (either in `ActionGridBar` or in `QToolBar`), we can try to adress this is in the drawing pipeline (via `MapperProxyStyle`). Perhaps it is also possible to handle this with `QIcon` states. (I need to add a custom icon engine in order to scale icons up like `ActionGridBar` does, but as a general feature.)

We could also add a single extra large button for the current color, similar to the symbol selector button.

lpechacek:
> Perhaps it is also possible to handle this with QIcon states.

The `QIcon` states approach is appealing to me. I gave a quick shot to it but the outcome was not as expected. The `QIconEngine` instance never sees `QIcon::Selected` state. Interestingly enough, it does receive `QIcon::Active` so that I could at least check that the frame gets rendered correctly. So far no idea why `QAction` does not propagate its state to its `QIcon`.

```
--- a/src/templates/paint_on_template_tool.cpp
+++ b/src/templates/paint_on_template_tool.cpp
@@ -29,6 +29,7 @@
 #include <QCursor>
 #include <QFlags>
 #include <QIcon>
+#include <QIconEngine>
 #include <QMouseEvent>
 #include <QPainter>
 #include <QPen>
@@ -81,6 +82,33 @@ void PaintOnTemplateTool::init()
        MapEditorTool::init();
 }
 
+class ColorButtonRenderer: public QIconEngine
+{
+public:
+       ColorButtonRenderer(QColor c) : button_color(c) {}
+       virtual void paint(QPainter *painter, const QRect &rect, QIcon::Mode mode, QIcon::State state) override;
+       virtual QIconEngine *clone() const override { return new ColorButtonRenderer(button_color); }
+private:
+       QColor button_color;
+};
+
+void ColorButtonRenderer::paint(QPainter *painter, const QRect &rect, QIcon::Mode mode, QIcon::State state)
+{
+       Q_UNUSED(state);
+       
+       painter->save();
+       painter->fillRect(rect, QBrush(button_color));
+       if (mode == QIcon::Selected)
+       {
+               auto pen = QPen(Qt::black);
+               pen.setWidth(3);
+               pen.setStyle(Qt::DotLine);
+               painter->setPen(pen);
+               painter->drawRect(rect.adjusted(2, 2, -2, -2));
+       }
+       painter->restore();
+}
+
 ActionGridBar* PaintOnTemplateTool::makeToolBar()
 {
        QSettings settings;
@@ -103,9 +131,7 @@ ActionGridBar* PaintOnTemplateTool::makeToolBar()
        };
        for (auto const& color: default_colors)
        {
-               QPixmap pixmap(1,1);
-               pixmap.fill(color);
-               auto* action = new QAction(pixmap, color.name(QColor::HexArgb), toolbar);
+               auto* action = new QAction(QIcon(new ColorButtonRenderer(color)), color.name(QColor::HexArgb), toolbar);
                action->setCheckable(true);
                toolbar->addAction(action, count % 2, count / 2);
                color_button_group->addButton(toolbar->getButtonForAction(action));
```

dg0yt:
> I gave a quick shot to it but the outcome was not as expected. The `QIconEngine` instance never sees `QIcon::Selected` state. Interestingly enough, it does receive `QIcon::Active` so that I could at least check that the frame gets rendered correctly. So far no idea why `QAction` does not propagate its state to its `QIcon`.

Actually you would have to look for `On`/`Off` `QIcon::State`.

I was playing with this as well but will need do more polishing before pushing anything. What I disliked about the icon approach is the fact that it doesn't use the tool / button panel and padding. I will try to polish the general style / tool button changes, and maybe we can combine this with icon changes.

dg0yt:
What you get now: flat, darker button background, and a checkmark fopr colors in `On` state.

palette

dg0yt:
Pushed changes to last commit:
- Move checkmark drawing to separate function.
- Create icon for physical pixel size, not for logical size. (Using logical size uncovered a shortcoming of the current action bar: It scales up the default pixmap of icons, but it fails to do the same for the `Off` state icon.)

src/gui/widgets/mapper_proxystyle.cpp
```diff
@@ -95,6 +95,16 @@ void MapperProxyStyle::drawPrimitive(QStyle::PrimitiveElement element, const QSt
 			drawSegmentedButton(segment, overridden_element, option, painter, widget);
 			return;
 		}
+		if (touch_mode && element == PE_PanelButtonTool)
+		{
+			auto const& palette = option->palette;
+			auto const& window = palette.window().color();
```
lpechacek, line 101: Suggest calling the variable `window_color` or `base_color` instead of plain `window`.

src/templates/paint_on_template_tool.cpp
```diff
@@ -50,6 +52,23 @@
 
 namespace OpenOrienteering {
 
+namespace {
+
+/// Draws a black or white checkmark, with good contrast to the background color. 
+void drawCheckmark(QPixmap& pixmap, const QColor& background)
+{
+	auto const icon_size = pixmap.width();
+	auto const luminance = 0.3 * background.redF() + 0.59 * background.greenF() + 0.11 * background.blueF();
```
lpechacek, line 61: Why does this method use another way of luminance computation than `MapperProxyStyle::drawPrimitive()`? FTR, `MapperProxyStyle::drawPrimitive()` uses Qt's `qGray()` function with weights [0.34375; 0.5; 0.15625].

src/templates/paint_on_template_tool.cpp
```diff
@@ -50,6 +52,23 @@
...
+	auto pen = QPen(QColor(luminance >= 0.5 ? Qt::black : Qt::white));
+	pen.setWidth(icon_size / 9);
+	QPainter p(&pixmap);
+	p.setPen(pen);
+	p.drawLine(6*icon_size/20, 11*icon_size/20, 8*icon_size/20, 13*icon_size/20);
+	p.drawLine(8*icon_size/20, 13*icon_size/20, 13*icon_size/20, 6*icon_size/20);
+};
```
lpechacek, line 68: The compiler reports an extra semicolon here.

lpechacek:
> Actually you would have to look for On/Off `QIcon::State`.

Ack. I don't know how I could misread the documentation that badly that I didn't understand the meaning of the widget state parameter.

src/templates/paint_on_template_tool.cpp
```diff
@@ -75,16 +75,71 @@ void PaintOnTemplateTool::init()
...
+	toolbar->addActionAtEnd(undo_action, 0, 0);
+	
+	auto* redo_action = new QAction(QIcon(QString::fromLatin1(":/images/redo.png")),
+	                                ::OpenOrienteering::MapEditorController::tr("Redo"),
+	                                toolbar);
+	connect(redo_action, &QAction::triggered, this, &PaintOnTemplateTool::redoSelected);
+	toolbar->addActionAtEnd(redo_action, 1, 0);
+	
+	auto* fill_action = new QAction(QIcon(QString::fromLatin1(":/images/scribble-fill-shapes.png")),
+	                                tr("Filled area"),
```
lpechacek, line 159: This `tr()` lacks full qualification. I think it was required for `lupdate` or so. Might be worth fixing.

src/gui/widgets/mapper_proxystyle.cpp
```diff
@@ -95,6 +95,16 @@ void MapperProxyStyle::drawPrimitive(QStyle::PrimitiveElement element, const QSt
...
+			auto const fill = QBrush(qGray(window.rgb()) > 127 ? window.darker(125) : window.lighter(125));
+			painter->setPen(Qt::NoPen);
+			painter->setBrush(fill);
+			painter->drawRoundedRect(option->rect, 5.0, 5.0, Qt::AbsoluteSize);
+			return;
+		}
```
lpechacek, line 107: For some reason, all tool modifier buttons appear as "active", i.e. dark grey, with this change. 
image

Maybe some other status information needs to be taken into account here.

lpechacek: Quick&dirty fix for the issue:
```
--- a/src/gui/widgets/mapper_proxystyle.cpp
+++ b/src/gui/widgets/mapper_proxystyle.cpp
@@ -99,7 +99,7 @@ void MapperProxyStyle::drawPrimitive(QStyle::PrimitiveElement element, const QSt
                {
                        auto const& palette = option->palette;
                        auto const& window = palette.window().color();
-                       auto const fill = QBrush(qGray(window.rgb()) > 127 ? window.darker(125) : window.lighter(125));
+                       auto const fill = QBrush(bool(qGray(window.rgb()) > 127) ^ !option->state.testFlag(QStyle::State_On) ? window.darker(125) : window.lighter(125));
                        painter->setPen(Qt::NoPen);
                        painter->setBrush(fill);
                        painter->drawRoundedRect(option->rect, 5.0, 5.0, Qt::AbsoluteSize);

dg0yt:
> Gotta retract my approval on the patch set. The modifier buttons require fixing.

That's okay. I noticed a number of issues, too, after finally deploying the right version to my mobile.

src/templates/paint_on_template_tool.cpp
dg0yt: `qGray()` is good enough. I just skipped it to quickly when it didn't gave the right color over green.

src/templates/paint_on_template_tool.cpp
dg0yt: I think it is okay. The full qualification is needed for other contexts than the the current class name IIRC. I will verify when this when updating translations.

src/gui/widgets/mapper_proxystyle.cpp
dg0yt: The dark panel is now used only for `On` state, not for `Sunken` or `Raised`.

dg0yt:
> typo in patch subject: `MapeEditorController: Force background on popups in touch mode`

Finally fixed this, too.

dg0yt:
I see ubsan killing the application sometimes at the end of testing. I do not fully understand how it relates to these changes - bisecting is made difficult by lack of reproducibility. It may even been an old issue.

dg0yt:
> I see ubsan killing the application sometimes at the end of testing. I do not fully understand how it relates to these changes - bisecting is made difficult by lack of reproducibility. It may even been an old issue.

This is unrelated to this PR. I can observe the ubsan issue already on master-v20200613.1.

mlerjen:
Wouldn't that be an opportunity to replace the white field with an eraser symbol, which it actually is, transparent, not white? Addtional the "fill only" option could be grayed out if eraser was selected as to combine  these two is meaningless...

lpechacek:
> Wouldn't that be an opportunity to replace the white field with an eraser symbol, which it actually is, transparent, not white?

Agree. This is also an opportunity for code cleanup as the program decides to choose to erase mode when [**white** color was selected](https://github.com/OpenOrienteering/mapper/blob/8c31fe5866e5dcb488a8145ca0134a5c2f71ac83/src/templates/paint_on_template_tool.cpp#L134). That is not nice.

BTW, did you know that pressing the right mouse button will do the erasing regardless of the selected color? :)

lpechacek:
Code proposal: https://github.com/lpechacek/mapper/commit/9c010d713884cfdc8d93d51c2d8811b163ab14dc

We are missing the eraser icon though.

dg0yt:
Offering a simple eraser icon. To get a checkmark when active.

eraser

dg0yt:
I merged the last two commits, and I restored the right mouse button behaviour (which is mentioned in the desktop statusbar).

lpechacek:
We forgot about one thing. With the introduction of fill areas mode and eraser mode, more status information needs to be stored. I.e. When switching the tools, the selected color remains in effect but the fill areas and eraser buttons get de-selected.

A draft fix is at https://github.com/lpechacek/mapper/commit/2cb1f30f0c23deb0ea3116fb9082a56422b0bfb3. I'll let the patch bake a bit more before placing a submit request.

dg0yt:
I see what you mean.

I think these settings don't have to be neither persistent (over restarts) nor global (for all files). I think we should simply store this in generic Map object properties (`QObject::setProperty`). This would also help with desktop mode when multiple map files can have different lifecycles of this tool.

This will need more thought once you want to add more customization (colors, pen width, etc.)

PR #1681, opened 2020-07-17
Deprecated methods cleanup

lpechacek:
Address deprecation warnings reported by GCC 10 compiler with Qt 5.15 on openSUSE Tumbleweed.

### DIFF
--- a/src/gui/autosave_dialog.h
+++ b/src/gui/autosave_dialog.h
@@ -57,7 +57,7 @@ public:
 	 * @param actual_path    The path which is currently selected.
 	 * @param parent         The parent window.
 	 */
-	AutosaveDialog(const QString& original_path, const QString& autosave_path, const QString& actual_path, MainWindow* parent = nullptr, Qt::WindowFlags f = 0);
+	AutosaveDialog(const QString& original_path, const QString& autosave_path, const QString& actual_path, MainWindow* parent = nullptr, Qt::WindowFlags f = {});
 	
 	/**
 	 * Destructor.
--- a/3rd-party/cove/app/colorseditform.cpp
+++ b/3rd-party/cove/app/colorseditform.cpp
@@ -78,7 +78,9 @@ class ColorsEditingDelegate : public QItemDelegate
 			QVariant v = model->data(model->index(index.row(), 0, index),
 									 Qt::BackgroundRole);
 			QColor c = v.value<QBrush>().color();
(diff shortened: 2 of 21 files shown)


lpechacek:
Messed up the MapCoord init. I will rework it.

lpechacek:
> Messed up the MapCoord init. I will rework it.

Seems to be a limitation in older versions of Clang compiler and not a code bug per se. Anyway, these compilers are still used for Android and MacOS builds, so the data members are still (zero) initialized in the default constructor.

@dg0yt, are you fine with merging these cleanup patches into `dev`, please?

src/gui/autosave_dialog.h
```diff
@@ -57,7 +57,7 @@ Q_OBJECT
 	 * @param actual_path    The path which is currently selected.
 	 * @param parent         The parent window.
 	 */
-	AutosaveDialog(const QString& original_path, const QString& autosave_path, const QString& actual_path, MainWindow* parent = nullptr, Qt::WindowFlags f = 0);
+	AutosaveDialog(const QString& original_path, const QString& autosave_path, const QString& actual_path, MainWindow* parent = nullptr, Qt::WindowFlags f = Qt::WindowFlags());
```
dg0yt, line 60: I'm fine with the proposed form. But usually I avoid spelling the type twice, by using `... = {}`

lpechacek: Thanks for the review. I'll adjust the line and merge the changes.

lpechacek: Changed all the occurrences to value initialization syntax and updated the commit message to reflect the fact that the code follows Mapper convention instead of the Qt style.

dg0yt:
How about applying these commits to master, and resetting dev to the state before these commits? I cannot merge master into dev in the current state, and some commits are needed to fix build errors on newer linux distributions.

lpechacek:
AFAICT, the merge conflict is easy to resolve. The affected code in `src/templates/paint_on_template_feature.cpp` was reworked and changes made in `dev` are now obsolete. That is, it is safe to take the code lines from `master` and discard the changes in `dev`. It can be achieved by manual editing during the merge or `git merge -s recursive -X theirs origin/master` command.

Anyway, I will rebase the patch set to `master` and submit there for the sake of simplicity.

lpechacek:
PR 1725 is in place.

PR #1692, opened 2020-08-04
Replace TextObjectAlignmentDockWidget by QToolBar

dg0yt:
This PR provides the changes
- to use QToolBar with the user-defined action button size in touch mode
- to replace the custom TextObjectAlignmentDockWidget by QToolBar

Thus it also indicates the feasibility of using a single row standard QToolBar for scribbling colors.


dg0yt:
Note the toolbar separator between horizontal und vertical alignment actions:

sample

PR #1706, opened 2020-08-24
Paint-on-template changes

dg0yt:
Task: GH-1242

### DIFF
--- a/src/templates/paint_on_template_tool.cpp
+++ b/src/templates/paint_on_template_tool.cpp
@@ -38,6 +38,7 @@
 #include <QPen>
 #include <QPixmap>
 #include <QPoint>
+#include <QPointF>
 #include <QPolygonF>
 #include <QRgb>
 #include <QSettings>
@@ -120,6 +121,34 @@ QIcon makeEraserIcon(int const icon_size)
 	return icon;
 }
 
+QIcon makeBackgroundDrawingIcon(int const icon_size)
+{
+	auto const background = QApplication::palette().color(QPalette::Window);
+	QRectF circle = {
+	    0.1*icon_size,  0.1*icon_size,
+	    0.8*icon_size,  0.8*icon_size
+	};
+	QPointF triangle[] = {
+	    { 0,  0 },
+	    { 0.8*icon_size, 0 },
+	    { 0, 1.4*icon_size },
+	};
+	
+	QPixmap pixmap(icon_size, icon_size);
+	pixmap.fill(Qt::transparent);
+	
+	QPainter p(&pixmap);
+	p.setPen(Qt::darkGreen);
+	p.setBrush(Qt::green);
+	p.drawEllipse(circle);
+	p.setPen(Qt::NoPen);
+	p.setBrush(QColor(isDark(background) ? Qt::lightGray : Qt::darkGray));
+	p.drawPolygon(triangle, std::extent<decltype(triangle)>::value);
+	p.end();
+	
+	return QIcon(pixmap);
+}
+
 }
 
 
@@ -194,6 +223,7 @@ ActionGridBar* PaintOnTemplateTool::makeToolBar()
 		}
 		connect(action, &QAction::triggered, this, [this, color]() {
 			erasing.setFlag(ExplicitErasing, false);
+			background_drawing_action->setEnabled(true);
 			setColor(color);
 		});
 		++count;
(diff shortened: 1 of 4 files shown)


dg0yt:
scribbling-1 
Change button arangement, new button for Gray.

---

scribbling-2 
Multiple blend modes for filling, via press-and-hold on fill mode button.

---

scribbling-3 
Template selection moved to scribble mode button (press-and-hold). Only templates in the current viewport, relative position indicated as icon.

lpechacek:
Thanks, Kai, for reworking my draft. I've got remarks from the field. In the previous days I've spent about 90 hours in the forest, extensively using the scribble mode.

My experience is that the multiply mode (blending) is not that useful. The background mode (`DestinationOver`) is enough. The use case is:  I draw details first (marshes, undergrowth) and then want to cast an area over it. I use the background composition mode. At this point it is good that colors are retained, the details are better visible. Second use case, I want to add a detail to existing drawing (e.g. extend a marsh), then I use the foreground mode. So from my vantage point, these two modes are enough.

It is, however, essential for me, that these two modes can be switched with a single quick tap for the sake of efficiency. So, with KISS principle in mind, I propose having a dedicated button for switching the mode. Thanks!

mlerjen:
Just to make sure... the area (foreground) mode also be available for the eraser.

dg0yt:
> Just to make sure... the area (foreground) mode also be available for the eraser.

Actually the erase is strictly foreground.

dg0yt:
> My experience is that the multiply mode (blending) is not that useful.
> ...
> So from my vantage point, these two modes are enough.

Understood. I would have waited for feedback from the master release, but we can skip this.

> ... I propose having a dedicated button for switching the mode. Thanks!

I see.
We need to find a better icon than in the first attempt, but this is doable.
Do you want to extend the background mode to lines and to eraser as well?

dg0yt:
scribbling

Background mode by distinct button, with new icon. 
Multiply mode removed. 
Pattern fill still available as alternative to solid fill, now orthogonal to foreground/background mode, and in a separate commit.

Still needs to be complemented by style changes to increase template list item size for Android / touch mode (work in progress).

mlerjen:
Does a change regarding "area" and "level" change the setting for all pens plus the eraser or just for the selected pen/eraser?

What happens e.g. if I am working with "blue-area-background" -> switch to eraser  -> switch to blue pen
I would hope for:
blue area background
eraser **area** foreground
blue area **background**

What happens e.g. if I am working with "blue area background" -> switch to red -> line -> foreground -> switch to blue pen.
I would hope for:
blue area background
red line foreground
blue area background

Do you see what I mean? Preserving the settings per pen would allow deviating use of the different pens, while settings per pen-set is only serving uniform use or resulting in a click overkill. (if eg. using red for foreground lines and yellow for background area it would take three clicks each to switch to and back). 

Let's just test it.

dg0yt:
> Does a change regarding "area" and "level" change the setting for all pens plus the eraser or just for the selected pen/eraser? ...

At the moment is the former: a global setting. 

I would love to see the color buttons being transformed to customizable pens (color, line width, area mode), but I do not want to spend too much time on this myself now. That's why I tried to spend some time on making the code more standard, more open to modification.

lpechacek:
> Do you want to extend the background mode to lines and to eraser as well?

Thanks, for incorporating my input! Background mode for lines, yes. For eraser, no.

> I would love to see the color buttons being transformed to customizable pens (color, line width, area mode)…

This idea occurred to me as well. Having buttons for "yellow area filling brush", "sharp green pencil" and the like, would take the tool to a new level. Let's take iterative approach to development, however.

I'll try to test this build soon.

dg0yt:
> Background mode for lines, yes.

Okay, adding this now.

dg0yt:
I refactored the main commit into a number of individual changes. The WIP style change is needed mostly to make menus usable on high-dpi touch devices (Android).

dg0yt:
Moved the style and template list changes to separate PRs.

lpechacek:
Brief feedback on the new UI.

1) Works fine, modulo a few nits. :)
2) The eraser tool uses thicker pen width. In fill area mode, the preview does not honor the higher width. I.e. the erased area is slightly bigger than the white preview indicates.
3) Clicking the eraser tool button for the second time re-selects the previous drawing color. I'd expect that repeated clicks on the button will keep the tool selected.
4) I failed to familiarize myself with the new draft selection dialog. When I long-click the scribble tool button with stylus, the menu appears right below the stylus. As I cannot read it, I instinctively lift the stylus from the display which, however, **immediately selects a template**.
5) I don't understand the icons showing overlap between the draft and viewport. Looks like the more white there is, the more overlap. I'd expect the opposite. The more colored icon, the more overlap.

dg0yt:
Thanks. I already very much split it into three PRs, and made more adjustments.

> Brief feedback on the new UI.
> 
>     1. Works fine, modulo a few nits. :)
> 
>     2. The eraser tool uses thicker pen width. In fill area mode, the preview does not honor the higher width. I.e. the erased area is slightly bigger than the white preview indicates.

I will check this again.

>     3. Clicking the eraser tool button for the second time re-selects the previous drawing color. I'd expect that repeated clicks on the button will keep the tool selected.

I already dropped that commit.

>     4. I failed to familiarize myself with the new draft selection dialog. When I long-click the scribble tool button with stylus, the menu appears right below the stylus. As I cannot read it, I instinctively lift the stylus from the display which, however, **immediately selects a template**.

Maybe it is better with the last style changes (already merged)? I was much more unhappy about the single-item menu and the two more clicks to select a template.

>     5. I don't understand the icons showing overlap between the draft and viewport. Looks like the more white there is, the more overlap. I'd expect the opposite. The more colored icon, the more overlap.

Feel free to experiment with color and pattern. I didn't want to overload the icon, and contrast for outdoor use is another relevant aspect.
This is now separate PR, and the icons are separate commit there.

dg0yt:
> 2\. The eraser tool uses thicker pen width. In fill area mode, the preview does not honor the higher width. I.e. the erased area is slightly bigger than the white preview indicates.

Done.

mlerjen:
A while ago we changed the brown from a lighter to a darker shade... which made it from a "background-color" to a "foreground-color". It is e.g. no longer combinable with red or purple as the colors are to much alike now.
Pondering on this feature (release please😆) I wondered if it would make sense to always apply a lighter  shade of the respective color if applied in area mode.  That way the area color never would dominate the fore-ground drawing.... just a thought.

dg0yt:
Assuming this is approved, at least for making a master release now.

lpechacek:
Ack. Functionality-wise I approve the patch set. I only briefly looked at the code but I doubt there are any serious problems in that area. I'm fine with the merge and sorry for the delays.

PR #1721, opened 2020-09-06
Update translations

dg0yt:


sikmir:
Translations still locked on weblate...

dg0yt:
> Translations still locked on weblate...

Thanks for reporting! Now unlocked again.

PR #1725, opened 2020-09-11
Deprecated methods cleanup (for master)

lpechacek:
Patch set from PR 1681 rebased to current master.

### DIFF
--- a/3rd-party/cove/app/colorseditform.cpp
+++ b/3rd-party/cove/app/colorseditform.cpp
@@ -78,7 +78,9 @@ class ColorsEditingDelegate : public QItemDelegate
 			QVariant v = model->data(model->index(index.row(), 0, index),
 									 Qt::BackgroundRole);
 			QColor c = v.value<QBrush>().color();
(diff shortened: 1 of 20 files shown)


lpechacek:
@dg0yt if you have no objections, I'll merge the patch set tomorrow evening.

dg0yt:
Thanks, Libor.

`QFontMetrics::horizontalAdvance()` is missing in some versions of Qt that we still support. I didn't want to add `#ifdef`, so I suggest switching to `QFontMetrics::boundingRect()` instead.

lpechacek:
The change makes sense. As the switch to `boundingRect()` effectively replaces the original patch, I'm squashing the two changes into one. In my opinion, it makes no sense to drag the intermediate code state into permanent history. Thanks!

dg0yt:
Saved current `dev` to https://github.com/OpenOrienteering/mapper/tree/dev-20200911, for a few days.
Resetting `dev` to https://github.com/OpenOrienteering/mapper/tree/9a52840598ba9f57baa231d9916d5ec1fbd56729 now.

PR #1738, opened 2020-09-25
PaintOnTemplateTool: Add color setup page to Settings

lpechacek:
Let's give a shot to this implementation of the scribble color settings dialog. It may feel rough but at the same time it might be "just right". Comments are welcome especially in the following areas:

* Are the scribble colors device-dependent or map-dependent? I.e. should the setting be stored on the device on in the map file?
* Is the user interface usable on mobile devices?

### DIFF
--- a/src/gui/util_gui.cpp
+++ b/src/gui/util_gui.cpp
@@ -50,6 +50,8 @@
 #include <QTextDocument>
 #include <QVariant>
 #include <QWidget>
+#include <QToolButton>
+#include <QIcon>
 
 #include "mapper_config.h"
 #include "settings.h"
@@ -378,7 +380,26 @@ namespace Util {
 		return maybe_markup;
 	}
 
-	
+
+	namespace ToolButton
+	{
+		/**
+		 * Returns a new QToolButton with a unified appearance.
+		 */
+		QToolButton* create(const QIcon& icon, const QString& text, const char* whats_this)
+		{
+			auto* button = new QToolButton();
+			button->setToolButtonStyle(Qt::ToolButtonIconOnly);
+			button->setToolTip(text);
+			button->setIcon(icon);
+			button->setText(text);
+			if (whats_this)
+				button->setWhatsThis(OpenOrienteering::Util::makeWhatThis(whats_this));
+			return button;
+		}
+	}  // namespace ToolButton
+
+
 }  // namespace Util
 
 
--- a/src/gui/widgets/color_list_widget.cpp
+++ b/src/gui/widgets/color_list_widget.cpp
@@ -64,14 +64,23 @@
 
 namespace OpenOrienteering {
 
+namespace {
+
+/// Local wrapper for Util::ToolButton::create() adding the What's This bit.
+QToolButton* createToolButton(const QIcon& icon, const QString& text)
+{
+	return Util::ToolButton::create(icon, text, "color_dock_widget.html");
+}
+
+}  // anonymous namespace
+
+
 ColorListWidget::ColorListWidget(Map* map, MainWindow* window, QWidget* parent)
 : QWidget(parent)
 , map(map)
 , window(window)
 {
 	react_to_changes = true;
-	
-	setWhatsThis(Util::makeWhatThis("color_dock_widget.html"));
 	
 	// Color table
 	color_table = new QTableWidget(map->getNumColors(), 7);
@@ -91,18 +100,18 @@ ColorListWidget::ColorListWidget(Map* map, MainWindow* window, QWidget* parent)
 	duplicate_action->setIcon(QIcon(QString::fromLatin1(":/images/tool-duplicate.png")));
 	
 	// Buttons
-	auto new_button = newToolButton(QIcon(QString::fromLatin1(":/images/plus.png")), tr("New"));
+	auto new_button = createToolButton(QIcon(QString::fromLatin1(":/images/plus.png")), tr("New"));
 	new_button->setPopupMode(QToolButton::DelayedPopup); // or MenuButtonPopup
 	new_button->setMenu(new_button_menu);
-	delete_button = newToolButton(QIcon(QString::fromLatin1(":/images/minus.png")), tr("Delete"));
+	delete_button = createToolButton(QIcon(QString::fromLatin1(":/images/minus.png")), tr("Delete"));
 	
 	auto add_remove_layout = new SegmentedButtonLayout();
 	add_remove_layout->addWidget(new_button);
 	add_remove_layout->addWidget(delete_button);
 	
-	move_up_button = newToolButton(QIcon(QString::fromLatin1(":/images/arrow-up.png")), tr("Move Up"));
+	move_up_button = createToolButton(QIcon(QString::fromLatin1(":/images/arrow-up.png")), tr("Move Up"));
 	move_up_button->setAutoRepeat(true);
-	move_down_button = newToolButton(QIcon(QString::fromLatin1(":/images/arrow-down.png")), tr("Move Down"));
+	move_down_button = createToolButton(QIcon(QString::fromLatin1(":/images/arrow-down.png")), tr("Move Down"));
 	move_down_button->setAutoRepeat(true);
 	
 	auto up_down_layout = new SegmentedButtonLayout();
@@ -110,10 +119,10 @@ ColorListWidget::ColorListWidget(Map* map, MainWindow* window, QWidget* parent)
 	up_down_layout->addWidget(move_down_button);
 	
 	// TODO: In Mapper >= 0.6, switch to ColorWidget (or generic) translation context.
-	edit_button = newToolButton(QIcon(QString::fromLatin1(":/images/settings.png")), QApplication::translate("OpenOrienteering::MapEditorController", "&Edit").remove(QLatin1Char('&')));
+	edit_button = createToolButton(QIcon(QString::fromLatin1(":/images/settings.png")), QApplication::translate("OpenOrienteering::MapEditorController", "&Edit").remove(QLatin1Char('&')));
 	edit_button->setToolButtonStyle(Qt::ToolButtonTextBesideIcon);
 	
-	auto help_button = newToolButton(QIcon(QString::fromLatin1(":/images/help.png")), tr("Help"));
+	auto help_button = createToolButton(QIcon(QString::fromLatin1(":/images/help.png")), tr("Help"));
 	help_button->setAutoRaise(true);
 	
 	// The buttons row layout
@@ -195,18 +204,6 @@ void ColorListWidget::showEvent(QShowEvent* event)
 	}
 }
 
-
-
-QToolButton* ColorListWidget::newToolButton(const QIcon& icon, const QString& text)
-{
-	auto button = new QToolButton();
-	button->setToolButtonStyle(Qt::ToolButtonIconOnly);
-	button->setToolTip(text);
-	button->setIcon(icon);
-	button->setText(text);
-	button->setWhatsThis(whatsThis());
-	return button;
-}
 
 void ColorListWidget::newColor()
 {
--- a/src/gui/widgets/color_wheel_widget.cpp
+++ b/src/gui/widgets/color_wheel_widget.cpp
@@ -1,0 +1,436 @@
+/*
+ *    Copyright (C) 2013-2020 Mattia Basaglia
+ *    Copyright 2021 Libor Pecháček
+ *
+ *    This file is part of OpenOrienteering.
+ *
+ *    OpenOrienteering is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    OpenOrienteering is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with OpenOrienteering.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "color_wheel_widget.h"
+
+#include <QAbstractButton>
+#include <QBrush>
+#include <QConicalGradient>
+#include <QFlags>
+#include <QGradientStops>
+#include <QHBoxLayout>
+#include <QLineEdit>
+#include <QLineF>
+#include <QMouseEvent>
+#include <QPainter>
+#include <QPainterPath>
+#include <QPalette>
+#include <QPen>
+#include <QPoint>
+#include <QPolygonF>
+#include <QPushButton>
+#include <QRect>
+#include <QRectF>
+#include <QRgb>
+#include <QSizeF>
+#include <QSizePolicy>
+#include <QString>
+#include <QVBoxLayout>
+#include <QVector>
+#include <QtGlobal>
+#include <QtMath>
+
+class QPaintEvent;
+class QResizeEvent;
+
+
+namespace OpenOrienteering {
+
+
+ColorWheel::ColorWheel(QWidget* parent)
+    : QWidget(parent)
+{
+	background_is_dark = palette().window().color().valueF() < 0.5;
+}
+
+
+ColorWheel::~ColorWheel() = default;
+
+
+// Get the user selected color
+QColor ColorWheel::color() const
+{
+	return QColor::fromHsvF(hue, sat, val);
+}
+
+
+// Set the widget selection color
+void ColorWheel::setColor(QColor c)
+{
+	auto oldh = hue;
+	// render red triangle for achromatic colors
+	hue = c.hsvHueF() >= 0 ? c.hsvHueF() : 0;
+	sat = c.hsvSaturationF();
+	val = c.valueF();
+
+	if (!qFuzzyCompare(oldh + 1, hue + 1))
+		renderValSatTriangle();
+	
+	update();
+	emit colorChanged(c);
+}
+
+
+void ColorWheel::paintEvent(QPaintEvent* event)
+{
+	Q_UNUSED(event)
+
+	QPainter painter(this);
+	painter.setRenderHint(QPainter::Antialiasing);
+	painter.translate(geometry().width() / 2, geometry().height() / 2);
+
+	// hue wheel
+	if (hue_ring.isNull())
+		renderHueRing();
+
+	painter.drawPixmap(-outerRadius(), -outerRadius(), hue_ring);
+
+	// hue selector
+	drawRingEditor(hue, painter, Qt::black);
+
+	// lum-sat triangle
+	if (inner_selector.isNull())
+		renderValSatTriangle();
+
+	qreal side = triangleSide();
+	qreal height = triangleHeight();
+
+	painter.rotate(selectorImageAngle());
+	painter.translate(selectorImageOffset());
+
+	qreal slice_h = side * val;
+	qreal ymin = side / 2 - slice_h / 2;
+
+	// clip with the triangular shape
+	auto clip = QPainterPath();
+	clip.addPolygon(QPolygonF(QVector<QPointF> { {0, side / 2},
+	                                             {height, 0},
+	                                             {height, side} }));
+	painter.setClipPath(clip);
+
+	painter.drawImage(QRectF(QPointF(0, 0), QPointF(height, side)),
+	                  inner_selector);
+	painter.setClipping(false);
+
+	// lum-sat selector
+	// we define the color of the selecto based on the background color of the
+	// widget in order to improve to contrast
+
+	auto const selector_position = QPointF(val * height, ymin + sat * slice_h);
+	auto const selector_ring_color = (background_is_dark && (val < 0.65 || sat > 0.43))
+	                           || val <= 0.5 ? Qt::white : Qt::black;
+	painter.setPen(QPen(selector_ring_color, 3));
+	painter.setBrush(Qt::NoBrush);
+	painter.drawEllipse(selector_position, selector_radius, selector_radius);
+}
+
+
+void ColorWheel::mouseMoveEvent(QMouseEvent* event)
+{
+	if (mouse_status == DragCircle)
+	{
+		hue = lineToPoint(event->pos()).angle() / 360.0;
+		renderValSatTriangle();
+	}
+	else if (mouse_status == DragInside)
+	{
+		QLineF glob_mouse_ln = lineToPoint(event->pos());
+		QLineF center_mouse_ln(QPointF(0, 0),
+		                       glob_mouse_ln.p2() - glob_mouse_ln.p1());
+
+		center_mouse_ln.setAngle(center_mouse_ln.angle() + selectorImageAngle());
+		center_mouse_ln.setP2(center_mouse_ln.p2() - selectorImageOffset());
+
+		QPointF pt = center_mouse_ln.p2();
+
+		qreal side = triangleSide();
+		val = qBound(0.0, pt.x() / triangleHeight(), 1.0);
+		qreal slice_h = side * val;
+
+		qreal ycenter = side / 2;
+		qreal ymin = ycenter - slice_h / 2;
+
+		if (slice_h > 0) sat = qBound(0.0, (pt.y() - ymin) / slice_h, 1.0);
+	}
+	else
+	{
+		return;
+	}
+
+	emit colorSelected(color());
+	emit colorChanged(color());
+	update();
+}
+
+void ColorWheel::mousePressEvent(QMouseEvent* event)
+{
+	if (event->buttons() & Qt::LeftButton)
+	{
+		setFocus();
+		QLineF ray = lineToPoint(event->pos());
+		if (ray.length() <= innerRadius())
+			mouse_status = DragInside;
+		else if (ray.length() <= outerRadius())
+			mouse_status = DragCircle;
+
+		// Update the color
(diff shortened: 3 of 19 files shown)


jmacura:
My thoughts on this matter:
Scribble colours should be device-dependent, as it is rather an interest of a particular mapper, to have his/her favourite colours ready for scribbling. I don't see any close link between the actual map, as a product, and a colour, which is being used at some point in time for its creation. (Think of an analogy: back in the time of pencils, a mapper had his/her colour palette of pencils in possesion, which he/she usually did not handed to another mapper with the tracing-paper map sheets 😉 ).

Other mapper folks might have different insight, of course.

In any case, a brief information about where these colour settings are stored might be included in the UI.

Disclaimer: I have not tried to run the testing build with this PR, just

src/gui/widgets/paint_on_template_settings_page.cpp
```diff
@@ -0,0 +1,246 @@
...
+namespace {
+
+struct ColorPresetEntry {
+	const char* name;
+	std::initializer_list<QColor> colors;
+};
+
+constexpr std::array<ColorPresetEntry, 2> color_presets = {
+    //: Paint on template color preset entry name.
+    ColorPresetEntry { QT_TR_NOOP("Traditional"),
```
dg0yt, line 63: In which "context" is this translation meant to appear for translators?

src/gui/widgets/paint_on_template_settings_page.cpp
```diff
@@ -0,0 +1,246 @@
...
+                       { QColor(255,   0,   0),
+                         QColor(255, 255,   0),
+                         QColor(  0, 255,   0),
+                         QColor(219,   0, 216),
+                         QColor(  0,   0, 255),
+                         QColor(209,  92,   0),
+                         QColor(  0,   0,   0) }
+    },
+    //: Paint on template color preset entry name.
+    ColorPresetEntry { QT_TR_NOOP("Revised 2020"),
```
dg0yt, line 73: (Translation context again)

src/gui/widgets/paint_on_template_settings_page.cpp
```diff
@@ -0,0 +1,246 @@
...
+{
+	auto button_num = 0;
+	for (const auto* preset_button : preset_buttons)
+	{
+		if (preset_button->isChecked())
+			break;
+		button_num++;
+	}
+	
+	if(button_num < preset_buttons.size())
```
dg0yt, line 200: `if (` (space!)

src/gui/widgets/paint_on_template_settings_page.cpp
```diff
@@ -0,0 +1,246 @@
...
+	{
+		initializePage(color_presets[button_num].colors);
+	}
+	else if(button_num == preset_buttons.size() && !custom_string_edit->text().isEmpty())
```
dg0yt, line 204: `if (` (space!)

src/settings.cpp
```diff
@@ -393,5 +399,38 @@ void Settings::setNmeaSerialPort(const QString& name)
 	}
 }
 
+std::vector<QColor> Settings::colorsStringToVector(QString config_string)
+{
+	auto const color_strings = config_string.split(QLatin1Char(','));
```
dg0yt, line 404: `...splitRef(...`

src/settings.cpp
```diff
@@ -393,5 +399,38 @@ void Settings::setNmeaSerialPort(const QString& name)
...
+}
+
+std::vector<QColor> Settings::paintOnTemplateColors() const
+{
+	return colorsStringToVector(getSetting(PaintOnTemplateTool_Colors).toString());
+}
+
+QString Settings::colorsVectorToString(const std::vector<QColor>& new_colors)
+{
+	auto strings = QStringList {};
```
dg0yt, line 422: `strings.reserve(...);`

src/templates/paint_on_template_tool.cpp
```diff
@@ -195,16 +195,8 @@ ActionGridBar* PaintOnTemplateTool::makeToolBar()
...
-	    qRgb(255,   0,   0),
-	    qRgb(255, 255,   0),
-	    qRgb(  0, 255,   0),
-	    qRgb(219,   0, 216),
-	    qRgb(  0,   0, 255),
-	    qRgb(209,  92,   0),
-	    qRgb(  0,   0,   0),
-	};
+	auto const default_colors = Settings::getInstance().paintOnTemplateColors();
+	int count = !(default_colors.size() % 2); // align colors with the other icons
```
dg0yt, line 199: Looks strange to me with type `int` and operator `!`. Do you mean:
~~~
int count = (default_colors.size() + 1) % 2;
~~~

dg0yt (commented):
Just saw the recent CI run and coudn't resist leaving some comments.

src/gui/widgets/paint_on_template_settings_page.cpp
lpechacek: Thanks for making me look into the docs again. The original aim was to provide the right context but I missed the `context` param at `QT_TRANSLATE_NOOP`. Fixed.

src/gui/widgets/paint_on_template_settings_page.cpp
lpechacek: Fixed.

src/settings.cpp
lpechacek: Added.

src/templates/paint_on_template_tool.cpp
lpechacek: The combination of `int` and boolean NOT is a common C idiom. But we are typing C++, so I've adopted your expression. :)

src/settings.cpp
lpechacek: Fixed. I keep forgetting about this one.

lpechacek:
Part of the comments didn't make it to GitHub, so I've posted them now. Thanks, Kai for the review!

lpechacek:
There are still problems on Android though. The configuration dialog does not use the screen width in full and the presets cannot be successfully activated.  Preset activation results in black entries with varying opacity. I'll continue debugging that part.
image
Also, the default color selection dialog is overly complex for the mobile device but that is a minor problem in my perception. The "custom string" field is intended as an aid for palette migration between devices and for sharing (i.e. either select all the text in the field, copy and paste into an e-mail or do the reverse). If you have proposals for enhancements or coding tips, feel free to share. Thanks!

lpechacek:
Further tests on Pixel 2 phone with Android 11 show that in portrait screen mode the color selection dialog overflows the screen and the OK button is beyond the right screen edge.  After switching to landscape mode the color list ignores the double-tap and never spawns the color selection dialog. Sigh...

The copy-paste on Android is magic on its own as it involves invoking the new text editing tool from the on-screen keyboard, but there is a way.

dg0yt:
For a linear arragement, it might be enough to start with add/remove/up/down buttons.

When offering drag and drop, I would prefer to see the same two-row arrangement as used by the tool. (Unfortunately we don't have drag'n'drop in `ActionGridBar`. However, this could be useful for future customizations.

For (loading) the presets, I would have choosen a combo box or menu button, possibly opening another input dialog for the custom string.

lpechacek:
> For a linear arragement, it might be enough to start with add/remove/up/down buttons.
> 
> When offering drag and drop, I would prefer to see the same two-row arrangement as used by the tool. (Unfortunately we don't have drag'n'drop in `ActionGridBar`. However, this could be useful for future customizations.

Yeah, `ActionGridBar`-like dialog with drag-and-drop was my initial idea but I wanted to make it with a minimum amount of code. That is how I arrived at the `QListWidget` as it offered the D'n'D at a low coding cost. But as the implementation is far from being perfect, it makes send to give up on D'n'D at this point and design the dialog in the Template List manner.

> For (loading) the presets, I would have choosen a combo box or menu button, possibly opening another input dialog for the custom string.

I see. On an afterthought, I doubt anyone will edit the palette config string by hand. So, two buttons like "Copy palette config to clipboard" and "Paste palette config", sitting next to "Apply preset" might be a better solution. Thanks for the inspiration.

lpechacek:
Intermediate status update in place of a code update.

1) Color selection dialog. The default Qt one is too heavyweight for mobile screens. I'm looking into the following options - borrow some small code from elsewhere/code own dialog OR copy and adapt Qt `QColorDialog` code OR rebuild Qt with `QT_SMALL_COLORDIALOG` defined. I'm currently testing the last option to see if it helps.
2) Broken presets on Android. There must be something with Qt on Android and I'm not going to root-cause the issue. Instead, I'll define the presets as character strings. It will be easier to define new and tweak existing presets after all.
3) The color config dialog. I'll rewrite it in the "Template list" manner. The benefit is that the current usage instructions will become obsolete as I'm reusing an existing UI design pattern.

I'm currently working on 1).

lpechacek:
> ... rebuild Qt with `QT_SMALL_COLORDIALOG` defined. I'm currently testing the last option to see if it helps.

The outcome is unsatisfactory from my POV. The dialog in the middle is the reduced `QColorDialog`. Six +/- controls for RGB and HSV. I wouldn't like to expose our users to anything like that. GTK color triangle is my current preference for the custom dialog.

image

mlerjen:
These colors are used for just notes. :-) 
The triangle for a rough and the HTML for an exact color definition  would be fair enough IMO.

dg0yt:
> borrow some small code from elsewhere/code own dialog

How about implementing something like this in Qt/C++:
https://github.com/Pes8/android-material-color-picker-dialog

lpechacek:
Yes, I'm stubborn. :) This is what I imagined:
image

Thanks for the suggestions. The Android color picker offers a per-component RGB selection. Honestly, I like the HSV model more as it matches my decision making. First I choose the color, then I decide about its "strength". I've extracted the color wheel implementation from @mattbas's [Color Widgets](https://gitlab.com/mattbas/Qt-Color-Widgets.git). The code requires a lot of cleanups though.

lpechacek:
FTR, I've finished the UI rework and I'm now happy with it. The changes include the button-driven edition instead of the drag-and-drop, and a table-like colors view where the hex-number color notiation stands separate from the color preview field. There is a lot of code cleanup required though and I continue work on that front. BTW, the right table column should not contain the hex-numbers. I'll remove them as part of the coming changes.
image

lpechacek:
I'm done with the changes for now. The UI functionality I imagined is there, clang-tidy and IWYU do not complain about the new code, and the code itself looks reasonable to me. Certainly, there are many nits to eradicate, which I accept.

Ad the `newToolButton()` de-duplication, I'm wondering if the `using` declaration is the right approach. While I'm not against it, it feels exotic in Mapper code. The alternative is a full symbol qualification.

The color wheel thingie is fairly isolated. I reworked the code mattbas' code into a more compact form with a bit of modernization. Given the isolated nature of the code bit, I didn't invest much time into it. Let me know if you want to see more effort investment there.

On an afterthought, I made the color settings page dependent on the touch mode. I.e. the dialog section is hidden in the desktop mode. The implementation is fairly straightforward and I strived for the elimination of duplicate code. One of the things I'd like to go is the `#ifdef Q_OS_ANDROID` but have no idea how to parameterize this part. Of course, the whole creation of the tool button bar (`list_buttons_layout`) is another candidate for de-duplication as the structure appears in about four other dialogs.

Last, but not least, we need the actual alternative color presets. In that regard, I rely on contributions from  @mlerjen and @ollesmaps. :smiley:

Ready for review now.

src/gui/util_gui.cpp
```diff
@@ -378,7 +380,26 @@ namespace Util {
...
 	}
 
-	
+
+	namespace Tools
+	{
+		/**
+		 * Returns a new QToolButton with a unified appearance.
+		 */
+		QToolButton* newToolButton(const QIcon& icon, const QString& text, const char* whats_this)
```
dg0yt, line 389: From the context in this file, I think the expected code pattern is more like:

~~~
namespace ToolButton
{
    QToolButton* create(const QIcon& icon, const QString& text, const char* whats_this)
    {
...
~~~

But I admit that this would require more changes add the calling spots, changing them to:
~~~
    auto* button = Util::ToolButton::create(...)
~~~
unless wrapping this call into small inline definitions of `newToolButton(()`.

src/gui/widgets/color_list_widget.cpp
```diff
@@ -62,6 +62,17 @@
...
 
+namespace {
+
+/// Local wrapper for newToolButton adding the What's This bit.
+QToolButton* newToolButton(const QIcon& icon, const QString& text)
+{
+	return OpenOrienteering::Util::Tools::newToolButton(icon, text, "color_dock_widget.html");
+}
+
+}  // anonymous namespace
```
dg0yt, line 73: How about moving the anonymous namespace into namespace OpenOrienteering?

src/gui/widgets/tag_select_widget.cpp
```diff
@@ -48,6 +48,17 @@
...
+
+/// Local wrapper for newToolButton adding the What's This bit.
+QToolButton* newToolButton(const QIcon& icon, const QString& text)
+{
+	return OpenOrienteering::Util::Tools::newToolButton(icon, text, "find_objects.html#query-editor");
+}
+
+}  // anonymous namespace
+
+
```
dg0yt, line 61: How about moving the anonymous namespace into namespace OpenOrienteering?

src/gui/widgets/template_list_widget.cpp
```diff
@@ -121,21 +121,13 @@ QVariant makeCheckBoxDecorator(QStyle* style, const QSize& size)
...
-	button->setToolTip(text);
-	button->setIcon(icon);
-	button->setText(text);
-	button->setWhatsThis(OpenOrienteering::Util::makeWhatThis("templates.html#setup"));
-	return button;
+	return OpenOrienteering::Util::Tools::newToolButton(icon, text, "templates.html#setup");
 }
 
-}
+}  // anonymous namespace
```
dg0yt, line 130: How about .. Well, I see, the namespace was already there :smile:

src/gui/widgets/color_wheel_widget.h
```diff
@@ -0,0 +1,173 @@
...
+ */
+class ColorWheelDialog : public QDialog
+{
+	Q_OBJECT
+public:
+	ColorWheelDialog(QWidget* parent = nullptr,
+	                 Qt::WindowFlags f = Qt::WindowFlags());
+	QColor currentColor() const;
+	void setColor(const QColor& c);
+	static QColor getColor(const QColor& initial, QWidget* parent = nullptr);
```
dg0yt, line 163: This API looks inconsistent: static `getColor()` vs. non-static `setColor()` plus `currentColor()`.

src/gui/widgets/color_wheel_widget.cpp
```diff
@@ -0,0 +1,435 @@
...
+}
+
+
+QColor ColorWheelDialog::getColor(const QColor& initial, QWidget* parent)
+{
+	ColorWheelDialog dlg(parent);
+	dlg.setColor(initial);
+	auto const sizex = QGuiApplication::primaryScreen()->availableSize();
+	auto const dlg_size = qMin(sizex.width(), sizex.height());
+	dlg.resize(QSize(dlg_size, dlg_size));
```
dg0yt, line 429: Is this for maximizing the dialog?
This is what I use in `TextBrowserDialog`:
https://github.com/OpenOrienteering/mapper/blob/d3fc644a8b6c4f5958704a315a7b252293ba071a/src/gui/text_browser_dialog.cpp#L87-L88

dg0yt (commented):
I did only a quick review on GH. 

The CI issue should be resolved. However, I don't know how to make the PR build again the HEAD of master. Pushing updates does help :smile:

src/gui/util_gui.cpp
lpechacek: The idea behind the `Tools` name that I will also move the duplicated toolbar setup code into this namespace one day. Then `Tools::newToolButton()` and `Tools::newToolBar()` would be the symbols in the namespace. But we are not yet there, so my name choice is a typical YAGNI instance.

Your proposal sounds much better to me, even at the cost of changing the call sites. So, if `Util::ToolButton::create()` is your final proposal, I'll implement it.

src/gui/widgets/template_list_widget.cpp
lpechacek: One thing I hate about these reviews is that unheard-before rules appear in the course of the discussion. I acknowledge that I'm no more than an average coder and I know there are always things to improve. However, no other maintainer ever dug that deep into my code, with the exception of Junio H., who was right on every point like you often are. Adhering to the Coding Conventions and the tool suggestions does not give a guarantee that we will not spend time on technicalities over and over again. My wish is to learn this project's coding style without the need for the eternal patch rewrites.

Ad the placement of anonymous namespace, you spotted yourself that it's inconsistent throughout the existing code. Most of the time anonymous namespaces are enclosed in `OpenOrienteering` and a few times it's outside. As far as unqualified symbol lookup is concerned, the exact placement plays almost no role. The symbol scope is restricted to the translation unit anyway, so we are talking about the availability of the enclosed symbols outside the `OpenOrienteering` curly braces and potential shadowing of symbols defined outside of the OO namespace. The **consistency point is relevant** in my view but otherwise, the discussion feels very academic to me as I see only marginal benefits from either case. If I missed some crucial moment, I'll be happy to learn where I made the mistake.

src/gui/widgets/color_wheel_widget.cpp
lpechacek: Yes, it is. Your code is much better and I'll adopt it. Thanks for the suggestion!

src/gui/widgets/color_wheel_widget.h
lpechacek: Correct. The API is supposed to mimic `QColorDialog`. The names should be `currentColor()`, `setCurrentColor()` and static `getColor()`. Will fix.

lpechacek:
> I did only a quick review on GH.

Thanks! Will implement the suggestions.

> ... However, I don't know how to make the PR build again the HEAD of master. Pushing updates does help smile

I'm not sure how to understand this sentence. Shall I rebase the patchset on top of the current `master`, rebase on a past commit, or shall I leave the branch point as is? I know you had reservations against patchset rebase in the past which I didn't fully understand. In my view, merging a patchset that has a minimal delta to the target branch makes the code history easier to navigate and mitigates merge conflicts.

src/gui/widgets/template_list_widget.cpp
dg0yt: I intentionally did mark the PR as "comments", not "changes requested".
I'm sorry but I lack experience of receiving reviews, apart from contribution to some open source projects. It is a pity that I seem to have failed again to carry what is important, what is a sidenote, or what I learned myself from a review. Unless there is another tool then these comments, I must leave more thoughts unsaid.

With the namespace, the whole point is whether one writes / a future developer reads:

~~~
  return OpenOrienteering::Util::Tools::newToolButton(icon, text, "color_dock_widget.html");
~~~

vs.

~~~
  return Util::Tools::newToolButton(icon, text, "color_dock_widget.html");
~~~

dg0yt:
> > ... However, I don't know how to make the PR build again the HEAD of master. Pushing updates does help smile
>
> I'm not sure how to understand this sentence. 

CI builds a merge of the PR's branch with the target branch (master here), identified as refs/pull/1738/merge.
But once a particular HEAD of the PR is pushed to GH, this reference stays at the state of master at the time of the push. When the HEAD of master moves, e.g. solving a CI problem, this change does not "upgrade" the builds of the PR. Newly scheduled builds will use the same commit (identified as refs/pull/1738/merge) and fail again.

However, when HEAD of the PR (lpechacek:issue-945-scribble-color-settings) changes in some way, refs/pull/1738/merge is updated, too, and picks up the new HEAD of master.

(Instead of updating the HEAD of the PR, I sometimes changed the target branch (twice). But this leaves irritating noise in the PR conversation.)

> Shall I rebase the patchset on top of the current `master`, rebase on a past commit, or shall I leave the branch point as is? 

It is not necessary to rebase. Choose what fits your work.

> I know you had reservations against patchset rebase in the past which I didn't fully understand. In my view, merging a patchset that has a minimal delta to the target branch makes the code history easier to navigate and mitigates merge conflicts.

In some cases it was hard to see/understand past PR conversations on GH after rebasing. And also the sequence of commits displayed in GH is sometimes surprising to me (author date vs. commit date). Apart from that, rebasing is fine. Maybe I am more used to it, after all.

src/gui/util_gui.cpp
dg0yt: I'm not planning to invent more proposals, so the existing pattern is a "final proposal". But it is really just a proposal for consideration. If you don't want to adopt it now, you might also a add `\todo` comment. Or just to decide to leave your code as is.

src/gui/widgets/template_list_widget.cpp
lpechacek: Yes, the path to `OpenOrienteering` symbols is another valid point for moving the namespace. Thanks!

Regarding the content of review feedback, I'd be happy with comments prefixed "side note", "optional change", "for your consideration" or similar. Chances are, I'll adopt most of your proposals because they are useful. It is the lack of clarity *when* the code change is suitable for merge that often drains my enthusiasm.

lpechacek:
I hope that I'll get to the branch refresh tomorrow evening or Monday latest. Thanks!

lpechacek:
Changes in this iteration:
* Renamed `newToolButton()` to `ToolButton::create()`. The local helper wrappers are now `createToolButton()`. Changed the call sites accordingly. Not yet perfect but better than the previous implementation.
* Moved anonymous namespace into `OpenOrienteering` in changed files.
* Fixed `ColorWheel` so that it displays a red triangle instead of a grey triangle for achromatic colors.
* Fixed `ColorWheelDialog` API and made the dialog show maximized on screen.
* Fixed `PaintOnTemplateSettingsPage` color table to allow only single-cell selection and addressed the dangling FIXME.

As I mentioned earlier, I'd like to get feedback in a form that makes it clear which comments point out structural defects or other serious problems, and which ones propose cosmetic changes. I also don't mind receiving feedback in code patch form. I understand that it might be easier to propose a concrete change than to try to steer me towards the desired goal, and I won't feel offended by such an action.

dg0yt:
> As I mentioned earlier, I'd like to get feedback in a form that makes it clear which comments point out structural defects or other serious problems, and which ones propose cosmetic changes. I also don't mind receiving feedback in code patch form. I understand that it might be easier to propose a concrete change than to try to steer me towards the desired goal, and I won't feel offended by such an action.

Fair enough.
However, adding patches and extended comments normally requires me to check-out and build the code. And I wonder if some people wouldn't prefer short peer-to-peer comments over another person's mine-is-better-than-yours patches on top of their hard work.

At the moment, I'm short on time.  I really tried to make that obvious, by writing that "I did only a quick review on GH." For a more thorough review you might need to wait a week or month, but I didn't want to leave that PR without encouraging feedback.

dg0yt:
> Changes in this iteration:

I'm not sure when I will be able to look at the code. I propose you set a fair deadline for review, and otherwise proceed and merge the code.

lpechacek:
> At the moment, I'm short on time. I really tried to make that obvious, by writing that "I did only a quick review on GH." For a more thorough review you might need to wait a week or month, but I didn't want to leave that PR without encouraging feedback.

Thanks, appreciated! I can wait. It took me four and half months to get the PR ready, after all.

Regarding timing, I'd love to get the functionality into v0.9.5. We've got three active users (me included) already, which in my view proves the relevance of this enhancement.

lpechacek:
A test APK is available from http://liborpechacek.wz.cz/oomapper/OpenOrienteering-Mapper-refs_pull_1738_merge_v20210207.1-Android-armeabi-v7a.apk .

dg0yt:
> Regarding timing, I'd love to get the functionality into v0.9.5.

We can release another master snapshot or even v0.9.5. But how do we want to deal with the warnings reported for this PR?

lpechacek:
>  But how do we want to deal with the warnings reported for this PR?

I completely forgot about examining the build results from Azure. I have found the warnings and will address them ASAP. I'll also add the Copy&Paste buttons for the custom settings field as discussed above. Thanks for the reminder!

lpechacek:
Ad `codespell`, after reading the associated commit message it is not clear to me why all the ignored words are listed on individual lines instead of being part of a comma-separated list (`-L copyable,endwhile,intrest`). Similarly, it is unclear why the word "childs" is listed in `codespell.yml` while the rest is in `codespell.sh`. I will not take the word blacklisting path this time and I'll rename the offending method instead. For the future, I'd like to be able to infer from the code structure and associated commentary how to extend the code in a compatible manner.

lpechacek:
AFAICT, the build is clean now.

Changes in this iteration:
* Renamed `ColorWheel::renderRing` and `ColorWheel::renderTriangle`
* Fixed `#include`s
* Added Copy and Paste buttons for the custom palette string field

dg0yt (approved):

lpechacek:
Thanks, Kai, for the review. I'm merging the patch set.

dg0yt:
Great. Concerning release planing, I will spend a short cycle on Vikazimut support similar to MapRunF. 
Then we can either release another master snapshot which should be pretty stable this time.
Or we just release v0.9.5, moving my GDAL/OGR work to v0.9.6.
(When my GDAL/OGR work is released, I will jump to v1.0, bumping requirements on dependencies.)

lpechacek:
@dg0yt, looking at known issues in v0.9.4 and the shortlog, a quick v0.9.5 feels appropriate to me. Then the highlights for v0.9.6 would be your GDAL/OGR support and the stylus-only input. I also plan to make progress on the symbol delete undo, which might be part of that release too.

Regarding v1.0 and the updated requirements, is the plan to maintain a stable v0.9 branch for some time or will the requirements fit all the target platforms?

PR #1740, opened 2020-09-29
Merge translation and symbol set updates

dg0yt:


dg0yt:
@OpenOrienteering/developers Meant to be the last merge before releasing v0.9.4.

Issue #1709, opened 2020-08-27
GPX tracks display in different position on desktop vs. Android (regression)


sfroyen:
### Steps to reproduce
1. Open a map on Android OOM and attach a GPS.
2. Go for a walk and sprinkle some point objects using GPS positioning
3. Copy the map from Android to a desktop and open in OOM
4. Observe that yhe point objects added using GPS coordinates no longer align with the GPX track from OOM Android.

Here's a screenshot from Android:

![oom_android](https://user-images.githubusercontent.com/22238167/91372935-43829080-e7d2-11ea-9333-cbef6e58f52e.png)

And here's one from the Mac:

![oom_android](https://user-images.githubusercontent.com/22238167/91373027-91979400-e7d2-11ea-807d-272b1ddfcbb9.png)

Notice that the track position where I added the purple plus sign shifts to the NE by about 1m. This shift is similar to the shift between NAD83 and WGS84 GPS lat/lon coordinates. Note that the shift only happens to the GPX track, not to the map objects that were added.

(I can close and restart Android OOM and the track does not move.)

Caveats:
The map I'm using was created using NAD 83 / UTM zone 13N and then converted to EPSG:6342 / NAD83(2011) when that option was added. 

Mapper Version: 0.9.3 and v20200613.1 (I've also tried 0.9.3)
Operating System: Android 7.0 and MacOS 10.15.6

dg0yt:
There is as a "GPX" setting in GDAL > Import with GDAL/OGR. Is it different on macOS vs. Android?

pkturner:
Could they be built with different versions of PROJ?

dg0yt:
> Could they be built with different versions of PROJ?

No, not for a single official release (v0.9.3) on Android, macOS, Windows. These binaries come out of Continuous Integration from the same source tarballs.

> There is as a "GPX" setting in GDAL > Import with GDAL/OGR. Is it different on macOS vs. Android?

This is important, because this represents different code paths.

sfroyen:
Android OOM did not have GPX checked under GDA/OGR. Once I checked that option, Android OOM behaves like the desktop version, i.e., the (saved) track is shifted. I will repeat the exercise with a new capture and report back.

sfroyen:
FYI - Unselecting the GDAL GPX option on the desktop and importing the GPX track again produces the initial Android result -- as expected.

dg0yt:
Okay, I was expecting this from line style on the screenshots. 
[...]

Issue #1264, opened 2019-06-13
UTM coordinates in NAD83 differ from WGS84


sfroyen:
### Steps to reproduce
1. Import Lidar generated contours and slope templates in UTM NAD83(2011) or more recent.
2. Import georeferenced templates in lat/lon WGS84 (e.g., Google satellite images)  or use an accurate GPS with OOM on Android.
3. Observe a systematic difference between the Lidar and the GPS positions.

### Actual behaviour
NAD83 and WGS84 in recent realizations no longer coincide. However, most lat/lon to UTM coordinate conversion software, including OOM, assumes they do. The error in Colorado is around 2 m, enough to be noticeable (at least to the mapper) on sprint maps. 

### Expected behaviour
OOM should perform lat/lon to UTM conversion correctly. Although it sounds like PROJ-6 should be able to handle this, it may be simpler to use a hack. For any orienteering map, the shift is a constant that can be implemented by a single delta-x delta-y pair.  Input fields for this can added to the map georeferencing screen (in lat/lon degrees, UTM meters or, possibly, even map coordinates mm). There are web sites that will the provide the shift (I'll follow up with one later) and survey precision GPS units usually come with software that does the same.

In principle, there could be data that come in UTM WGS84. I do not know if we need to handle such cases.

### Configuration
Mapper Version: latest
Operating System: all

dg0yt:
Can you provide correct pairs of geographic coordinates (WGS84) and projected coordinates (UTM) which are handled incorrectly by Mapper?

---

There is no doubt that "NAD83 and WGS84 in recent realizations no longer coincide." It is for a good reason:  
"NAD 83 is defined to remain constant over time for points on the North American Plate, whereas WGS 84 is defined with respect to the average of stations all over the world. Thus the two systems naturally diverge over time." [[Wikipedia on NAD]](https://en.wikipedia.org/wiki/North_American_Datum).

IIUC your key claims is that:  
*Mapper incorrectly transforms Lat/Lon from WGS84 (?) to UTM, and an extra x/y shift would be needed for correction.*

I would rather expect this is what can be said for NAD83, not UTM. There are "grid shift" files which provide the correction for NAD83. I never tried to make this work with Mapper. But even without these accurate files, it should be possible to to include a local shift adjustment to the NAD83 PROJ.4 specification.

[...]

PR #1757, opened 2020-10-21
Test positioning of GPS track in NAD 83 map

pkturner:
This is not a regular PR with potential source code for production. Consider it an example map if you wish. I hope it will be adopted for testing where NAD 83 applies.

This test code should be useful in monitoring issues GH-1264 and GH-1709. It extends the template track test, with

 - a map based on the NAD 83 datum

 - a series of template tracks with dates corresponding to various realizations of WGS 84

The `template-track-NA.xmap` is like `template-track.xmap`, with the CRS changed to EPSG:6342 (the CRS mentioned in GH-1709), and with a suitable reference point and grivation.
To create the test, the steps were:

 - Specify a rectangle in EPSG:6342 projected coordinates, of the same size as the track in template-track.gpx. This will be the position where tracks are created.

 | side | coordinate |
 | :---- | --------: |
 | top |    4429412.72 N |
 | left |    499280.37 E |
 | bottom | 4429393.45 N |
 | right   | 499305.52 E |

 - Convert the track rectangle to map coordinates.

 - Create 4 line objects in the map, outside of the track by 0.15 mm, similar to the lines in template-track.xmap.

 - Convert the track rectangle to NAD 83 (2011) geographic coordinates. EPSG:6318 applies. The conversion is a single step of UTM transformation.

 - For various dates between 1997 and 2020, convert the track rectangle from NAD 83 to WGS 84 coordinates using HTDP (see below).

 - For each date's WGS 84 track rectangle, create a file `template-track-NA.<year>.gpx` containing that date's coordinates, with the <time/> of the track as specified in the HTDP conversion.

Because the GPS coordinates of an NAD 83-based map are a moving target, Mapper should be sensitive to the <time/> element present in GPX tracks, if it is to position the track accurately. The provided tracks all have _different_ times and lat/lon coordinates in order to exercise this need.
    
### Conversions

Searching online, I found that the US's National Geospatial-Intelligence Agency maintains the WGS 84 reference frame and applies it to the GPS system's operational control. The G873 ... G1762 designate realizations of WGS 84 which have succeeded one another. The number 1762 indicates the GPS calendar week when G1762 was first used.

To best convert between WGS 84 and NAD 83, I used the National Geodetic Survey's HTDP (Horizontal Time-Dependent Positioning) utility online. When converting a position using HTDP, you specify its date and a particular realization of WGS 84.

I selected various dates to exercise the different reliazations of WGS 84 which were used for GPS from 1997 to the present.

 | epoch | realization | EPSG |
 | :--- | ---: | :---: |
 |    1997.0 | G873 |  EPSG:7659 |
 |   2001.0 | G873 |  EPSG:7659 |
 |   2003.0 | G1150 | EPSG:7661 |
  |  2012.0 | G1150 | EPSG:7661 | 
  | 2013.5 | G1674 | EPSG:7663  |
   | 2014.0 | G1762 | EPSG:7665 |
   | 2019.73 | G1762 | EPSG:7665 |

Eventually, I compared the results of the HTDP time-dependent conversions to conversions using PROJ 7.0.1 
   `cs2cs EPSG:6342 EPSG:7665 -f %.8fa`
finding that even though EPSG:7665 is not really valid before late 2013, the time-dependent Helmert transformation provided by PROJ is potentially usable. For all of the above dates, that single PROJ conversion is accurate to within 3 cm.

This PR includes a change to `template_track.cpp`, etc. to show how the map `template-track-NA.xmap` and its template tracks would appear in an ideal Mapper. That change is not intended as an answer to the GH-1264 and GH-1709 issues. When Mapper is built with all the commits on this branch (a4f1b50d), and the GDAL setting for importing GPX is disabled, the template tracks align nicely with one another and with the map objects. Thus, the test map and its template tracks are validated.

What I find notable:

- Since 1997, NAD 83 has moved about 40 cm relative to WGS 84.

- PROJ's use of the Helmert transformation defaults to the epoch which appears in its parameters. This date (1997) is arbitrary and has no significance to what Mapper is doing. If Mapper adopts EPSG:7665 or EPSG:9057 as the geographic CRS, it would be preferable  to use the current date.

- The original behavior of Mapper misses the mark by 1.37 m due to omitting any datum adjustment.

- The new behavior reported in GH-1709 uses the Helmert transformation at the 1997 epoch, reducing the error to 40 cm. While the new behavior is called a "regression" in some of the GH-1709 comments, it actually reduces the error by 70%.


pkturner:
Updated the `pkturner:datum-test` branch to **488f74b** to share common code with the `pkturner:template-consistency` branch.

pkturner:
Pushed a change which simplifies the "Improve accuracy" commit, shared with #1764.

pkturner:
Force-pushed changes the branch for compatibility with the current `master` which has some new `TemplateTrack` tests.

PR #1759, opened 2020-10-22
Fix GeoTIFF CRS reading on modern GDAL/PROJ

dg0yt:


dg0yt:
This change will cause WKT to be used in omap files, in the georeferencing saved for the template.
This doesn't seem to cause much harm: even Mapper on Ubuntu 18.04 (PROJ 4.9, GDAL 2.3) loads the template correctly, although PROJ 4.9 doesn't understand WKT.

PR #1764, opened 2020-10-28
Test consistency of GPS track positioning

pkturner:
This adds tests to `template_t `'s `templateTrackTest` and `ogrTemplateTest` which

- Check that an object created from a GPX track using v0.8.4 remains in the same position relative to the track.
- Check that an object created from a GPX track using v0.9.3 remains in the same position relative to the track.

It also adds `templateTypesConsistentTest` which checks that a track is positioned the same, independent of whether GDAL loading of GPX tracks is enabled.

These tests use maps georeferenced with EPSG:6342 which is based on the NAD 83 (2011) datum. They check for the issues mentioned in the discussion of #1709 "GPX tracks display in different position on desktop vs. Android (regression)" and thus several of them fail as of v0.9.4.

The tolerances used in these tests have been changed arbitrarily, to better demonstrate the tests' effectiveness.

### DIFF
--- a/src/templates/template_track.cpp
+++ b/src/templates/template_track.cpp
@@ -343,6 +343,46 @@ void TemplateTrack::unloadTemplateFileImpl()
 	track.clear();
 }
 
+QPointF TemplateTrack::getCenter() const
+{
+	qreal left;
+	qreal right;
+	qreal top;
+	qreal bottom;
+	int count = 0;
+	for (int i = 0; i < track.getNumSegments(); ++i)
+	{
+		int size = track.getSegmentPointCount(i);
+		for (int k = 0; k < size; ++k)
+		{
+			const TrackPoint& point = track.getSegmentPoint(i, k);
+			if (count == 0)
+			{
+				left = right = point.map_coord.x();
+				top = bottom = point.map_coord.y();
+			}
+			else
+			{
+				if (point.map_coord.x() < left)
+					left = point.map_coord.x();
+				else if (right < point.map_coord.x())
+					right = point.map_coord.x();
+				if (point.map_coord.y() < bottom)
+					bottom = point.map_coord.y();
+				else if (top < point.map_coord.y())
+					top = point.map_coord.y();
+			}
+			count += 1;
+		}
+	}
+	if (count > 0)
+	{
+		QPointF cp = QPointF((left+right)/2, (bottom+top)/2);
+		return is_georeferenced ? cp : templateToMap(cp);
+	}
+	return QPointF();
+}
+
 void TemplateTrack::drawTemplate(QPainter* painter, const QRectF& /*clip_rect*/, double /*scale*/, bool on_screen, qreal opacity) const
 {
 	painter->save();
--- a/test/template_t.cpp
+++ b/test/template_t.cpp
@@ -359,6 +359,64 @@ private slots:
 		QCOMPARE(qRound(latlon.latitude()), 50);
 		QCOMPARE(qRound(latlon.longitude()), 8);
 	}
+	
+	void templateTypesConsistentTest_data()
+	{
+		QTest::addColumn<QString>("map_file");
+		QTest::addColumn<int>("template_index");
+
+		QTest::newRow("TemplateTrack georef")     << QStringLiteral("testdata:templates/template-track.xmap") << 0;
+		QTest::newRow("TemplateTrack OGR georef") << QStringLiteral("testdata:templates/template-track.xmap") << 2;
+		QTest::newRow("TemplateTrack NAD83")      << QStringLiteral("testdata:templates/template-track-NA.xmap") << 0;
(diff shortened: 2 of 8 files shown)


pkturner:
Like #1757 this is not a PR with code that solves a problem; just a test to demonstrate a problem. It may help address the issue #1709, and when that issue is addressed, it may be a useful basis for a test to be added to `template_t`.

test/template_t.cpp
```diff
@@ -83,6 +84,17 @@ QPointF center(const Template* temp)
 	return temp->templateToMap(temp->getTemplateExtent().center());
 }
 
+void center(const Template* temp, QPointF &center_point, QDateTime &center_time)
```
dg0yt, line 87: Do you really want to deal with `center_time`? Rather than adding another overload of `center`, changing `center(const TemplateTrack* temp)`, but leaving `templateTrackTest`/`ogrTemplateTest` mostly unchanged, would look much clearer.

dg0yt (commented):
Thanks. Note that you can mark such PR as drafts.

test/template_t.cpp
pkturner: I thought `center_time` would be needed, but the test doesn't use it, not even for #1757. I will simplify.

dg0yt:
> It also adds `templateTypesConsistentTest` which checks that a track is positioned the same, independent of whether GDAL loading of GPX tracks is enabled.

Looks I lost one comment: this is meant to be already covered by `templateTrackTest` + `ogrTemplateTest` which each test four different configurations of GPX templates.

pkturner:
> > It also adds `templateTypesConsistentTest` 
> 
> this is meant to be already covered by `templateTrackTest` + `ogrTemplateTest` which each test four different configurations of GPX templates.

Covering this consistency directly is distinct enough that it seemed worth bringing up as a possible test.

pkturner:
> (They don't test the absolute geospatial position, but the relative position on the map.)

That's as it should be. Until the `geographic_crs_spec` issues are settled in relation to PROJ and GDAL, it's hard to say how the Georeferencing class could be tested for absolute position. That's basically what was needed for #1757, and the best I could do was create a map as proxy for the correct geospatial position.

src/templates/template_track.cpp
```diff
@@ -343,6 +344,53 @@ void TemplateTrack::unloadTemplateFileImpl()
 	track.clear();
 }
 
+void TemplateTrack::getCenter(QPointF &center_point, QDateTime &center_time) const
```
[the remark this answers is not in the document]
pkturner: Made it pure by dropping the calculation of `center_time`. The change has been pushed.

test/template_t.cpp
pkturner: Simplified as suggested. The change even removed `center(TemplateTrack *)` from `template_t.cpp`. The change has been pushed.

test/template_t.cpp
```diff
@@ -67,19 +67,10 @@ using namespace OpenOrienteering;
...
-	temp->drawTracks(&painter, false);
-	painter.end();
-	return QRectF(picture.boundingRect()).center();
-}
-
 QPointF center(const Template* temp)
 {
 	if (auto* template_track = qobject_cast<const TemplateTrack*>(temp))
-		return center(template_track);
+		return template_track->getCenter();
```
dg0yt, line 73: I found that I can simply use `template_track->calculateTemplateBoundingBox().center()`. It differs from the proposed `TemplateTrack::getCenter()` by:
- handling also waypoints.
- calling `templateToMap` for each point, instead of for the final result.

Given the low number of points in the test data, I will adopt to use the existing function, and not add `getCenter()`.

dg0yt (commented):
For now, I would amend the new tests with expect failures, and merge it. It may be useful for testing the changes I'm currently working on: Using an explicit CRS with geospatial vector data. This should allow to override GDAL's default "WGS 84" for GPX data, and thus enable finding more accurate transformations.

dg0yt:
> template-track-NA-084.xmap is also an example of a map which
triggers the warning about "Georeferencing mismatch".

In `TemplateTest::ogrTemplateTest`, it triggers a warning "ERROR 4: template-track-NA-2019.gpx: File or directory not found". 
- By the syntax, this message comes from GDAL.
- It is issued during map loading, i.e. before template loading.

I assume we try to load georeferencing information before the template file path is resolved. This is bound to fail.

dg0yt:
> In `TemplateTest::ogrTemplateTest`, it triggers a warning "ERROR 4: template-track-NA-2019.gpx: File or directory not found".
> 
> * By the syntax, this message comes from GDAL.
> * It is issued during map loading, i.e. before template loading.
> 
> I assume we try to load georeferencing information before the template file path is resolved. This is bound to fail.

This can be ignored for now. It is from the attempt to identify the template type from the contents. This should go away with the template type now being saved in the data.

PR #1771, opened 2020-11-03
Test NAD83 GPX templates

dg0yt:
Closes GH-1764 (pristine version of this PR).

@pkturner I can add you as a GH project member ("Registered Contributor" group) to acknowledge your input and to allow to add you formally as a reviewer to PRs.

### DIFF
--- a/test/data/templates/template-track-NA-084.xmap
+++ b/test/data/templates/template-track-NA-084.xmap
@@ -1,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="7">
+    <notes></notes>
+    <georeferencing scale="4000" declination="8.00" grivation="8.00">
+        <projected_crs id="EPSG">
+            <spec language="PROJ.4">+init=epsg:6342</spec>
+            <parameter>6342</parameter>
+            <ref_point x="500000.000000" y="4430000.000000"/>
+        </projected_crs>
+        <geographic_crs id="Geographic coordinates">
+            <spec language="PROJ.4">+proj=latlong +datum=WGS84</spec>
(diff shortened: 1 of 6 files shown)


pkturner:
In `TemplateTrack` I did not see `calculateTemplateBoundingBox` with its function so close to the `getCenter` which I added. It's much nicer the way you reuse that method.

pkturner:
I'd be happy to review some PRs, in the formal way.

dg0yt:
> In `TemplateTrack` I did not see `calculateTemplateBoundingBox` with its function so close to the `getCenter` which I added. It's much nicer the way you reuse that method.

Neither did I when I wrote the `QPicture` trick ;-)

dg0yt:
I'm still editing this series of patches...

dg0yt:
Now the tests also "expect" the failures on PROJ 4.9/GDAL 2 (Ubuntu 18.04).

pkturner (commented):
All the changes look good.

I suggest noting alongside the test code or data that the GPX track referenced by the `TemplateTrack NAD83` and `OgrTemplate NAD83` tests is specific to the 2019.73 epoch.

I trust your judgment regarding whether to add an explanation that the NAD83 datum is shifting about 2 cm per year, and that with Mapper built with PROJ 7.1.0 and not specifying an epoch, the track positioning error is about 0.4 m. The test's tolerance is set to 0.125 mm (0.5 m), such that it will be effective for catching regressions.

dg0yt:
Regarding the epoch, we could add `<notes>` to the maps, or `<!--comments-->` to the GPX file. But the GPX file has timestamps anyway. And for having explicit CRS with vector data, will probaly add test cases which will add another element to template configuration, so it will be recorded in another way.

I still wonder whether to keep the separate consistency test. Another alternative is to merge the three tests into a single test (`vectorTemplateTest`) which starts with the GDAL part and is shortcut when the current data is not suitable for templateTrack.

PR #1772, opened 2020-11-03
treewide: Fix warnings reported by Clang 11

lpechacek:
A tiny change which will make build with Clang 11 possible in Fedora 33 and openSUSE Tumbleweed.

### DIFF
--- a/src/gdal/ogr_file_format.cpp
+++ b/src/gdal/ogr_file_format.cpp
@@ -502,10 +502,10 @@ namespace {
 	 * 
 	 * \see CutoutOperation::operator()
 	 */
-	class ClippingImplementation : public OgrFileImport::Clipping
+	class ClippingImplementation final : public OgrFileImport::Clipping
 	{
 	public:
-		~ClippingImplementation() final = default;
+		~ClippingImplementation() = default;
 		
 		explicit ClippingImplementation(MapCoordVector boundary)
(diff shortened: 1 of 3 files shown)


src/gdal/ogr_file_format.cpp
```diff
@@ -502,7 +502,7 @@ namespace {
 	 * 
 	 * \see CutoutOperation::operator()
 	 */
-	class ClippingImplementation : public OgrFileImport::Clipping
+	class ClippingImplementation final : public OgrFileImport::Clipping
```
dg0yt, line 505: I saw this warning just yesterday in Qt Creator (static analyzer), but didn't stop for investigation. Is this an error without `-Werror`? `final` on class *and* destructor seems redundant. (Same for class `Progress`).

dg0yt (commented):
Basically :+1: but see comment. Just merge (as rebase) when you think it is ready.

src/gdal/ogr_file_format.cpp
lpechacek: It is merely a warning. I took the liberty to eliminate most of the diagnostic messages that clang spit out. Moving the keyword from the destructor to the class level seems appropriate to me.

dg0yt (approved):

lpechacek:
Done. Thanks for the review, Kai!

PR #1774, opened 2020-11-05
Fix inconsistent track positioning

dg0yt:
Fixes `OgrTemplate` vs. `TemplateTrack` inconsistency.
Fixes inconsistency for different generations of PROJ.

### DIFF
--- a/test/template_t.cpp
+++ b/test/template_t.cpp
@@ -331,7 +331,7 @@ private slots:
 		QVERIFY(map.getTemplate(template_index)->loadTemplateFile());
 		QCOMPARE(temp->getTemplateState(), Template::Loaded);
 
-		QEXPECT_FAIL("OgrTemplate from v0.9.3", "Position changed to match TemplateTrack", Continue);
+		QEXPECT_FAIL("OgrTemplate NAD83", "Synthetic test file", Continue);
 		auto const expected_center = map.calculateExtent().center();
 		if (QLineF(center(temp), expected_center).length() > 0.125) // 50 cm
 			QCOMPARE(center(temp), expected_center);
@@ -376,7 +376,7 @@ private slots:
 		QVERIFY(map.getTemplate(template_index)->loadTemplateFile());
 		QCOMPARE(temp->getTemplateState(), Template::Loaded);
 
-		QEXPECT_FAIL("OgrTemplate from v0.9.3", "Position changed to match TemplateTrack", Continue);
+		QEXPECT_FAIL("OgrTemplate NAD83", "Synthetic test file", Continue);
 		auto const expected_center = map.calculateExtent().center();
 		if (QLineF(center(temp), expected_center).length() > 0.25) // 1 m
 			QCOMPARE(center(temp), expected_center);
--- a/test/template_t.cpp
+++ b/test/template_t.cpp
@@ -320,23 +320,46 @@ private slots:
 		QFETCH(QString, map_file);
 		QFETCH(int, template_index);
 		
-		Map map;
-		MapView view{ &map };
-		QVERIFY(map.loadFrom(map_file, &view));
-		
-		QVERIFY(map.getNumTemplates() > template_index);
-		auto const* temp = map.getTemplate(template_index);
-		QCOMPARE(temp->getTemplateType(), "TemplateTrack");
-		QCOMPARE(temp->getTemplateState(), Template::Unloaded);
-		QVERIFY(map.getTemplate(template_index)->loadTemplateFile());
-		QCOMPARE(temp->getTemplateState(), Template::Loaded);
-
-		QEXPECT_FAIL("OgrTemplate NAD83", "Synthetic test file", Continue);
-		auto const expected_center = map.calculateExtent().center();
-		if (QLineF(center(temp), expected_center).length() > 0.125) // 50 cm
-			QCOMPARE(center(temp), expected_center);
-		else
-			QVERIFY2(true, "Centers do match");
+		QBuffer buffer;
+		{
+			Map map;
+			MapView view{ &map };
+			QVERIFY(map.loadFrom(map_file, &view));
+			
+			QVERIFY(map.getNumTemplates() > template_index);
+			auto const* temp = map.getTemplate(template_index);
+			QCOMPARE(temp->getTemplateType(), "TemplateTrack");
+			QCOMPARE(temp->getTemplateState(), Template::Unloaded);
+			QVERIFY(map.getTemplate(template_index)->loadTemplateFile());
+			QCOMPARE(temp->getTemplateState(), Template::Loaded);
+			
+			QEXPECT_FAIL("OgrTemplate NAD83", "Synthetic test file", Continue);
+			auto const expected_center = map.calculateExtent().center();
+			if (QLineF(center(temp), expected_center).length() > 0.125) // 50 cm
+				QCOMPARE(center(temp), expected_center);
+			else
+				QVERIFY2(true, "Centers do match");
+			
+			QVERIFY(map.exportToIODevice(buffer));
+		}
+		{
+			Map map;
+			QVERIFY(map.importFromIODevice(buffer));
+			
+			QVERIFY(map.getNumTemplates() > template_index);
+			auto const* temp = map.getTemplate(template_index);
+			QCOMPARE(temp->getTemplateType(), "TemplateTrack");
+			QCOMPARE(temp->getTemplateState(), Template::Unloaded);
+			QVERIFY(map.getTemplate(template_index)->loadTemplateFile());
+			QCOMPARE(temp->getTemplateState(), Template::Loaded);
+			
+			QEXPECT_FAIL("OgrTemplate NAD83", "Synthetic test file", Continue);
+			auto const expected_center = map.calculateExtent().center();
+			if (QLineF(center(temp), expected_center).length() > 0.125) // 50 cm
+				QCOMPARE(center(temp), expected_center);
+			else
+				QVERIFY2(true, "Centers do match");
+		}
 	}
 	
 #ifdef MAPPER_USE_GDAL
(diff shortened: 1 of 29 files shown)


test/template_t.cpp
```diff
@@ -354,8 +354,13 @@ private slots:
 		QVERIFY(map.getTemplate(template_index)->loadTemplateFile());
 		QCOMPARE(temp->getTemplateState(), Template::Loaded);
 
+#if !defined(ACCEPT_USE_OF_DEPRECATED_PROJ_API_H) || PJ_VERSION >= 600
+		QEXPECT_FAIL("TemplateTrack from v0.8.4", "Unsupported WGS 84 -> NAD 83 transformation", Continue);
+		QEXPECT_FAIL("OGRTemplate from v0.9.3", "Unsupported WGS 84 -> NAD 83 transformation", Continue);
```
dg0yt, line 359: New errors with modern PROJ.

test/template_t.cpp
```diff
@@ -354,8 +354,13 @@ private slots:
...
+#else
 		QEXPECT_FAIL("TemplateTrack NAD83", "Unsupported WGS 84 -> NAD 83 transformation", Continue);
 		QEXPECT_FAIL("OgrTemplate NAD83", "Unsupported WGS 84 -> NAD 83 transformation", Continue);
```
dg0yt, line 362: Errors now limited to legacy PROJ.

test/template_t.cpp
```diff
@@ -427,10 +432,6 @@ private slots:
 			ogr_template_center = center(temp);
 		}
 		
-#if !defined(ACCEPT_USE_OF_DEPRECATED_PROJ_API_H) || PJ_VERSION >= 600
-		QEXPECT_FAIL("TemplateTrack NAD83", "Unsupported WGS 84 -> NAD 83 transformation", Continue);
-		QEXPECT_FAIL("OgrTemplate NAD83", "Unsupported WGS 84 -> NAD 83 transformation", Continue);
-#endif
```
dg0yt, line 433: Inconsistency resolved.

dg0yt:
Now I see a fairly consistent behaviour for this code built against PROJ 7.1 and PROJ 4.9.
- The 093 tests consistently fails. I assume this is the next step: using WGS 84 (G1762) for coordinate conversion.
- I still left the global "georeferencing mismatch" warning unchanged. This might need a review.
  - At least for some test maps, the mismatch is now removed. OTOH, there may be a mismatch now for features drawn aligned to the templates shown misplaced in v0.9.4.
  - The global warning uses a slightly different trigger than the new fix: The former looks at projected coordinates (>0.9 m). For the fix, I chose to look at map coordinates (>= 0.1 mm).

pkturner (commented):
I committed a suggested minor change. No problem if you want to undo it. 

Regarding the tests for track positioning consistency, I view them as catching inconsistencies that potentially cause difficulties to users. I am not expecting that a fully satisfactory fix is possible.

a43b8ad8 OgrFileImport: Fix SRS/transform handling     
This change involves more GDAL than I can understand. I'm not able to review it effectively.

07138aa1 OgrTemplate: Fix inconsistent track positioning     
I like that this fix is lightweight and specific compared to the idea of setting GDAL's geographic CRS to a special version of WGS84. I just wonder whether at some future time GDAL will do a better job of transforming GPX tracks to EPSG:6342 coordinates, and the need for this special case will vanish.

0acab0cc Track/OgrTemplate: Handle georef inconsistency    
This fixes up the track in a case where the map's Georeferencing has an inconsistent geographic reference point. Rather than position the track according to the map's Georeferencing, the track is adjusted by the offset which would bring the map's geographic reference point into agreement with its projected reference point.

That will fix a potential problem if

- The map's Georeferencing was set using the same Mapper version that handled the template track.
- The user prefers the track to keep its former, likely less accurate position.

Since this latter commit generates broken corner cases, consider instead leaving the track handling as is. The response to user issues would be: 

> When Mapper is revised, and when it is rebuilt with a new version of PROJ or GDAL, the accuracy of its datum conversions may be improved. This type of change affects the alignment between georeferenced templates and the map.

dg0yt:
Thanks for review!

> I committed a suggested minor change. No problem if you want to undo it.

I had this version too, but I wanted to avoid to permanently change the mismatching geospatial points. Now the fix is limited to georeferenced template-track-compatibility mode, and I verified that we don't write the georeferencing to the XML file, so this concern is not relevant. 
Looking at the offset in paper dimensions still has its point. Do we care about 0.5 m real offset at 1:10000: 0.05 mm?

> Regarding the tests for track positioning consistency, I view them as catching inconsistencies that potentially cause difficulties to users. I am not expecting that a fully satisfactory fix is possible.

Indeed. I'm still using Ubuntu 18.04 with PROJ 4.9, but my Android tablet is using the same modern version as packaged for Windows or macOS.

> [a43b8ad](https://github.com/OpenOrienteering/mapper/commit/a43b8ad85c97cc160692d0ee0fe5c84a3a0ae905) OgrFileImport: Fix SRS/transform handling
> This change involves more GDAL than I can understand. I'm not able to review it effectively.

This is not so much related to GDAL but a separate cleanup in our API:
- The return value of `setSrs()` indicates whether to continue at all.
- Iff there is a `data_transform` (after `setSrs()`), use it.

This change became necessary by the following change which broke the assumption that we always have a transformation when `setSrs()` returns true. (And there may be other cases where this assumption breaks).

> [07138aa](https://github.com/OpenOrienteering/mapper/commit/07138aa15f4682a89fec06ef8f4bd6089d19b706) OgrTemplate: Fix inconsistent track positioning
> I like that this fix is lightweight and specific compared to the idea of setting GDAL's geographic CRS to a special version of WGS84. I just wonder whether at some future time GDAL will do a better job of transforming GPX tracks to EPSG:6342 coordinates, and the need for this special case will vanish.

This fix is looking backwards, on legagcy template track compatibility. It ensures that transformation is not done via OGR SRS API but literally through the same setup of georeferencing object as in class `TemplateTrack`.
For the future, for improving the situation with `EPSG:9057` GPX and `EPSG:63242` map, we need a new mode of compatibility between `TemplateTrack` and `OgrTemplate`.

> [0acab0c](https://github.com/OpenOrienteering/mapper/commit/0acab0ccf750637c0a9d5cddf7a8e534281f9e60) Track/OgrTemplate: Handle georef inconsistency
> This fixes up the track in a case where the map's Georeferencing has an inconsistent geographic reference point. Rather than position the track according to the map's Georeferencing, the track is adjusted by the offset which would bring the map's geographic reference point into agreement with its projected reference point.
> 
> That will fix a potential problem if
> 
>  * The map's Georeferencing was set using the same Mapper version that handled the template track.
> 
>  * The user prefers the track to keep its former, likely less accurate position.
> 
> 
> Since this latter commit generates broken corner cases, consider instead leaving the track handling as is. The response to user issues would be:
> 
> > When Mapper is revised, and when it is rebuilt with a new version of PROJ or GDAL, the accuracy of its datum conversions may be improved. This type of change affects the alignment between georeferenced templates and the map.

This is the same direction as "flagging" the affected templates. This will need some different way of unit testing.

I was wondering whether to store the version of PROJ somewhere but this is not enough. We may also need some regional grid shift files to reproduce precise results. So detecing and flagging the inconsistency of reference points may be all that we can do.

dg0yt:
`template-track-NA-093-GDAL.xmap` showcases our dilemma: It is now *consistent* independent of the GPX setting. But the fix *changed the position* to where TemplateTrack would put it.

I studied if I can look at the saved template type and change the behaviour if I find `"OgrTemplate"` to consistently use the positioning which was created with GDAL. It turned out to be doable. 
With his change, the second template (`"OgrTemplate"`) of `template-track-NA.xmap` will fail. But this is a synthetic test file. Files saved from Mapper won't mix `OgrTemplate` and `TemplateTrack` types with same expected alignment.

But what I really learned (with `PROJ_DEBUG` set to 3) is the actual difference between the transformations used by our template classes:

- `OgrTemplate` uses the CRS provided by GDAL/OGR: `SRS_WKT_WGS84_LAT_LONG` which is defined as EPSG:4326:
https://github.com/OSGeo/gdal/blob/62e48c92acded33b40873380543a2937fb01bea1/gdal/ogr/ogr_srs_api.h#L70-L71

  When creating pipelines to EPSG:6342, this results in a trivial 3-step-pipeline of 
  axisswap > unitconvert > utm, 
  probably similar to:
  ~~~
  $ projinfo -o PROJ  -s WGS84 -t EPSG:6342
  [...]
  +proj=pipeline +step +proj=axisswap +order=2,1 +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=utm +zone=13 +ellps=GRS80
  ~~~

- `TemplateTrack` uses our geographic CRS definition which is `+proj=latlong +datum=WGS84`.

  This results in a 7-step-pipeline of
  unitconvert > push > cart > helmert > inv > pop > utm, 
  probably similar to:
  ~~~
  $ export PROJ_USE_PROJ4_INIT_RULES=YES
  $ projinfo -o PROJ  -s '+type=crs +proj=latlong +datum=WGS84' -t '+init=epsg:6342'
  [...]
  +proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=push +v_3 +step +proj=cart +ellps=WGS84 +step +proj=helmert +x=0.99343 +y=-1.90331 +z=-0.52655 +rx=0.02591467 +ry=0.00942645 +rz=0.01159935 +s=0.00171504 +convention=coordinate_frame +step +inv +proj=cart +ellps=GRS80 +step +proj=pop +v_3 +step +proj=utm +zone=13 +ellps=GRS80
  ~~~

I still need to turn the proof-of-concept into reviewable commits. Since I need to move away from our geographic CRS definition, it includes changes similar to #1598.

---

For maximum reproducibility of template positioning, we might want to save not just the CRS but the actual transformation pipeline definitions. This would remove most of the dangerous assumptions about the transformation process.

pkturner:
> Looking at the offset in paper dimensions still has its point. Do we care about 0.5 m real offset at 1:10000: 0.05 mm?

I agree that paper dimensions are a better measure of whether the shift is large enough to be helpful.

Another thought was, "If the presence of a large offset means that Mapper should fix up the track, why not also fix up the track when the offset is so small as not to be noticed?"

pkturner:
> For maximum reproducibility of template positioning, we might want to save not just the CRS but the actual transformation pipeline definitions. This would remove most of the dangerous assumptions about the transformation process.

To understand that this is feasible, I would need to look into how the API of PROJ supports dumping the pipeline. I have used projinfo and the old-style PROJ string pipelines, though not WKT pipelines. Helmert transformation parameters could easily be saved, but I expect a grid shift would be too large. I like this approach to "reproducibility of template positioning".

dg0yt:
> Another thought was, "If the presence of a large offset means that Mapper should fix up the track, why not also fix up the track when the offset is so small as not to be noticed?"

1. We are still unsure about which corner cases we break.
2. We need to be aware of numerical stability. However, this might be sufficiently covered by a `qIsNull` length test.

dg0yt:
> To understand that this is feasible, I would need to look into how the API of PROJ supports dumping the pipeline. I have used projinfo and the old-style PROJ string pipelines, though not WKT pipelines. Helmert transformation parameters could easily be saved, but I expect a grid shift would be too large. I like this approach to "reproducibility of template positioning".

I think that all what `projinfo` shows can be exported and reused. Both PROJ and WKT syntax include Helmert parameters. (Grid files are a different story, but probably we could detect their names, and we might help users carrying them around alongside maps or templates.)
But I didn't find a similar feature in GDAL. All I know is that I can [create transformations](https://gdal.org/api/ogr_srs_api.html#ogr__srs__api_8h_1a375495b20ad75ec886fcf25858a6c94b) either in a way that comes down to `proj_create_crs_to_crs`, or by supplying "a user defined coordinate transformation pipeline" (as understood by PROJ, probably).

dg0yt:
>  * `OgrTemplate` uses the CRS provided by GDAL/OGR for GPX data: `SRS_WKT_WGS84_LAT_LONG` which is defined as EPSG:4326:

GDAL 2 (Ubuntu 18.04, or Mapper 0.8.4) uses `+proj=longlat +datum=WGS84 +no_defs`.

dg0yt:
The adjustment is now done on the projected coordinates, and type-based adjustments are added.

Still, not yet ready: The following use case breaks:
1. Enable GDAL GPX setting.
2. Load NA-93-PROJ test file.
   (GPX track template is marked with type "TemplateTrack", but loaded with OgrTemplate.)
   **Positioning is fixed** with these changes.
3. Save the file (possible with a different name).
   (GPX track template is saved from OgrTemplate, and marked with this type.)
4. Close the file.
5. Re-open the saved file.
   (GPX track template is loaded with OgrTemplate again.)
   **Positioning is broken now.**

How to continue:
- We need to think how we deal with the map's broken georeferencing, and instead of just adding a warning, we might need to update it, too. 
  - It is reused by these templates, so it contributes to the positioning problems.
  - I would tend to update the geographic reference point.
- In positioning fixup, we might change the templates from georeferenced (automatic position) to manual positioning. Well, for the vector templates, this requires more changes from my local WIP branches.

pkturner:
>     In positioning fixup, we might change the templates from georeferenced (automatic position) to manual positioning.   
One way to prevent future inconsistency of the type detected by GH-#1771 (the `nad83-tests` branch) would be to convert georeferenced templates to manual positioning. In the excerpt quoted above, I think the purpose was different. Converting templates is not without drawback, but the idea has stuck in my thoughts for a while.

dg0yt:
I decided to take a step back in approaching the inconsistencies. I now try to isolate
a) the effect of the PROJ version,
b) the effect of GDAL/OGR transformations,
c) the effect of Mapper's different classes.

For now, this resulted in another bug report to PROJ:
https://github.com/OSGeo/PROJ/issues/2442 WGS84 axis order affects choice of transformation to EPSG:6342

dg0yt:
> For now, this resulted in another bug report to PROJ:
> [OSGeo/PROJ#2442](https://github.com/OSGeo/PROJ/issues/2442) WGS84 axis order affects choice of transformation to EPSG:6342

Lesson learnt so far: 
- Legacy PROJ has no clue how to improve the EPSG:4326-to-EPSG:6342 transformation. With legacy PROJ, this affects `TemplateTrack` and `OgrTemplate` alike.
- Modern GDAL with modern PROJ, as used by `OgrTemplate`, would select only EPSG-defined transformations when asked for EPSG:4326-to-EPSG:6342. Such a transformation does not exist. Thus modern OgrTemplate produces the same result as legacy PROJ.
- Modern PROJ, as used by `Georeferencing` and `TemplateTrack` with our legacy definition of WGS84 (non-EPSG) would select any pre-defined transformation. Here, it finds an ESRI-based transformation one. 

Conclusions:
For modern PROJ, we might want to switch to EPSG:4326 (authoritative, similar to GDAL, thus aligning transformations), and take care of providing geographic data in matching axis order.
However, I also remember the proposal to switch to the latest realisation of WGS84 instead.

dg0yt:
Abandoning this approach in favor of #1780.

PR #1780, opened 2020-11-22
Another approach to fix track positioning consistency

dg0yt:
... now reverting to v0.8 behaviour.

TODO: catch v0.9.0-4 maps.


dg0yt:
| Mapper version | PROJ version | XML format version | Remark |
|---|---|---|---|
|v0.8.4|4.9.3 (Superbuild)|7|**Consistent.**|
|v0.9.0|6.2.0 (Superbuild)|9|Assumed to be inconsistent in the same way as v0.9.4.|
|v0.9.4|7.1.1 (Superbuild)|9|**Affected by #1709:** `Georeferencing`/`TemplateTrack` use different transformation than GDAL or PROJ 4.9. `TemplateTrack` misplaces templates from files created with PROJ < 6 (`template-track-NA-084.xmap`). | 
|"|4.9.3 (Ubuntu 18.04)|9|Not affected by #1709: `Georeferencing`/`TemplateTrack` uses same transformation as GDAL. Templates created with `TemplateTrack` on PROJ >= 6 are misplaced (`template-track-NA-093-PROJ.xmap`).|
| This PR|7.1.1 (Superbuild)|9|**#1709 resolved:** `Georeferencing`/`TemplateTrack` use same transformation as GDAL and PROJ 4.9. Templates created with TemplateTrack on PROJ >= 6 are misplaced (template-track-NA-093-PROJ.xmap). |
| "|4.9.3 (Ubuntu 18.04)|9|Not affected by #1709: `Georeferencing`/`TemplateTrack` uses same transformation as GDAL and modern PROJ. Templates created with TemplateTrack on PROJ >= 6 are misplaced (template-track-NA-093-PROJ.xmap).|
|v0.9.5|any|9 (?!)|**#1709 resolved:** `Georeferencing`/`TemplateTrack` uses same transformation as GDAL and PROJ 4.9. **Desired:** Inconsistencies wrt v0.9.0-v0.9.4 mostly resolved.|

There is some more variety wrt to inconsistency, depending on when the map was created on how it was edited. When we see a map, we cannot find out exactly its history. Trying to analysze:
|XML format version|Map `Georeferencing` consistent|Resolution|
|-----|--------|-------------|
|7    |   -   | Map was created before v0.9.0, and not edited in a later version. *Desired resolution:* Do nothing. | 
|8, 9| yes | 1a) Map was created in v0.9.4, but without inconsistencies (e.g. UTM/WGS84). *Desired resolution:* Do nothing.|
|8, 9| yes | 1b) Map was created before v0.9.0, and edited in v0.9.4, but without inconsistencies (e.g. UTM/WGS84). *Desired resolution:* Do nothing.|
|8, 9| yes | 1c) Map was created before v0.9.0, but edited in v0.9.4, ignoring inconsistencies (e.g. UTM/NAD83). *Desired resolution:* Unsure. `TemplateTrack`s added in v0.9.4 would now be placed in a different location. We might fix this particular case. However, there may also be tracks added before v0.9. Simplest solution: Do nothing.
|8, 9|  no  | 2a) Map was created in v0.9.4 with modern PROJ, in a now inconsistent way (e.g. UTM/NAD84). `TemplateTracks`  would now be placed in a different location. *Desired resolution:* Unsure. We might fix this particular case only if we accept this solution also for 1c. Simplest solution: Do nothing.
|8, 9|  no  | 2b) Map was created before v0.9.0 and edited in v0.9.4. Inconsistencies in the map's georeferencing were removed (e.g. UTM/NAD83). `TemplateTracks` would now be placed in a different location. *Desired resolution:* Unsure. We might fix this particular case, but we don't no how the use dealt with the previous inconsistencies. Simplest solution: Do nothing.

I'm not sure if I managed to correctly analyze all scenarios. However, I feel like we should just fix the inconsistency as is, even if it adds some possibly undesired changes to the NAD83 world again. But AFAIU NAD83 is served better by finally allowing to override the EPSG:4326 source CRS for GPX with a a more relevant realization, instead of adding more complex migration paths between various configurations of Mapper and PROJ versions.

pkturner:
The more essential NAD83 issue is #1264, which differs from the consistency issue #1709. (The tests proposed in #1757 are aimed at the former.) It's good to separate the issues.

> I feel like we should just fix the inconsistency as is, even if it adds some possibly undesired changes to the NAD83 world again. But AFAIU NAD83 is served better by finally allowing to override the EPSG:4326 source CRS for GPX with a a more relevant realization

That would be an excellent improvement in the longer term, which would address #1264 for exactly the users who need it. Is that change feasible? I don't see that GDAL supports any datum for GPX tracks except EPSG:4326, unless overridden globally with a different setting of the `SRS_WKT_WGS84_LAT_LONG` macro. Changing the macro would not give Mapper users the nice flexibility hinted at.

pkturner (commented):
I love the simplicity of the change to the `Georeferencing` code, achieving consistency between `OgrTemplate` and `TemplateTrack` behaviour, as well as consistency with some previous versions of Mapper.

The `template-track-NA.xmap` map has objects carefully positioned relative to projected coordinates. I did not craft its `geographic_crs` `ref_point_deg` to be part of a test. I doubt adding it to  `testFileGeorefTest`  improves the test.

dg0yt:
> The `template-track-NA.xmap` map has objects carefully positioned relative to projected coordinates. I did not craft its `geographic_crs` `ref_point_deg` to be part of a test. I doubt adding it to `testFileGeorefTest` improves the test.

I see. However, I prefer to have this documented in the expected failure, instead of having to remember this fact when wondering why this file omitted. Maybe you can propose a better wording. Or we fix the geographic coordinates.

dg0yt:
> I don't see that GDAL supports any datum for GPX tracks except EPSG:4326, unless overridden globally with a different setting of the `SRS_WKT_WGS84_LAT_LONG` macro. Changing the macro would not give Mapper users the nice flexibility hinted at.

We don't have to patch GDAL. We can use an alternative source CRS when constructing the transformation when we find (particular combinations of GDAL driver and) WGS84.

PR #1782, opened 2020-11-23
OgrFileImport: Initialize data_srs on construction

lpechacek:
This commit fixes a bug found by Valgrind. This is a minimal fix to the issue at hand, which raises a question of to what degree we want to adhere to the ES.20 (Always initialize an object) guideline. `OgrFileImport` class contains, in its `private:` section, an even mix of constructor-initialized members and member-initialized ones. For instance, the `default_*_symbol` members are initialized in the single class constructor, and the many POD members (`empty_geometries`, `no_transformation`, ...) have default member initializers. `OGRSpatialReferenceH`, which is a pointer in disguise, fell through the cracks.

Do we want to routinely provide in-class default member initializers for anything that might potentially be a POD, or do we stick to the current habits and rely on specialized tools to spot the bugs?

### DIFF
--- a/src/gdal/ogr_file_format_p.h
+++ b/src/gdal/ogr_file_format_p.h
@@ -349,7 +349,7 @@ private:
 	
 	ogr::unique_srs map_srs;
 	
-	OGRSpatialReferenceH data_srs;
+	OGRSpatialReferenceH data_srs = {};
 	
 	ogr::unique_transformation data_transform;
 	


dg0yt:
> Do we want to routinely provide in-class default member initializers for anything that might potentially be a POD, or do we stick to the current habits and rely on specialized tools to spot the bugs?

In general, I would prefer to routinely provide in-class default member initializers if it is trivially possibly and useful. In particular, it shall not require additional includes.

dg0yt (approved):
Thanks!

dg0yt:
How to proceed?

lpechacek:
> How to proceed?

Sorry for delays. I had a dental surgery on Tuesday, si everything is even slower than usual.

The next steps are
1) Document the coding agreement about member initialization in coding style.
2) Merge this commit to `master`.

I kept the merge request open as a reminder about 1). I'll reverse the order of steps for the sake of neater status appearance.

Thanks for the ping. 😀

lpechacek:
>     1. Document the coding agreement about member initialization in coding style.

FTR, I've added a sentence to the Coding Conventions: https://github.com/OpenOrienteering/mapper/wiki/Coding-conventions/_compare/6c518374a5d31b1b8f7a244babb471ca7a4f776b...d6d41390ab96141e818bbb6121076acc9bf06d0b

PR #1793, opened 2020-12-06
Add KML/KMZ ground overlay support for templates, import and export

dg0yt:
This is a proof-of-concept exploring the feasibiliy of:
- using GDAL's virtual file system support
- extending `OgrFileImport` into creating child templates
- extending `OgrTemplate` (`TemplateMap`) to display child templates

With GDAL's LIBKML driver (available in Linux distro's GDAL, but not yet in our Superbuild), this allows for viewing KMZ files as used by [MapRunF](http://maprunners.weebly.com/maprunf.html), as created following: 
  http://maprunners.weebly.com/maps---kmz-files.html 
via 
  http://www.p.fne.com.au/rg/cgi-bin/o-utils.cgi 

It could be used to complement a future MapRunF compatible KMZ export.

With Mapper loading such files as a map, and already being able to export geospatial vector data, it is also possible to open a KMZ file, draw a single path object for a course, and export as KML. However, I didn't test if this works with MapRunF.


src/templates/template.h
```diff
@@ -297,7 +298,16 @@ Q_OBJECT
...
 	 */
-	LookupResult tryToFindTemplateFile(const QString& map_path);
+	virtual LookupResult tryToFindTemplateFile(const QString& map_path);
+	
+	/**
+	 * A check if the template file exists.
+	 * 
+	 * The default implementation checks if the file exists in the file system.
+	 * This maybe overriden to handle virtual file systems such as KMZ files.
+	 * If returns false, template loading will fail with "No such file.".
```
dg0yt, line 308: "If it returns ..."

dg0yt (commented):
The `Template::isRasterGraphics()` property does not match hybrid templates.

dg0yt:
Now correctly rendering and exporting KMZ files with groundoverlay raster tiles (JPG, 300 dpi, 512 px). Verified with Google Earth. Should work with MapRunF.
Needs GDAL for export. Needs GDAL with LibKML for reading.

Eric-P-Kmz:
Hi,

I participate in the development of the free Vikazimut application. Is it possible to do a kmz export of the same format as that of OCAD in OOM (with a single image in the export)?
We use the OCAD export format to georeference the maps to import into the app. The division into several images of the map is not compatible with our import.

dg0yt:
> Is it possible to do a kmz export of the same format as that of OCAD in OOM (with a single image in the export)?
> We use the OCAD export format to georeference the maps to import into the app. The division into several images of the map is not compatible with our import.

Before adding more code and UI to care for the limitations of Vikazimut, did you consider extending Vikazimuth to handle tiles?

Anyway, it is possible and not too hard to do add an KMZ export with a single raster image instead of tiles. It might come with a simple option to choose the tile size ("256x256", "512x512", **"No tiles"**).

dg0yt:
@Eric-P-Kmz Does the attached file work for you? It contains a KMZ of the complete map example.
[tmp.zip](https://github.com/OpenOrienteering/mapper/files/6038461/tmp.zip)

Eric-P-Kmz:
This format is perfect for Vikazimut.
Régis is currently absent but he is contacting you this weekend for your participation in the project. Thank you for your interest

dg0yt:
Merci beaucoup. I'm looking into simple course files right now.

PR #1814, opened 2020-12-22
Georeferencing fixes

dg0yt:


dg0yt:
The key contribution of this PR is allowing "natural" transformations with Georeferencing objects which actually use a geographic CRS in place of the projected CRS.
- The unit "angular degrees" is retained in projected coordinates.
- "Projected" axis order (QPointF x-y) is easting-northing regardless of CRS axis order.
- Behaviour is identical behaviour for legacy PROJ, modern PROJ and GDAL coordinate transformations.
- The claimed behaviour is covered by test data (https://github.com/OpenOrienteering/mapper/pull/1814/commits/cd3f7984a22c14c4212c7352f8e4e35c7f9d8018).

I would really like to integrate this change soon, but I would like to invite review/testing before doing so, with focus on unexpected behavioural changes with templates which use geographic CRS.

pkturner:
I had one test failure, having built the `georeferencing-fixes` branch on Gentoo Linux with GDAL 3.0.4 and PROJ 7.0.1.

    $ test/template_t   
    ********* Start testing of TemplateTest *********   
    Config: Using QtTest library 5.15.1, Qt 5.15.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 9.3.0)   
    PASS   : TemplateTest::initTestCase()   
    PASS   : TemplateTest::gdalUtilTest()
    PASS   : TemplateTest::worldFilePathTest()
    PASS   : TemplateTest::worldFileUnitTest()
    PASS   : TemplateTest::worldFileTemplateTest()
    PASS   : TemplateTest::templatePathTest()
    PASS   : TemplateTest::templateImageDrawableTest()
    ERROR 4: гео.tiff: No such file or directory
    PASS   : TemplateTest::geoTiffTemplateTest()
    FAIL!  : TemplateTest::geoTiffSRSTest() Compared values are not the same
       Actual   (projected): QPointF(310000,518999)
       Expected (expected) : QPointF(310000,519000)
       Loc: [/home/scott/mapper/alternate/test/template_t.cpp(406)]
    PASS   : TemplateTest::templateTrackTest(TemplateTrack georef)
    ...

pkturner:
I noticed that PROJ documents `proj_angular_output` as indicating whether the output of the projection is in radians. There's also `proj_degree_output` that tells whether the output is in degrees. I'd consider the projection _geographic_ if its output is in either radians or degrees.

At the same time, `isGeographic()` is no longer used outside the `Georeferencing` class. It is used only to control whether to convert between radians and degrees. So the method's name is inconsistent with how it is used.

The implementation

	case PJ_TYPE_GEOGRAPHIC_CRS:
	case PJ_TYPE_GEOGRAPHIC_2D_CRS:
	case PJ_TYPE_GEOGRAPHIC_3D_CRS:
		return true;
	case PJ_TYPE_OTHER_COORDINATE_OPERATION:
		return bool(proj_angular_output(pj, PJ_FWD));
appears to test for the type of projection in some cases and for the need for unit conversion in another.

pkturner:
Not exactly on topic:   
Having implemented Mapper's automatic grid scale factor calculation on the assumption that the grid units are meters, I keep getting surprised that the `Georeferencing` class works rather well if the grid uses other units. The trick is that the _automatic_ "grid scale factor" is more than the regular grid scale factor, also incorporating a units factor.

dg0yt:
> Not exactly on topic:
> Having implemented Mapper's automatic grid scale factor calculation on the assumption that the grid units are meters, I keep getting surprised that the `Georeferencing` class works rather well if the grid uses other units. The trick is that the _automatic_ "grid scale factor" is more than the regular grid scale factor, also incorporating a units factor.

All these situations are purely internal, used with templates but not with the main map. As long as we don't loose numeric precision, the actual factor value doesn't play a significant role.

dg0yt:
> I noticed that PROJ documents `proj_angular_output` as indicating whether the output of the projection is in radians. There's also `proj_degree_output` that tells whether the output is in degrees. I'd consider the projection _geographic_ if its output is in either radians or degrees.

I noticed `proj_degree_output`, too, but it is new in PROJ 7.1.0. I don't have a use case/test case for it at the moment. My starting point was inconsistent behaviour for WGS84, either with regular axis order or swapped order. Oh, and I didn't even notice the `in radians`...

For the record, the starting point of the PROJ changes:
https://github.com/OSGeo/PROJ/issues/2027

> At the same time, `isGeographic()` is no longer used outside the `Georeferencing` class. It is used only to control whether to convert between radians and degrees. So the method's name is inconsistent with how it is used.

This is a good point. Since we care only about the unit, we should resolve this. But I would postpone this change to before the 1.0 release, when I also want to cut legacy PROJ support.

dg0yt:
> > I noticed that PROJ documents `proj_angular_output` as indicating whether the output of the projection is in radians. There's also `proj_degree_output` that tells whether the output is in degrees. I'd consider the projection _geographic_ if its output is in either radians or degrees.
> 
> I noticed `proj_degree_output`, too, but it is new in PROJ 7.1.0. I don't have a use case/test case for it at the moment. My starting point was inconsistent behaviour for WGS84, either with regular axis order or swapped order. Oh, and I didn't even notice the `in radians`...

Maybe "in radians" is a PROJ documentation error, and it really just means angular units? The geographic CRS we test with do use degrees, and still `proj_angular_output` gives the answer `isGeographic` needs to know.

And it seems to me that we don't have to care about unit degrees or radians in Mapper: Whether going the class Georeferencing way (i.e. via WGS84), or directly CRS-to-CRS via PROJ, or via GDAL: At some point an external transformation will take care of the unit. The bug is to *sometimes* fiddle with the unit in the current implementation, leading to the inconsistent behaviour which is adressed by this PR.

dg0yt:
I would really like to merge this PR here soon, but I want to see the `proj_angular_output` issue sorted out first:

> Maybe "in radians" is a PROJ documentation error, and it really just means angular units? The geographic CRS we test with do use degrees, and still `proj_angular_output` gives the answer `isGeographic` needs to know.

@pkturner Can you verify this? If it is a PROJ documentation error, we should submit a PR to PROJ.

pkturner:
@dg0yt 
> Maybe "in radians" is a PROJ documentation error, and it really just means angular units? The geographic CRS we test with do use degrees, and still `proj_angular_output` gives the answer `isGeographic` needs to know.

With an examination of the PROJ 7.2.0 source code, I verified that the documentation is accurate.  Internally, the units are one of

    PJ_IO_UNITS_WHATEVER  = 0,  /* Doesn't matter (or depends on pipeline neighbours) */     
    PJ_IO_UNITS_CLASSIC   = 1,  /* Scaled meters (right), projected system */     
    PJ_IO_UNITS_PROJECTED = 2,  /* Meters, projected system */    
    PJ_IO_UNITS_CARTESIAN = 3,  /* Meters, 3D cartesian system */    
    PJ_IO_UNITS_RADIANS   = 4,  /* Radians */    
    PJ_IO_UNITS_DEGREES   = 5,  /* Degrees */    

and `proj_angular_output` returns `true` exactly when the units are `PJ_IO_UNITS_RADIANS`. I wish I understood whether the `isGeographic` code

    case PJ_TYPE_GEOGRAPHIC_2D_CRS:
    case PJ_TYPE_GEOGRAPHIC_3D_CRS:
	    return true;

behaves as needed where the `ProjTransform` class may need to convert between degrees and radians. I will continue to investigate.

pkturner:
On the **georeferencing-fixes** branch, the only use of `isGeographic` in Mapper is to determine whether to convert coordinates between radians and degrees. Surprisingly, this is only done in the `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` case. Otherwise, apart from the `Georeferencing` test, `isGeographic` is not used at all!

In the PROJ source code for `cs2cs` I found the interesting function `instantiate_crs` which extracts from a CRS (1) whether it is geographic, (2) the conversion factor between the coordinate system's units and radians, and (3) whether it is latlong or longlat.

I am thinking of recommending to remove `isGeographic` from the public interface of `ProjTransform` and `Georeferencing`, and to change from testing `isGeographic` to testing how Mapper handles a geographic CRS that uses degrees as units.

dg0yt:
`isGeographic()` is already used by `TemplateImage`, to handle (certain) world files which are based on WGS 84. In fact, we would need to know the actual unit if we would want to support also radians for world files.

Independent of unit, `isGeographic()` is useful as an indicator for whether we must apply a projection or if we could use the coordinates as is, with a more simple transformation. This is why it is (sowewhat) relevant for dealing with geospatial vector data. We can transform the coordinates to the map's projected CRS in both cases. But for reproducible and fast transformations, it might be easier (and when the map is not georeferenced, it might be necessary) to use reproducible stereographic or orthograpic projections for the geographic case, and no coordinate transformation for the projected case.

Last not least, I hope that testing simpler behaviours like `isGeographic` contributes to getting more complex behaviours right. The complex behaviours need testing, too, but they cannot explain well how something went wrong.

With PROJ 4,  there was `pj_is_latlong()`. Following the [migration guide to PROJ 6](https://proj.org/development/migration.html#function-mapping-from-old-to-new-api), we shall use `proj_get_type()` instead. Now this turned out to be insufficient. Some of these cases are caught by the usage of `proj_angular_output` which is the replacement suggested by the [migration guide from PROJ 4 to PROJ 5](https://proj.org/development/migration.html#id2).

`instantiate_crs()` looks promising for doing `isGeographic()` right. Thanks for this hint! Once we solved the units, we are going to stumble over the axis order ;-)

pkturner:
> `isGeographic()` is already used by `TemplateImage`, to handle (certain) world files which are based on WGS 84. In fact, we would need to know the actual unit if we would want to support also radians for world files.

I noted that this PR removes the use of `isGeographic` from `TemplateImage`.

It seems reasonable to expect is that `Georeferencing::toProjectedCoords(LatLon)` and `toGeographicCoords(QPointF)` would use _degrees_ for the `LatLon` coordinates, and for the `QPointF` coordinates would use the units specified by the CRS. The CRS could be geographic (in radians or degrees) or projected (in meters, US survey foot, etc). @dg0yt Is that your understanding? If `Georeferencing` works that way, then the `TemplateImage` handling of world files would not be concerned with units — the template's `Georeferencing` would accept world file coordinates directly.

pkturner:
> when the map is not georeferenced, it might be necessary to use reproducible stereographic or orthographic projections for the geographic case, and no coordinate transformation for the projected case.

This is an interesting possibility for using `isGeographic`. On the one hand, a template with a projected CRS already has a map-like flatness. On the other hand its CRS may involve some distortion that is less than ideal for an orienteering map, such as enters into the difference between `template_scale_x` and `template_scale_y` in the `TemplateTransform` if the CRS is web Mercator `EPSG:3857`.

dg0yt:
> I noted that this PR removes the use of `isGeographic` from `TemplateImage`.

Oh, you are right.

dg0yt:
> The CRS could be geographic (in radians or degrees) or projected (in meters, US survey foot, etc).

For the actual map, the current assumption by editor components is that the projected CRS is in meters. I wouldn't mind relaxing that assumption in the future. But internally, `Map` operates with coordinates in millimeters (`MapCoord`). This is fixed, and it forms the base for certain interpretations such as the map scale.

For `Map` objects used as internal storage in `TemplateMap`/ `OgrTemplate`, I suppose that the unit doesn't have to be meters. The coordinates stored in the map are in millimeters, and they are transformed either via `Georeferencing` or via the template's transformation parameter in order to align the template on the map.
Still, it is important to know the intented scale when it comes to symbols that come with vector data.

dg0yt:
`instatiate_crs` yielded the key to what was wrong with `ProjTransform::isGeographic`, b2a070b.

I would suggest to squash the fixes to `ProjTransform::isGeographic` (no need to carry the invalid attempt from 2463714), revert the caching of is_geographic, but leave `Georeferencing::isGeographic` in the API for now.

pkturner:
> leave `Georeferencing::isGeographic` in the API for now

That's reasonable. The [b2a070b](https://github.com/OpenOrienteering/mapper/commit/b2a070bb070d6d261396a163c805b570a43df3ae) change is nice.

Issue #84, opened 2012-07-18
WMS (Web map service) support


dg0yt:
**puzzlepaint reported on Sourceforge [[tickets:#84](https://sourceforge.net/p/oorienteering/tickets/84)]:**

Add support to use templates downloaded from web map services. Suggested by Haavard Tveite.

bagage:
What would be the current status for this issue? Do you @dg0yt plan to work on it and/or have you any guidance about how to do it?

dg0yt:
My suggestion is to base WMS support on GDAL (cf. http://www.gdal.org/frmt_wms.html). However, getting a single image from a WMS might be fairly simple even without GDAL if you don't have to fiddle with different projections. Note that the proliferation of HTTPS-only services adds even more requirements to such efforts (cf. #814).

This is nothing I'm planning work to on very soon. Apart from higher-priority issues, there is a lot of code which needs a review and refactoring before I want to add new complex features.

bagage:
Well, using GDAL might be overkill but I think it makes sense. But do you know WMS servers what we could use in Mapper?

Because actually with legal restrictions I don't know any WMS server that could suit our need, eg. Google does not allow use of its data for mapmaking, OpenStreetMap is ODBL license (while orienteering maps usually aren't), I doubt Bing allows it… so do you know any one that would allow us without law infringement?

dg0yt:
There are regions and countries which have open geo data. Otherwise the principle is as with any other kind of template / base map one would use: The mapper (i.e. the user, not the software) has to find out the legal terms and to acquire the required permissions.

Symbian9:
@dg0yt, Did you have a look onto Marble sourcecode?
* https://marble.kde.org/
  * https://github.com/KDE/marble

About WMS in Marble
* https://api.kde.org/stable/kdeedu-apidocs/marble/html/classMarble_1_1WmsServerLayout.html
* https://userbase.kde.org/Marble/WizardMap
* https://techbase.kde.org/Marble/CustomMaps

bilbo2:
In many countries there are local WMS services provided by a government organization.
In many countries stuff made by government is considered "public work" or "official work" and effectively is in public domain. So, it can be used to draw maps, without even need to ask for special permission!

Quality varies, sometimes it is good enough to draw some rudimentary map straight from it.

[...]

PR #1816, opened 2020-12-23
WMS POC: Load single WMS tile at map reference point

dg0yt:
Proof-of-concept for using WMS, by opening XML files as described in
https://gdal.org/drivers/raster/wms.html and generated by
`gdal_translate`. Works with GDAL's OSM WMS xml file:
https://raw.githubusercontent.com/OSGeo/gdal/master/gdal/frmts/wms/frmt_wms_openstreetmap_tms.xml

Task: GH-84


dg0yt:
Unbenannt

Issue #1290, opened 2019-07-19
More RC-friendly file format


otoomet:
### Steps to reproduce
1. create a new file
2. add a few simple objects
3. save it as _.xmap_ file

### Actual behaviour
The file will be large, and apparently includes undo information, zoom level, view and such.  Could not find a description of the file format.

### Expected behaviour
Option to save only strictly _map data_ (and preserve the order and formatting of objects saved).  The undo/view information will interfere with RC systems, such as _git_ and create a lot of spurious differences even when the actual map is not changed.  

I would like to keep some of the maps under RC and keep the diffs as simple as possible.  Ideally there will also be a some sort of graphical diff-merge tool (see #741) on top of the text-diffs.

### Configuration
Mapper Version: 0.8.4
Operating System: ubuntu 16

dg0yt:
- The unstable version already creates additional linebreaks in the regular, compact *.omap* files, for better use in version control.
- The 'non-map' data is another story. At least undo history can be cleaned manually before saving for version control, via Edit > Clear undo/redo history. This should mitigate the most serious sub-issue.
- Side effects of color or symbol changes are probably the biggest show-stopper for this type of version control at the moment.

dg0yt:
FTR the request for XML format documentation is tracked in #182.

otoomet:
Thanks.

> Side effects of color or symbol changes are probably the biggest show-stopper for this type of version control at the moment.

I am not sure I understand what this means... 

As a side note, #1097 asks pretty much the opposite.

dg0yt:
When you add/remove a color, it will trigger changes in a number of symbols.
When you add/remove a symbol, or changes symbol order, it will trigger changes in a number of objects.

I don't see a conflict with #1097, because it will affect objects only when they are modified anyway. It addresses revision control at an object level, while `git` and Co. can only do revision control on a file level. The question is how to implement #1097 without adding too much extra data.

jmacura:
For the record: I was trying to write a conversion script from .omap to SVG, but found the file structure not really friendly to 3rd party tools. It contains the "noise" (undos & redos, map view), weirdly named elements ("barrier") and, despite being verbose in general, the coordinates itself are not atomic and hard to parse...

dg0yt:
[...]

PR #1820, opened 2020-12-24
Create efficient, order-preserving container for object tags

dg0yt:


dg0yt:
Task: GH-1290 (revision-control friendly file format)

Issue #1830, opened 2020-12-29
Cannot add scribble template


lpechacek:
Affects the pre-release build. v0.9.4 is fine.

### Steps to reproduce
1. Open a georeferenced map.
2. Tap-and-hold Paint on template button.
3. Select Add template.

### Actual behaviour
Mapper says that it cannot open the new scribble template.

### Expected behaviour
In v0.9.4 the template gets added and opened.

### Configuration
Mapper Version: master v20201227.4
Operating System: Android and openSUSE Tumbleweed

dg0yt:
Confirmed.

dg0yt:
Reproducible on the desktop.

PR #1832, opened 2020-12-29
PaintOnTemplateFeature: Fix setupTemplate()

dg0yt:
Indicate that template configuration is finished before loading the
template file, so that that function takes care of setting the state
to Loaded on success.
Complements aee1057f.
Fixes GH-1830 (Cannot add scribble template).


dg0yt:
Updated the commit message. 
I will merge this as soon as soon as CI passes.

lpechacek:
I'm in the forest now. I'll have a look at the fix when I'm back at home in the evening.

lpechacek (commented):
Ack on the patch. Scribble templates can be added again.

Issue #1835, opened 2020-12-30
Null pointer dereference in CombinedSymbol::hasRotatableFillPattern()


lpechacek:
Mapper segfaults during map editing. Please note that the combined symbol 508.2.1 has somewhat special definition.

### Steps to reproduce
1. Open the attached `reproducer.omap`.
2. Attempt to select one of the objects in view.

### Actual behaviour
The debug build terminates with:
```
/data/src/oomapper/src/core/symbols/combined_symbol.cpp:416:39: runtime error: member call on null pointer of type 'const struct Symbol'
/data/src/oomapper/src/core/symbols/combined_symbol.cpp:416:39: runtime error: member access within null pointer of type 'const struct Symbol'
Segmentation fault (core dumped)
```

### Expected behaviour
Application allows operations on the objects in question.

### Configuration
Mapper Version: master-v20201227.4-9-gca4f29466038
Operating System: openSUSE Tumbleweed

Archive containing `reproducer.omap`: [reproducer.zip](https://github.com/OpenOrienteering/mapper/files/5754016/reproducer.zip)

lpechacek:
The bug seems to boil down to mishnadling of combination symbols where one of its components is "none".

lpechacek:
Minimal reproducer:
```
<?xml version="1.0" encoding="UTF-8"?>
<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="9">
<notes></notes>
<georeferencing scale="15000"><projected_crs id="Local"/></georeferencing>
<colors count="0">
</colors>
<barrier version="6" required="0.6.0">
<symbols count="1">
[...]

PR #1836, opened 2020-12-30
CombinedSymbol: Fix hasRotatableFillPattern()

dg0yt:
Some patterns may be null. :facepalm: 
Amends 095bb634.
Resolves GH-1835 (Crash with certain objects).


lpechacek (approved):
Although I don't understand why the using-declarations for `begin` and `end` were dropped, the patch fixes the issue at hand. Thanks!

PR #1871, opened 2021-02-08
OpenStreetMap lookup of reference point to show marker

dl3sdo:
Map Georeferencing now shows a marker at the reference point in OpenStreetMap lookup of the position.
In addition the lookup operates with an improved precision.


dg0yt:
Tnx es 73.

PR #1872, opened 2021-02-16
Stylus only input

lpechacek:
This is a semi-RFC pull request. The first patch addresses issue 1368. The remaining three is an optional enhancement that can be implemented in a less intrusive way (read, by working around the design problem).

If you have a different idea about how to control the "disable touch input" feature or where to do the event filtering, let me know.


dg0yt:
Thanks for the PR.

I wonder how we can avoid the user being locked out after loosing the expensive stylus, or when accidently activating the feature while actually using a cheap stylus. Should the activation of the feature, or continued use after 10 min of inactivity, require a confirmation with the stylus?

lpechacek:
> I wonder how we can avoid the user being locked out after loosing the expensive stylus, or when accidently activating the feature while actually using a cheap stylus.

After losing the stylus the user can press the App Overview device button, or gesture, and remove Mapper from the list. That way the application terminates and the user can restart it, enable touch display input, and continue work. Map autosave minimizes damage in this case. In case of accidental activation before editing starts, the same procedure applies and no data loss occurs.

> Should the activation of the feature, or continued use after 10 min of inactivity, require a confirmation with the stylus?

The idea of confirming the activation of touch input filtering with a stylus sounds very good to me! After clicking the option in Settings a confirmation dialog would pop up where the OK button would ignore touch input. It requires more code but makes the application more user-friendly. Unless an alternate proposal appears, I'll implement this confirmation dialog.

mlerjen:
It is a good point to request the activation of the feature be confirmed by stylus. 
I though would restrain from a re-confirmation requests after an inactivity e.g. after every short break - several times a day. 
This because the scenario of just loosing a pen seems hardly realistic while a pen ceasing to function is not unusual. As continuing without a working stylus would be no option, carrying spare pens with oneself is to be considered a must. Thus in this context getting locked out due to lack of stylus would mean to have to go home anyway.😁 

Any chance to get a deactivation of the palm protection by a touch gesture?🤔😁

dg0yt:
> Any chance to get a deactivation of the palm protection by a touch gesture?

Maybe a combination of a long touch, and then drawing a certain pattern. Vikazimut has a 'Z' pattern to leave a course early.

src/gui/main_window.cpp
```diff
@@ -232,6 +232,24 @@ void MainWindow::setIgnoreTouch(bool on)
 	ignore_touch_input = on;	
 }
 
+bool MainWindow::setIgnoreTouchAndWarn(bool on)
+{
+	if (on && !ignoreTouch())
+	{
+		QMessageBox dialog(QMessageBox::Warning, tr("Warning"),
+		                          tr("Display touch input will be ignored in the editor."),
+		                          QMessageBox::Cancel | QMessageBox::Ok);
```
dg0yt, line 241: For ensuring stylus availability, maybe we just need an event filter on the OK button, sending warnings on touch-down, mouse-down, or key-down events, and otherwise dropping these events?

dg0yt (approved):

src/gui/main_window.cpp
lpechacek: I'm looking into it. However, I'm struggling with setting up the input event filtering on the `QPushButton` instance. For some reason, my code works on Linux but on Android, the OK button accepts the touch event. The current failing version of `setIgnoreTouchAndWarn()` is below. Still investigating.

```
bool MainWindow::setIgnoreTouchAndWarn(bool on)
{
        if (!on || ignoreTouch())
        {
                setIgnoreTouch(on);
                return true;
        }

        QMessageBox dialog(QMessageBox::Warning, tr("Warning"),
                           tr("The application will ignore display touch input beyond this point. Confirm that your alternate pointing device works by activating the Ok button."),
                           QMessageBox::Cancel | QMessageBox::Ok);
        dialog.setInformativeText(tr("If that does not work, go back to the editor settings and deactivate the \"Ignore display touch\" option."));
        dialog.setAttribute(Qt::WA_AcceptTouchEvents);
        dialog.raise();
        auto* ok_button = dialog.button(QMessageBox::Ok);
        ok_button->setAttribute(Qt::WA_AcceptTouchEvents);
        ok_button->installEventFilter(this);
        setIgnoreTouch(true);
        auto confirmed = dialog.exec() == QMessageBox::Ok;
        setIgnoreTouch(confirmed);

        return confirmed;
}
```

dg0yt: Does the event filter consume (accept) the touch events, or do they fall through as unhandled, being resent as mouse event or similar?

lpechacek: FTR, this code works as intended. It's based on plain `QDialog` and performs a complicated breakdance to reuse the event filter code. But it's at least a starting point.

Interesting observations:
1) The OK button receives focus even when it ignores the touch. The focus change is likely handled by the parent then.
2) Setting a filter on the parent dialog makes also the Back button inert to touch. It seems that the GUI event **filtering** works in the parent->child direction while the **event handling** is child->parent as hinted by `QObject::event()` docs. There is little explanation for these concepts or I've searched the wrong places.

```
bool MainWindow::setIgnoreTouchAndWarn(bool on)
{
	if (!on || ignoreTouch())
	{
		setIgnoreTouch(on);
		return true;
	}

	QDialog message(this);
	auto* layout = new QVBoxLayout();
	layout->addWidget(new QLabel(tr("Warning")));
	auto* text = new QLabel(tr("The application will ignore display touch input beyond this point. Confirm that your alternate pointing device works by activating the Ok button."));
	text->setWordWrap(true);
	layout->addWidget(text);
	auto* info = new QLabel(tr("If that does not work, go back to the editor settings and deactivate the \"Ignore display touch\" option."));
	info->setWordWrap(true);
	layout->addWidget(info);
	auto* button_layout = new QHBoxLayout();
	auto* button_back = new QPushButton(tr("Back"));
	button_layout->addWidget(button_back);
	button_layout->addStretch();
	auto* button_ok = new QPushButton(tr("OK"));
	button_layout->addWidget(button_ok);
	layout->addLayout(button_layout);
	message.setLayout(layout);
	connect(button_back, &QPushButton::clicked, &message, &QDialog::reject);
	qApp->removeEventFilter(this); // FIXME add activateEventFilter/deactivateEventFilter to MainWindow
	setIgnoreTouch(true);
	button_ok->setAttribute(Qt::WA_AcceptTouchEvents);
	button_ok->installEventFilter(this);
	connect(button_ok, &QPushButton::clicked, &message, &QDialog::accept);
	message.setWindowModality(Qt::WindowModal);
	auto confirmed = message.exec() == QDialog::Accepted;
	qApp->installEventFilter(this);
	QApplication::processEvents(QEventLoop::ExcludeUserInputEvents, 100 /* ms */);
	setIgnoreTouch(confirmed);
	
	return confirmed;
}
```

dg0yt:
I have a working local development build for my old tablet now. I will check the event filtering in the next days. 

Another idea after testing is to change the stylus confirmation dialog from Okay/Cancel to Okay/Continue with touch input.

dg0yt:
> I have a working local development build for my old tablet now. I will check the event filtering in the next days.

I see, event handling is a mine field, and Android is an extra level.

> Another idea after testing is to change the stylus confirmation dialog from Okay/Cancel to Okay/Continue with touch input.

This is implemented in my commit. Note that the Android "Back" key has the same effect as "Continue with touch input".

I also managed to add the input device hint when you touch the "OK" button. 

So, please review and test. If you are happy with the PR, go ahead and merge.

lpechacek:
> So, please review and test. If you are happy with the PR, go ahead and merge.

Many thanks for the patch, Kai. I like the new approach to user interaction and the implementation looks good to me.

I've cleaned up the series and dropped code that allowed going back to the home screen. The touch input filtering is now implemented by the first and last patch. I've left the member initialization patch in place as it guards against random memory corruption.

If there are no objections to the updated series, and CI results are green, I'll merge the change by tomorrow noon.

Changes in this iteration:
* Dropped `MapEditor: Prevent crash on object destruction` patch - not needed (yet).
* Renamed `setIgnoreTouchAndWarn()` to `warnAndSetIgnoreTouch()`, made it return no value and updated the in-code description.
* Added commit message to the last patch.

dg0yt:
Works very well for me.
The new dialog appears after the import warning dialog. Like that dialog, it hides the messages which indicate the asynchronous loading of templates, but it doesn't suspend this process.
:+1:

lpechacek:
Merged. Thanks to all involved for their time and contributions!

Issue #1875, opened 2021-02-22
Passable rock face line width wrong in ISSprOM 2019


tdobra:
### Steps to reproduce
1. Use any passable rock face symbol (202) in the ISSprOM 2019 symbol set.

### Actual behaviour
Line width is 0.25 mm.

### Expected behaviour
Line width should be 0.3 mm - from specification.

### Configuration
Mapper Version: 0.9.2
Operating System: Windows and MacOS

PR #1876, opened 2021-02-22
Change passable rock face width to 0.3 mm

tdobra:
Resolves issue #1875


tdobra:
I would like some help please as to why this is failing build checks. There is clearly something I don't understand about how symbol sets are stored/implemented in Mapper.

dg0yt:
> I would like some help please as to why this is failing build checks. There is clearly something I don't understand about how symbol sets are stored/implemented in Mapper.

Normally, the scaled files are generated from the src files (by the `SymbolSetTool` "test"). The failing check indicates that the tool generates different files than your manual editing. If you don't want to build Mapper from source and run SymbolSetTool, give me a fews days to do the integration.

tdobra:
Thanks Kai, I'll leave the integration to you. As you can see, I initially edited the OMAP at 4000, then later edited the XMAP in src/ thinking it was the cause of the failed check. As far as I am aware, these two sets of edits are the same, so you can do whatever you need to do with them.

dg0yt:
**202 Passable rock face, small (rounded ends, no tags)* has a change of start offset/end offset in the omap file but not in the src xmap file.

tdobra:
Sorry, not updating the offset in the src XMAP is an error on my part; I thought I had remembered all my edits. Thank you for checking and fixing it. Next time, I think you said I should only edit the src XMAP and autogenerate the scaled OMAP. However, for my education, is this error the only reason why my pull request failed the checks?

dg0yt:
>  Next time, I think you said I should only edit the src XMAP and autogenerate the scaled OMAP. 

Ideally, yes. If auto-generation is a barrier, just change the XMAP. This format is also more suitable for review in source form.

> However, for my education, is this error the only reason why my pull request failed the checks?

Yes. There is only CI configuration that includes the failing step, but this is enough to catch the inconsistency. Note that running `SymbolSetTool` has more effects, such as updating translation sources for symbol sets.

Thank you very much for the contribution. It really helps to move things forward.

PR #1879, opened 2021-02-23
Update CI

dg0yt:


dg0yt:
Oops, the encoding change is obsolete already. The document now states:

> NOTE: This was previously designed to use %25 instead of %AZP25 as the escape sequence. We decided to go with %AZP25 instead since %25 was used somewhat often because of its role in url encoding. Some agents may continue to emit warnings for %25 as this change rolls out (or if you haven't updated to the most recent agent). These warnings are safe to ignore.

And `%25` is what we are using.

Issue #1883, opened 2021-02-25
Vikazimut compatibility


dg0yt:
_Originally posted by @Eric-P-Kmz in https://github.com/OpenOrienteering/mapper/issues/1793#issuecomment-782428545 :_
> Hi,
>
> I participate in the development of the free Vikazimut application. Is it possible to do a kmz export of the same format as that of OCAD in OOM (with a single image in the export)?
> We use the OCAD export format to georeference the maps to import into the app. The division into several images of the map is not compatible with our import.

- [x] Export KMZ with single raster image file
- [x] Export single course in IOF format

dg0yt:
@eric-p-kmz Here is another ZIP file with a simple course in IOF Data Standard 3.0 format.
[course.zip](https://github.com/OpenOrienteering/mapper/files/6047683/course.zip)

Eric-P-Kmz:
Hi,

We have continued to improve the link between OOM and Vikazimut for creating routes from OOM.
We've managed the differences between the IOF and OOM formats in the XML file, but the map scale is missing from your XML.
Is it possible to add this tag to the XML?
<RaceCourseData>
  <Map>
      <Scale>10000</Scale>
  </Map>
...
</RaceCourseData>

- Second request, would it be possible to create multiple races in one XML?

- If you can make it easier to create the definition table, it would remove a big drawback to creating courses under OOM.

dl3sdo:
Hi @Eric-P-Kmz,
thank you for your ideas to improve XML export. I will analyze them and eventually create a PR.
Since this issue is closed I propose to continue discussion in a new issue, so please open a new one with your input above.
Thank you and Best Regards, Matthias

PR #1884, opened 2021-02-25
Add tile size configuration to KMZ export

dg0yt:
...including single-image option.

Task: GH-1883 (Vikazimut compatibility)

### DIFF
--- a/src/gdal/kmz_groundoverlay_export.cpp
+++ b/src/gdal/kmz_groundoverlay_export.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2020 Kai Pastor
+ *    Copyright 2020-2021 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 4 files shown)


dg0yt:
Unbenannt

lpechacek:
Apologies for off-topic upfront. I know virtually nothing about KMZ tiles, but working with the scribble layers, I dreamt about an adaptive scribble layer that would automatically add tiles as I paint instead of the current manual method where I add tiles when I need them. The scribble would be a single georeferenced layer then. Would KMZ be the direction to explore, or is there a better option, please?

dg0yt:
@lpechacek I thought about this, too. It is clear that we need a tiled scribble layer. But I don't think the KMZ approach fits well for saving: it means updating a raster image file inside a big ZIP file via GDAL. Is this safe when an Android device is under pressure? I cannot use QSaveFile here. 
But the tiled scribble layer might benefit from how I enabled the loading off tiled KMZ files as templates. And WMS support will need tiles, too.

mlerjen:
Loading and creating maptiles on demand from wms, xyz- datastructures mbtiles and alike would be huge.

lpechacek:
> But I don't think the KMZ approach fits well for saving: it means updating a raster image file inside a big ZIP file via GDAL.

Good point. I missed the ZIP part. It disqualifies the idea for the moment.

> But the tiled scribble layer might benefit from how I enabled the loading off tiled KMZ files as templates. And WMS support will need tiles, too.

Good to know. This scribbling enhancement is low on my list but I'm still looking around if there is low hanging fruits. Thanks!

dg0yt:
> > But I don't think the KMZ approach fits well for saving: it means updating a raster image file inside a big ZIP file via GDAL.
> 
> Good point. I missed the ZIP part. It disqualifies the idea for the moment.

I have to add that KML + raster tiles could also be used without zipping/KMZ.

> > But the tiled scribble layer might benefit from how I enabled the loading off tiled KMZ files as templates. And WMS support will need tiles, too.
> 
> Good to know. This scribbling enhancement is low on my list but I'm still looking around if there is low hanging fruits. Thanks!

That's why I do this KMZ stuff: it is low hanging fruits, with benefits for users(Google Earth, MapRun, Vikazimut) and for us  (exploring tile handling).

Issue #1906, opened 2021-03-12
Manual installed in a wrong place


sikmir:
### Steps to reproduce
```sh
$ cmake \
    -DMapper_MANUAL_QTHELP:BOOL=OFF \
    -DCMAKE_INSTALL_DOCDIR=/home/sikmir/test/share/doc/Mapper \
    -DCMAKE_INSTALL_MANDIR=/home/sikmir/test/share/man \
    -DCMAKE_INSTALL_PREFIX=/home/sikmir/test ..
$ make
$ make install
```

### Actual behaviour
```sh
$ tree /home/sikmir/test
/home/sikmir/test
├── bin
│   └── Mapper
├── home
│   └── sikmir
│       └── test
│           └── share
│               └── doc
│                   └── openorienteering-mapper
│                       └── manual
...
└── share
...
    ├── man
    │   └── man1
    │       └── Mapper.1
...
```

### Expected behaviour
It works properly with Mapper 0.9.4:
```sh
$ tree /home/sikmir/test
/home/sikmir/test
├── bin
│   └── Mapper
└── share
...
    ├── doc
    │   └── openorienteering-mapper
...
    │       └── manual
...
    ├── man
    │   └── man1
    │       └── Mapper.1
...
```

### RCA

I guess `${CMAKE_INSTALL_PREFIX}` not needed here:
https://github.com/OpenOrienteering/mapper/blob/9e874fe7d9d1b19e7560a2ed63be9a5da5cd0908/doc/manual/install-html.cmake.in#L26

### Configuration
Mapper Version: 0.9.5
Operating System: Linux

dg0yt:
There was a change to properly use the `GNUInstallDirs` module from CMake.

IIUC, `install-html.cmake.in` is only used when building with `Mapper_MANUAL_QTHELP:BOOL=OFF`. However, it didn't break the Android build. Hm...

sikmir:
Yes, I use `Mapper_MANUAL_QTHELP:BOOL=OFF`.

PR #1907, opened 2021-03-13
doc/manual: Fix html pages installation

dg0yt:
Ensure an absolute path, but don't blindly prepend
CMAKE_INSTALL_PREFIX.
Fixes GH-1906 (manual pages installed in wrong place).


sikmir (approved):
It works now.

PR #1911, opened 2021-03-15
OcdFileExport: Export OCD template opacity

lpechacek:
A small nit spotted while working with .ocd export. This is "obviously correct", in my opinion, and I'll merge the change as soon as CI tests finish.


dl3sdo:
Hi Libor, I noticed that you didn't put the new line in an else path, i.e. the output will happen even if the warning above would be triggered. Best Regards, Matthias

lpechacek:
> I noticed that you didn't put the new line in an else path, i.e. the output will happen even if the warning above would be triggered.

@dl3sdo: Thanks for the remark, Matthias, but I'm not completely sure that I understand the meaning. The "dim" template parameter is independent of the template map's spatial transformation. That is, even if the spatial transformation is erroneous, information about the template dimming is valid. That is why I add the `d` code regardless of the warning.

dl3sdo:
@lpechacek: my fault, I didn't really check the meaning of the d code element in conjunction with the code above.

PR #1922, opened 2021-03-24
MapEditorController: Rename Cutout and Cut Away tools

lpechacek:
The rename is on the user interface side and within the map editor code and the tool icons have been refreshed for easier navigation in the UI. The rename makes a lot of noise but there are no behavior changes. The underlying converged tool implementation keeps its original name.

I'd be glad for a quick review. Thanks!


src/gui/map/map_editor.cpp
```diff
@@ -1066,8 +1066,8 @@ void MapEditorController::createActions()
...
 	convert_to_curves_act = newAction("converttocurves", tr("Convert to curves"), this, SLOT(convertToCurvesClicked()), "tool-convert-to-curves.png", QString{}, "toolbars.html#convert_to_curves");
 	simplify_path_act = newAction("simplify", tr("Simplify path"), this, SLOT(simplifyPathClicked()), "tool-simplify-path.png", QString{}, "toolbars.html#simplify_path");
-	cutout_physical_act = newToolAction("cutoutphysical", tr("Cutout"), this, SLOT(cutoutPhysicalClicked()), "tool-cutout-physical.png", QString{}, "toolbars.html#cutout_physical");
-	cutaway_physical_act = newToolAction("cutawayphysical", tr("Cut away"), this, SLOT(cutawayPhysicalClicked()), "tool-cutout-physical-inner.png", QString{}, "toolbars.html#cutaway_physical");
+	clip_area_act = newToolAction("cliparea", tr("Clip area (formerly Cutout)"), this, SLOT(clipAreaClicked()), "tool-clip.png", QString{}, "toolbars.html#cutout_physical");
+	erase_area_act = newToolAction("erasearea", tr("Erase area (formerly Cut away)"), this, SLOT(eraseAreaClicked()), "tool-erase.png", QString{}, "toolbars.html#cutaway_physical");
```
dg0yt, line 1070: I think this somewhat inconvenient for translators:
- First, they will have to translate "Clip area (formerly Cutout)".
- Later, they will have to translate "Clip area".

So for translation, we should use a `tr("Clip area")` IMO. Then the question is how to implement the transition in the UI. 
- We might try something like `tr("%1 (formerly: %2").arg(tr("Clip area"), tr("Cutout"))`. This additional string could be reused for future transitions. However, it might need an extra comment to translators.
- Alternatively, we might drop the old names completely, and just document the change.

dg0yt (commented):
Basically :+1:, very much appreciated. I have only comment with regard to translation.

With regard to the PDF manual error, I think I would work around it using another (temporary) text substitution rule in markdown preprocessing for LaTeX. After merging this PR here, I can push the new icons to the mapper-manual repository, update the pages there, and push the updated pages as usual back into mapper.  I would offer to prepare a commit with these rules.

src/gui/map/map_editor.cpp
lpechacek: Thanks, Kai, for the remark. I thought about the translators too and considered the suffix removal a low-effort task and worth the hassle. However, on second thought, I think we can omit the parenthesized text altogether. The new name always comes with the well-known tool icon, so few people will be surprised. Let's go with the new names only and save contributors time.

Another aspect is that a graphical application changes and evolves. It's not an industrial software component that would need compatibility hacks. I'm happy for the opportunity to leave my professional niche and exercise new mental patterns. :)

lpechacek:
Ad the manual page update, it was on my list too. I didn't know that the change will cause build failure though. I'll be happy for any help here. Thanks for the offer!

While we are at it, shall we also change the anchor names to match the action identifiers — i.e. `#cutaway_physical` and `#cutout_physical` to `#erase_area` + `#clip_area` — both in the code and in the manual?

src/gui/map/map_editor.cpp
dg0yt: > Let's go with the new names only and save contributors time.

Agreed.

dg0yt:
> While we are at it, shall we also change the anchor names to match the action identifiers — i.e. `#cutaway_physical` and `#cutout_physical` to `#erase_area` + `#clip_area` — both in the code and in the manual?

Yes. I'll leave the code changes to you. Don't wait, I don't mind a short period of inconsistency.

lpechacek:
Changes in this iteration:
* Dropped the parenthesized text
* Aligned the hypertext anchor names with the action names
* Reworded the first patch commit message

I've aligned the patch author dates with the commit dates. At least that is my understanding of how rebases and patch set reworks should be done in this project. I'll merge the change as soon as CI completes. Thanks!

PR #1923, opened 2021-03-25
Added a Info.plist file for mac os so that the app doesn't look blurry af

FelixFrog:
I just added a Info.plist file to make the .app bundle able to run well on HiDPi displays and not blurry. I tried and it compiles fine
Screenshot 2021-03-25 at 10 19 37
Before
Screenshot 2021-03-25 at 10 19 18
After


packaging/CMakeLists.txt
```diff
@@ -178,6 +178,7 @@ elseif(APPLE)
 	set_target_properties(Mapper PROPERTIES
 	  MACOSX_BUNDLE_INFO_STRING          "${CPACK_PACKAGE_DESCRIPTION_SUMMARY}"
 	  MACOSX_BUNDLE_ICON_FILE            "Mapper.icns"
+    MACOSX_BUNDLE_INFO_PLIST           "${PROJECT_SOURCE_DIR}/packaging/apple/Info.plist"
```
lpechacek, line 181: Thanks, Francesco, for the code contribution! I'll leave this pull request review to Kai as I know absolutely nothing about Mac. However, the whitespace error is glaring, and cannot help myself not to point it out. ;-)

The line should start with a tab and two spaces for proper alignment. Below is an illustration how a properly formatted line would look like in Qt Creator with whitespace visualization on.
image

dg0yt:
Thank you very much for the PR. It is good to see this getting resolved. I lack a Retina Mac, so I'm unable to test myself and need your help to proceed.

- As you may see from the checks in this pull request, the macOS build fails now. The problem is not the compile step, but the `package` step.
- I suspect that by providing a custom Info.plist, we loose all the other properties which CMake would insert in the template it provides. Cf. https://github.com/Kitware/CMake/blob/master/Modules/MacOSXBundleInfo.plist.in. 
(Coincidentally, the last commit to that file reverted ... guess what: https://github.com/Kitware/CMake/commit/286c75f7f034c5fdcd43bcb755da74d09c809642)

So we either need to provide a complete file (possibly as a template, for use with `configure_file`), or we try to use `plutil` to add this property after the build (cf. https://stackoverflow.com/a/60218931).

In addition, I would kindly ask you verify doesn't break:
- The right scale at 100% zoom.
- The right scaling in print preview and printing.

Last not least, I want to add a reference to the Qt documentation:
https://doc.qt.io/qt-5.12/highdpi.html#macos-and-ios 
It raises another point which deserves attention/testing:
"macOS ... may apply further virtualization such that device pixels do not correspond 1:1 to display pixels ... [when] configured with 'display scaling' enabled."

I really hope you can help us with testing. I can provide help with implementing the changes.

FelixFrog:
First and foremost, I fixed the indenting issue. Secondly you were right: and as the documentation stated specifying a custom `Info.plist` file didn't take into account the other variables. Fortunately cmake actually runs `configure_file` on the `.plist.in` file, so it was only a matter of inserting the variables directly into it.
As far as scaling goes, there doesn't seem to be any issue, since it only affects menus and UI parts, not the map itself.
Finally, what the Qt docs actually refer to with "may apply further virtualization such that device pixels do not correspond 1:1 to display pixels" is the sistem-wide monitor scaling that is applied when in mirroring mode with another display. So this is nothing to worry about and is completely indipendent from OOM.
If you need further help for testing don't hesitate to ask me.

dg0yt:
I had a try on my MBA today, and I can notice a positive difference even without Retina display.

However, I would like to not duplicate too much from what CMake is to do. So I created another PR, #1933, which takes the `plutil` approach. Please have a look if it works for you.

FelixFrog:
For me it compiles fine without problems and it fixes the issue!

dg0yt:
Thanks!

Issue #1896, opened 2021-03-08
Ocad export/ import - real word coordinates angle/ map north


ollesmaps:
### Steps to reproduce
1. Import into OOM ocd (real word coordinates angle has many digits 4,0821 / it is possible to have in Ocad up to 8 digits after comma).
2. Export the same file from OOM into ocd
3. The real word coordinates angle in Ocad is now 4,08 (maximum 2 digits after comma).

### Expected behaviour
The angle shall be the same

### Configuration
Mapper Version: 0.9.3
Operating System: Win, Android

lpechacek:
There is no rounding applied to the number in the OCD import and export path. Would you mind sharing a demo file for debugging, please?

By the way, did you use the map georeferencing dialog with this map? I can imagine that it might be the source of the rounding.

ollesmaps:
I did not modify any numbers. Just plain import/ export.
Sample attached, numbered in order of exports:
[kost_deklinace_oc12_2.zip](https://github.com/OpenOrienteering/mapper/files/6130215/kost_deklinace_oc12_2.zip)

dg0yt:
Class `Georeferencing` applies `roundDeclination()` to the arguments given to the setters for declination/grivation, and OCD import must go through these setters.

https://github.com/OpenOrienteering/mapper/blob/9e874fe7d9d1b19e7560a2ed63be9a5da5cd0908/src/core/georeferencing.cpp#L841-L853

pkturner:
While I agree that it would sometimes be helpful if Mapper preserved all the digits provided by OCD, there are good reasons for rounding to two decimal places.

- Avoid complexity in the Georeferencing dialog.
- Give the user control over all the decimal digits that Mapper respects.

Perhaps Mapper could change its rounding of grivation to be 8 places, while still rounding declination to 2.

ollesmaps:
Is this rather?
- rounding
- or automatic rotation to that new degree?

In Ocad it triggers this:
![angle](https://user-images.githubusercontent.com/20535899/111062582-0e0d6e00-84aa-11eb-8b08-967f703d6ab6.jpg)

lpechacek:
> Is this rather?
> 
>     * rounding

Yes, it's pure rounding of the numeric value. There is no rotation applied to match the rounded value. The OCAD warning would be a reason to rectify the export issue. We should strive for exporting the same information bits that we receive where possible.

[...]

PR #1925, opened 2021-03-26
Increase precision of grivation for better OCAD import/export

pkturner:
This implements what was discussed for #1896, increasing the precision with which Mapper represents grivation, from 2 decimal places to 8.

The basic change is an increase in the number of decimal places when grivation is set, and when it is saved to XML or OCD. That change opens questions of "When should the precision present in grivation be respected, or taken advantage of?" In this PR, some georeferencing actions which affect grivation are revised for the added precision.

Most of the attention to enhanced precision occurs in the Georeferencing dialog. If the user changes the georeferencing CRS (while grivation is unlocked) keeping "Geographic coordinates", the grivation is changed precisely, based on the change in convergence.  If the user changes the reference point it is handled the same way.

Beyond that, if `Map::rotateMap` is called with the "adjust declination" option, the grivation is rotated by the given angle, rather than being set from the less precise declination.

A more pedestrian approach to this enhancement would be to maintain and save 8 decimal places of precision for both the declination and grivation, while exposing just 2 decimal places to the user. The additional precision would need to be saved in the XML file format, even for declination.

I was not content to implement the enhancement that way. If the Georeferencing state includes
  -  the declination rounded to 2 places
  -  the grivation rounded to 8 places
  -  the convergence (not rounded)

then that should be enough. In fact, we don't _want_ the declination to have more precision. So I discovered and made use of a nifty way to extract a precise declination from those properties of the Georeferencing state. This is the `Georeferencing::getDeclinationSetting()` function.


lpechacek:
Thanks, Scott, for your interest in my code review! I know little about the georeferencing corner of Mapper code but I'll dive deeper into it in the coming days to verify the code changes.

I've got a handful of comments. Some of them are optional enhancements. I'll tag those accordingly.

src/core/georeferencing.h
```diff
@@ -137,39 +137,57 @@ friend QDebug operator<<(QDebug dbg, const Georeferencing& georef);
 	
 	
 	/**
-	 * @brief Returns the precision of the grid scale factor.
+	 * @brief Returns the precision of the combined/auxiliary scale factors.
```
lpechacek, line 140: The comment updates seem unrelated to the number precision topic. If that is the case, I'd like to see them in a separate patch.

src/core/georeferencing.h
```diff
@@ -717,6 +735,20 @@ double Georeferencing::roundDeclination(double value)
...
+	// This must match the implementation in grivationRound().
+	return 8u;
+}
+
+inline
+double Georeferencing::roundGrivation(double value)
+{
+	// This must match the implementation in grivationPrecision().
+	return floor(value*1E8+0.5)/1E8;
+}
```
lpechacek, line 741: (optional code enhancement)
I like that the grivation rounding implementation is coherent with the declination rounding. What stands out in both places is the open possibility for inconsistency reflected in the two comments.

My proposal would be:
```suggestion
inline
constexpr unsigned int Georeferencing::grivationPrecision()
{
	return 8u;
}

inline
double Georeferencing::roundGrivation(double value)
{
	constexpr auto multiplier = pow(10, grivationPrecision());
	return floor(value*multiplier+0.5)/multiplier;
}
```
The same pattern would apply to declination rounding. If that weren't a header file, a common rounding function would be beneficial. At this place, the new symbol would pollute the namespace, though.

test/georeferencing_t.cpp
```diff
@@ -299,6 +299,58 @@ void GeoreferencingTest::testGridScaleFactor()
 		QCOMPARE(georef.toProjectedCoords(map_coord), sw);
 }
 
+void GeoreferencingTest::testGrivationAffected()
```
lpechacek, line 302: For some reason, if I check out commit with this test, `georeferencing_t` fails. Are all the necessary infrastructure changes in place for the test to succeed at this point in the patch series?

test/georeferencing_t.cpp
```diff
@@ -342,9 +342,10 @@ void GeoreferencingTest::testGrivationAffected()
 	// Change CRS from UTM 32 to EPSG:3857.
 	// Convergence of the latter is 0, so the change in grivation should equal
 	// the present convergence.
-	georef.setProjectedCRS(QString::fromLatin1(QTest::currentDataTag()), epsg3857_spec);
+	auto const declination_setting = georef.getDeclinationSetting();
+	georef.setProjectedCRS(QString::fromLatin1(QTest::currentDataTag()), epsg3857_spec, std::vector<QString>(), false);
 	georef.setGeographicRefPoint(latlon, true, false); // Fix up scale factor, not grivation.
-	georef.updateGrivation(); // Fix up grivation.
```
lpechacek: As to my understanding, this was the last use of `updateGrivation()`. Could this patch also contain the removal of the function from the code?

src/core/map.cpp
```diff
@@ -581,7 +581,7 @@ void Map::rotateMap(double rotation, const MapCoord& center, bool adjust_georefe
 	if (adjust_declination)
```
lpechacek, line 581: GitHub does have the function for comments on the commit message. Is the verb in the correct form in "... when **adjust** declination?"

pkturner:
Libor, thank you for looking over this PR. I requested you due to the interest you showed in #1925 (precision of grivation in OCAD import/export).

While I am able to respond to all of today's comments, I apologize in advance that I will shortly be unavailable. I will not be able to help put the finishing polish on this PR until May.

src/core/georeferencing.h
pkturner: You're correct. I will rebase and split off this change.

src/core/georeferencing.h
pkturner: I like your method of eliminating the need for the comment, though I don't trust `pow` to return an integral value. 

`constexpr auto multiplier = floor(pow(10, grivationPrecision()) +0.5)`

would work.

test/georeferencing_t.cpp
pkturner: That removal is in the next commit 811e4615, which removes `initDeclination()` as well as `updateGrivation()`. I like removing both together, as the two methods are twins. Would you squash the two patches?

src/core/map.cpp
pkturner: You caught me there! It should be " ... when **adjusting** declination". I left it incorrect for fear of the commit message being judged too long. For me, every commit message requires an effort at compression.

I'll fix it.

src/core/georeferencing.h
dg0yt: Please note that `pow` isn't `constexpr` by the standard. That's why I used the comment in `roundDeclination`.
https://en.cppreference.com/w/cpp/numeric/math/pow 

A compile-time solution for `int` is possible with templates (e.g. in an anonymous namespace early in the cpp file):
~~~
template<int N> constexpr int power(int base)
{
	return base * power<N - 1>(base);
}

template<> constexpr int power<0>(int)
{
	return 1;
}

static_assert(power<0>(10) == 1, "10^0");
static_assert(power<3>(10) == 1000, "10^3");
~~~

pkturner: I was leaning toward keeping the PR as is, thinking that the template adds complexity that is not justified by the small risk of the precision and rounding methods getting out of alignment.

However, the `power()` template makes it possible to eliminate the literal `1E6` that appears in the definition of 'getDeclinationSetting()`. That literal is another value which needs to be kept coordinated with the precisions, and which I had not taken the trouble to comment because it would need to be mentioned in 3 places. So I will add the template suggested by @dg0yt in order to statically verify the relationships that needed comments. I'll push a revision to the branch with that change.

It was necessary to introduce protected methods in the `Georeferencing` class to return the "multiplier" contstants that are used by the `roundXxxxxx()` methods, in order to express `static_assert`s for the relationships.

src/core/georeferencing.h
pkturner: Split off 20afafc.

src/core/georeferencing.h
pkturner: Implemented the template in f530fcf, applying it to `scaleFactorPrecision()` and `declinationPrecision()`. The comments indicating how the implementations must match are replaced with a `static_assert` in `georeferencing.cpp`.

src/core/map.cpp
pkturner: Fixed with 4417fe7.

test/georeferencing_t.cpp
pkturner: No. Perhaps the Mapper project has a convention that a test should not be introduced into the code until it passes. I have been a little influenced by the idea of "Write the tests before you write the code."

In this case, the test failure demonstrates how the code used by the `GeoreferencingDialog` fails to adjust grivation accurately when the user changes the CRS.

lpechacek: The conventions are created on the go. Don't get stressed about the surprises. :)

My comment was motivated by code "bisectability" (git bisect). That means the code must compile after every change (i.e. after every patch in the series must take the code base into a consistent state). I personally extend it to the tests too - i.e. tests must pass after every change. I think that this approach makes bug hunting easier.

The "write test, then code" is certainly beneficial and I use it too. However, the order should IHMO be reversed for the version control system - make the change, then test its validity.

test/georeferencing_t.cpp
lpechacek: Thanks for the confirmation. That is what I had in mind. The method removal is logically coupled with the removal of its uses.

src/core/georeferencing.h
lpechacek: Good point about the non-standard `constexpr` behavior of `pow`., Kai.

I like the static checks in `georeferencing.cpp` as it spots the potential discrepancy at compile time, Scott. I cannot tell, however, how Kai will approach the change.

src/core/georeferencing.h
```diff
@@ -696,25 +712,34 @@ constexpr unsigned int Georeferencing::scaleFactorPrecision()
 	return 6u;
 }
 
+inline
+constexpr unsigned int Georeferencing::scaleFactorMultiplier()
+{
+	return 1000000.0;
```
lpechacek, line 735: `return 1E6;` might be better aligned to the next patch which adds `1E8` for grivation precision.

src/core/georeferencing.h
```diff
@@ -696,25 +712,34 @@ constexpr unsigned int Georeferencing::scaleFactorPrecision()
...
 constexpr unsigned int Georeferencing::declinationPrecision()
 {
-	// This must match the implementation in declinationRound().
 	return 2u;
 }
 
+inline
+constexpr unsigned int Georeferencing::declinationMultiplier()
+{
+	return 100.0;
```
lpechacek, line 753: Possibly `1E2` in place of the `100.0` to align the style with the next patch.

src/core/georeferencing.cpp
```diff
@@ -859,6 +859,12 @@ void Georeferencing::setScaleFactors(double combined_scale_factor, double auxili
 	}
 }
 
+double Georeferencing::getDeclinationSetting() const
+{
+	constexpr auto precision_ratio = power<grivationPrecision() - declinationPrecision()>(10);
+	return ((grivation + convergence)*precision_ratio + declination)/(precision_ratio + 1);
```
lpechacek, line 865: A remark. I don't know the required math skills level to understand this code, and  I spent almost an hour trying to understand how the above calculation works. I may be that I'm not educated enough to get the meaning, or I'm looking at it from the wrong side.
1) I never studied the internals of `Georeferencing`. As to my understanding, cartographers use positive declination values for a shift to the right when looking towards the north. Same for meridian convergence - positive right, negative is left. In my area, declination is about 4°, and meridian convergence of the popular local projection is about 7°. Grivation is a sum of the two numbers, which is 4°+7°=11° and it's aligned with the numbers produced by [the tool](https://mapy.orientacnisporty.cz/en/convergenceclass ) we are using in my country. However, `Georeferencing` uses the opposite sign for convergence. That means ` grivation + convergence` is (counterintuitively) declination. When "positive means rotation to the right - always" would hold true, the declination calculation would be `grivation - convergence`. In other words, it took me some tens of minutes to understand that the calculation in `getDeclinationSetting()` is, in fact, "(declination*1000000+declination)/1000001".
2)  I don't get why the precision ratio is 10^6 and why the divisor contains "+1".
3) The "setting" in the method name associates with some "mode" or "flags" for me. I don't expect a method with such a name to return a number with enhanced precision or some combined figure.

If you decide to address my remarks, please do so through the code and commit message rather than comments in this pull request. Think about future contributors who will fix and enhance this code. In my opinion, they deserve to understand it without excessive effort.

lpechacek (commented):
Overall, I like that the patchset addresses the grivation precision is on .ocd export! It also extends the corresponding tests, which is great. :+1:

My biggest worry is my inability to understand the `getDeclinationSetting()`. I cannot tell if that part of the code is correct as I wasn't able to get through the basics. That might be a sign of a bug in the code, insufficient documentation, missing skills on my side, or a combination of the three.

The rest are stylistic issues whose removal will better align the new code with the existing parts. Kai may provide more feedback in this area as he's, in my opinion, far more knowledgeable on that topic.

Thanks for the time and effort you have invested in this enhancement so far!

test/georeferencing_t.cpp
dg0yt: I also try to introduce the test first, but explicitly mark the expected failures. This proves the existence of the bug. The bug fix is forced to remove the expected failure, which underlines the desired effect.

pkturner: I will change this to mark the expected failures.

test/georeferencing_t.cpp
pkturner: I will perform the squash.

src/core/georeferencing.h
pkturner: I plan to make this change.

src/core/georeferencing.h
pkturner: I expect to make this change.

src/core/georeferencing.cpp
pkturner: How to combine convergence with grivation is one of those conventions that could have gone either way, and intuition doesn't work for me, either. I'm often referring to the code in `setDeclination()` to check whether to add or subtract. An explanation of Mapper's convention can be found via the manual's **Georeferencing: Further Reading** section which refers to [Larry Simons (2005): Magnetic Declination & Grid Convergence](http://www.threelittlemaids.co.uk/magdec/explain.html).

pkturner: @lpechacek Thanks for your helpful remarks after the type of effort you put into it. I plan to address your 3 remarks in code changes (and comments in the code). The alternative name which comes to mind is `getPreciseDeclination()`. I hope that will make sense to our developers.

pkturner:
As I said, it will be some weeks before I am available again to develop the code. Thanks, @lpechacek for the review. The `getDeclinationSetting()` method can be made more comprehensible, and I look forward to resuming the work before too long.

src/core/georeferencing.cpp
lpechacek: Thanks, Scott, for the pointer. It is far more understandable than anything I read before. I spent a few hours today studying the peculiarities of the national grid and the tool I referenced before. It turned out that the national grid has negative convergence over all the area. However, the tool shows positive converegence figure which drove me into the weeds. It means that, point 1) of my comment is invalid and I'll mark it as such. I appreciate your patience. :)

lpechacek: As a side note, my research resulted in the discovery of the following document: [_OGP Geomatics guidance note 21 - Grid convergence_](https://ge0mlib.ru/papers/Guide/IOGP/373-21.pdf). The document mentions (defines?) two different conventions for the meridian convergence figure. One is called "Gauss-Bomford convention" and the other is "Survey convention". They differ in the sign and countries which use them. We may want to document that Mapper is using the Gauss-Bomford convention like [ArcGIS did](https://developers.arcgis.com/net/uwp/api-reference/html/M_Esri_ArcGISRuntime_Geometry_SpatialReference_GetConvergenceAngle.htm).

pkturner:
I'm back, with no more lengthy vacations planned.

I've revised the PR as discussed above. Most significant is changing the name of `getDeclinationSetting()` to `getPreciseDeclination()`, and revising that method's code to be more descriptive and better motivated (I hope).

Issue #1928, opened 2021-03-30
Font names invisible due to excessive font icon size


biker2s:
I found this minor issue in previous versions as well. I solve it by opening the file with 0.9.2 version to change it and then I work again with the latest one. The problem exists from 0.9.3 version in my computer. I didn't try it in other device.

![text](https://user-images.githubusercontent.com/36484895/112949150-93974c00-9141-11eb-942e-ce51c0b889d0.jpg)

Thumbs up for 0.9.5!!!
It sounds really interesting.....

dl3sdo:
Thank you for reporting this issue, indeed that is really weird behaviour. I confirm it for Windows10 and will check it on Linux.

dg0yt:
Could you please double-check that the problem exists in v0.9.3?   
v0.9.4 added 'freetype' for PDF template support. This is the change most likely to have font-related side effects.

dl3sdo:
Reproducible on Ubuntu and Windows7.
Problem exists in 0.93 as well but not in 0.92.

lpechacek:
WTH? Replicated on openSUSE Tumbleweed.

lpechacek:
Bisected down to 3f11710594c1. Disabling the `ScalingIconEnginePlugin` in `src/main.cpp` makes the problem go away. The connection to icon scaling also explains the fancy pattern from the screenshot. We are seeing the TrueType icon (red T + black T) scaled to the width of the drop-down container. :)

Symbian9:
@dg0yt, May you force OBS to refresh `openorienterring-mapper-master`/`openorienterring-mapper-unstable` builds with latest `master` branch as actually all available builds for Linux (i.e. Debian 10) are affected by this issue.

PR #1929, opened 2021-03-31
ScalingIconEngine: Keep aspect ratio

dg0yt:
QIconEngine::actualSize() is documented to return the requested size
by default, but the actual implementations behave differently. In
particular, they seem to preserve the aspect ratio of the source,
fitting the actual size into the requested size. This commit changes
our icon engine to follow this pattern.
Fixes GH-1928 (Unreadable font names due to excessive font icon size).

### DIFF
--- a/src/gui/scaling_icon_engine.cpp
+++ b/src/gui/scaling_icon_engine.cpp
@@ -20,6 +20,8 @@
 
 #include "scaling_icon_engine.h"
 
+#include <algorithm>
+
 #include <Qt>
 #include <QtGlobal>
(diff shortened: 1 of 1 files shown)


dg0yt:
For the record, I don't think the `QFontComboBox`' behaviour of supplying the whole icon rect fits well with the default behaviour of `QIconEngine::actualSize()`. This particular icon just don't triggers strange side effects with other engines. I'm too tired of Qt to open another QTBUG.

dl3sdo:
Resolves also #2243.

Issue #1926, opened 2021-03-30
OCD area object pattern rotation not applied after import


dg0yt:
### Steps to reproduce
1. Import an OCD 8/9/10/11/12 file with rotated object patterns, e.g. vineyard
2. Patterns are not shown with the user defined rotation.
3. Activate "adjustable per object" for the fill patterns of the symbol.
4. Now the patterns are shown correctly

### Actual behaviour
Rotations are loaded but not applied, due to incorrect symbol configuration. They are lost on export.

### Expected behaviour
The symbol properties must be set up correctly.

### Configuration
Mapper Version: v0.9.5
Operating System: any

### More information

Before https://github.com/OpenOrienteering/mapper/pull/1700/commits/918dd9b0df1cc752dd65e83c0c259585e946e557 (v0.9.4), all area symbols were created with rotatable patterns during OCD import. This worked for vineyards et al. That commit put the pattern rotation property into the object rotation flag. Maybe this is the wrong place.

Note that even OCD 12 format is affected, despite being the subject of investigation for GH-1677/GH-1700 which are the context of the commit.

lpechacek:
Interesting. I'm seeing exactly the opposite. Rotation is enabled on objects where it's supposed to be disabled. There's something fishy and I'll take a closer look. Thanks for the report!

dg0yt:
AFAIU there is no such bit for fill pattern rotation, and that's why the fill pattern rotation area symbol property was unconditionally set to true before 918dd9b. This is different from the point symbol rotation property.

dg0yt:
To make an example: individual rotation can be set for marsh or undergrowth objects in OCAD?

lpechacek:
> To make an example: individual rotation can be set for marsh or undergrowth objects in OCAD?

Yes. I tried to rotate the undergrowth area pattern first and then I had to take a short break to recover.
![ocad-demo](https://user-images.githubusercontent.com/5584406/113339720-33fc9480-932b-11eb-9fe4-21fe111f98c3.gif)

The north orientation comes into play during map rotation. There is a checkbox (default ON) that can be unchecked, and then patterns of area symbols with the north orientation set are not rotated, while all other area patterns change their angle. This is the essence of #1677 which I misunderstood on the first attempt.
![image](https://user-images.githubusercontent.com/5584406/113340107-c00ebc00-932b-11eb-8b2b-c82d35535648.png)

Replicated with OCAD 8.12 demo edition.

[...]

PR #1931, opened 2021-03-31
Revert "OcdFileImport: Honor area hatch/structure north orientation"

lpechacek:
This reverts commit 918dd9b0df1cc752dd65e83c0c259585e946e557 as explained in issue #1926.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -1559,7 +1559,6 @@ Symbol* OcdFileImport::importAreaSymbol(const Ocd::AreaSymbolV8& ocd_symbol)
 	setupBaseSymbol(symbol, ocd_symbol.base);
 	setupAreaSymbolCommon(
 	            symbol,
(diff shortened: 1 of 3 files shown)


lpechacek:
Interestingly enough, the seemingly trivial revert demands more attention than expected. I'm converting the PR to a draft until the CI checks settle.

dg0yt:
Would you mind removing the CI/codespell commit? (I commented in my PR against your branch, and I did more changes in the meantime.)

lpechacek:
OK, no problem. I'll leave the codespell trouble all on you. At least, we know that all the checks pass with the fix. I'll reset the branch back to 808892c.

lpechacek:
The patchset is now clean from my POV. It does compile cleanly, tests pass and the structure is obvious. I've removed only a part of the FileFormatTest patch because I see value in the bundled test infrastructure rewrite. I have therefore deleted only the are hatch rotation test.

lpechacek:
Superseded by #2055. I no longer think that we should return to the original behavior.

PR #1948, opened 2021-05-08
iof_course_export: Fix typo in 'CourseDate' and replace 'CourseContro…

### LINKED ISSUE #1947: Simple course exported as IOF-XML does not comply with IOF XML schema
### Steps to reproduce
1. Select a course and export as Simple Course to IOF Data Standard 3.0 (xml)
2. Validate resulting file against IOF Data Standard 3.0 XSD ( https://github.com/international-orienteering-federation/datastandard-v3/blob/master/IOF.xsd ) with use of any validation tool (i.e. https://www.freeformatter.com/xml-validator-xsd.html )

### Actual behaviour
Validation fails - at first due to basic mistake - CourseDate tag is used instead fo CourseData.
After easy fix it throws a lot of absent but required attributes:
```
Cvc-complex-type.2.4.a: Invalid Content Was Found Starting With Element 'CourseControl'. One Of '{"http://www.orienteering.org/datastandard/3.0":PunchingUnitId, "http://www.orienteering.org/datastandard/3.0":Name, "http://www.orienteering.org/datastandard/3.0":Position, "http://www.orienteering.org/datastandard/3.0":MapPosition, "http://www.orienteering.org/datastandard/3.0":Extensions}' Is Expected., Line '8', Column '63'.
(...)
```

### Expected behaviour
Exported data should pass the validation or Mapper should not claim to export valid IOF XML course files.

### Configuration
Mapper Version: 0.9.5
Operating System: Windows

dl3sdo:
Fixes #1947

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -147,6 +147,7 @@ jobs:
         cmakeArgs: >
           -DCMAKE_BUILD_TYPE=Debug
           -DMapper_CI_ENABLE_COVERAGE=1
+          "-DIOF_3_XSD_FILE=$(IOF_3_XSD_FILE)"
         buildArgs:
     - template: publish-coverage.yml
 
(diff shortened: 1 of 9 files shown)


dg0yt (commented):
Now it is fine from my side. Export compared against test file. And test file validated against schema.
Leaving a short time for other to have a look.

Issue #1951, opened 2021-05-12
Secondary point symbol


ollesmaps:
### Steps to reproduce
1. Export from Ocad MTBO symbol set - symbol Forbidden route
2. Import into OOM
3. Skipped secondary point 

### Actual behaviour
Secondary point missing. See zipped att.

### Expected behaviour
Correct import

### Configuration
Mapper Version: 0.9.5
Operating System: Win

![forbidden](https://user-images.githubusercontent.com/20535899/117962369-71d3dc00-b31f-11eb-8dba-0b9b78296338.jpg)
[forbidden-route.zip](https://github.com/OpenOrienteering/mapper/files/6465601/forbidden-route.zip)

lpechacek:
Confirmed. Mapper issues a warning on import ("In line symbol 711.0 'Forbidden route (automatically)': Skipped secondary point symbol.") but I don't see why the import routine skips the secondary symbol.

lpechacek:
Minimal reproducer created with OCAD 8. There should be dots and squares alternating on the line.
[ocad8-minimal-reproducer.zip](https://github.com/OpenOrienteering/mapper/files/6478197/ocad8-minimal-reproducer.zip)

dl3sdo:
![fobidde-route](https://user-images.githubusercontent.com/14225666/118262850-c610c480-b4b5-11eb-8170-1f14c2770678.jpg)
Output with OCAD 12

dl3sdo:
Making progress....
The 	else if (attributes.secondary_data_size > 0) part has a number of issues (of which the first one is the 'else' of course).
Removing the line_symbol->mid_symbol = new OcdImportedPointSymbol(); and replacing the 'elements' parameter in the line below by reinterpret_cast<const Ocd::PointSymbolElementV8*>(coords) leads to something like this:
![fobidde-route_mod](https://user-images.githubusercontent.com/14225666/118285175-7a6c1400-b4d1-11eb-82e6-a31836cba66d.jpg)

dl3sdo:
I manually modified the symbol in Mapper (see attachment). However, adding the 'falling' rows with an offset leads to adding an offset to the end symbol and thus will lead to a line getting longer as configured. I don't see an easy way to import these kind of OCAD symbols as OCAD treats the secondary point symbol in combination with the main one, whereas Mapper processes all line parts forming an element (e.g., the mid element) independently.
[...]

Issue #1506, opened 2020-01-25
Sometimes dash corners are not enough


samlaminilepre:
### Steps to reproduce
1.Draw a cliff with 90° angles
If you use a dash corners is even worst

### Actual behaviour
Tag lines will cross each other, if you use a dash corners is even worst

### Expected behaviour
It would be great to be able to delete a mid simbol when needed, without having to cut it and use another simbol without tag lines or other mid simbol.
### Configuration
Mapper Version:0.9.1
Operating System:windows
![scarpate](https://user-images.githubusercontent.com/45790530/73120557-03270f80-3f70-11ea-9a33-13c3484f25c6.JPG)

rem-maps:
I also have problems with dash corners - can someone help?
Maybe there is already a way to solve my problem but I can't find it.

Readability is a major target for me - I want my maps to be 'clean' and readable.

In the attached pictured you see what I want to achieve - but I cannot achieve it either with normal corner point (I don't want the mid point tag so near the fence corner) or with dash point (I don't want the mid point tag in that short section of fence) .

The only workaround I can figure out is to duplicate symbol and have one of them without flag 'show at least one mid symbol'.
But I should duplicate too many symbols and cut fence into two or more parts and I don't like it.

Basically I would like to have a type of point that removes any mid point nearby.

Thanks

![fence](https://user-images.githubusercontent.com/59602742/73557092-e77ca700-4450-11ea-9e4f-c75400c155d4.png)

mlerjen:
@rem-maps 
![image](https://user-images.githubusercontent.com/15611653/80283413-a296a280-8717-11ea-936d-7c785525b45a.png)
By deactivating the "Show at least one mid symbol" option you can shoo away tags from certain sections.
![image](https://user-images.githubusercontent.com/15611653/80283501-4c762f00-8718-11ea-96c7-4ddace8c0384.png)

rem-maps:
@mlerjen  I don't get your sentence at all -  I don't understand if there are spelling mistakes or just I don't understnad your english. Can you re-write it?
>>Modified symbol opted out the SALOMS-option but also with dash. see just below here.
By deactivating the "Show at least one mid symbol" option you can shoo away tags from certain sections.

is your suggestion different from what I already said?
>>The only workaround I can figure out is to duplicate symbol and have one of them without flag 'show at least one mid symbol'.

mlerjen:
@rem-maps: my bad.

rem-maps:
[...]

PR #1958, opened 2021-05-20
Handle secondary symbols for OCD line symbols

dg0yt:
OCD secondary symbols are placed in the middle between regular mid
symbols. Mapper doesn't have explicit secondary symbols, so this must
be implemented via a combined symbol.
For test data such as MTBO symbol 711 Forbidden route, it turned out
that the secondary symbol also needs to be mirrored.

No export yet. Feedback and contributions welcome.


dg0yt:
Task: GH-1951

dg0yt:
> For test data such as MTBO symbol 711 Forbidden route, it turned out
that the secondary symbol also needs to be mirrored.

Actually, no. I'm afraid we just never imported the secondary symbol correctly because we never passed the right offset to setupPointSymbolPattern. :roll_eyes:

dg0yt:
I tried to break none of the existing reconstructions of line symbols (framing, double line). But I didn't test with many OCD files.

src/fileformats/ocd_file_export.cpp
```diff
@@ -2111,18 +2155,31 @@ void OcdFileExport::exportCombinedSymbol(OcdFile<Format>& file, const CombinedSy
...
 		{
-			if (num_parts < 3)
+			if (num_parts < 4)
 				parts[num_parts] = part;
 			++num_parts;
 		}
 	}
 	
+	const LineSymbol* secondary_points = nullptr;
+	if (combined_symbol->getType() && num_parts > 1)
```
dg0yt, line 2170: `...getType() == Symbol::Line`

src/fileformats/ocd_file_export.cpp
```diff
@@ -2111,18 +2155,31 @@ void OcdFileExport::exportCombinedSymbol(OcdFile<Format>& file, const CombinedSy
...
+	{
+		auto const* last_part = parts[num_parts - 1];
+	    if (maybeSecondaryPoints(last_part))
```
dg0yt, line 2173: <TAB>

src/fileformats/ocd_file_import.cpp
```diff
@@ -1525,30 +1525,51 @@ void OcdFileImport::setupLineSymbolPointSymbols(OcdFileImport::OcdImportedLineSy
...
+	line_symbol->setColor(nullptr);
+	line_symbol->setLineWidth(0);
 	
-	int part = 0;
-	full_line->setPart(part++, main_line, true);
-	main_line->setHidden(false);
-	main_line->setProtected(false);
-	main_line->setName(main_line->getName() + tr(" - main line"));
+	auto* point_symbol = new OcdImportedPointSymbol();
+	const Ocd::OcdPoint32* coords = reinterpret_cast<const Ocd::OcdPoint32*>(elements) + attributes.primary_data_size;
```
dg0yt, line 1540: Oh. Do we check boundaries, in similar contexts?

dg0yt (commented):
Homework...

dg0yt:
We might add a similar convenience symbol to our symbol sets. This should expose the new code paths to automated testing.

src/fileformats/ocd_file_import.cpp
lpechacek: We discussed that in an OpenOrienteering "team internal" discussion back in August 2018. The conclusion was that I don't mind crashes resulting from invalid input, while you promoted more convenient behavior in the face of invalid data. 

I understand your point, but it adds error handling code. From my POV, as long as we don't de-reference such a pointer for **write**, we should be shielded from the worst security issues.

src/fileformats/ocd_file_export.cpp
```diff
@@ -1975,6 +1979,46 @@ void OcdFileExport::setupTextSymbolFraming(const TextSymbol* text_symbol, OcdTex
...
+	if (symbol->getType() != Symbol::Line)
+		return false;
+	
+	auto* line_symbol = static_cast<const LineSymbol*>(symbol);
+	if (line_symbol->getColor()
+	    || line_symbol->getLineWidth()
+	    || line_symbol->hasBorder())
+		return false;
+	
+	auto const is_empty = [line_symbol](auto getter) -> bool {
```
lpechacek, line 1996: Having `is_populated()` function here might help legibility. We are looking for a line symbol that has the mid-symbol populated and the other symbols empty. The test `!is_empty(...)` requires interpretation.

src/fileformats/ocd_file_export.cpp
```diff
@@ -1975,6 +1979,46 @@ void OcdFileExport::setupTextSymbolFraming(const TextSymbol* text_symbol, OcdTex
...
+	if (is_empty(&LineSymbol::getMidSymbol))
+		return false;
+	if (!is_empty(&LineSymbol::getStartSymbol))
+		return false;
+	if (!is_empty(&LineSymbol::getDashSymbol))
+		return false;
+	if (!is_empty(&LineSymbol::getEndSymbol))
+		return false;
+	
+	return true;
```
lpechacek, line 2009: I'd be fine with a more compact:
```
if (is_populated(&LineSymbol::getEndSymbol)
    || is_populated(&LineSymbol::getStartSymbol)
    || is_populated(&LineSymbol::getDashSymbol))
        return false;

return is_populated(&LineSymbol::getMidSymbol);
```

Which better expresses the qualification criteria for a possible secondary symbol representation.

src/fileformats/ocd_file_export.cpp
```diff
@@ -1984,12 +2028,12 @@ int OcdFileExport::checkCombinedSymbol(const CombinedSymbol* combined_symbol) co
 	// The implementation must mirror exportCombinedSymbol()!
 	
 	auto num_parts = 0;  // The count of non-null parts.
-	const Symbol* parts[3] = {};  // A random access list without holes
+	const Symbol* parts[4] = {};  // A random access list without holes
```
lpechacek, line 2031: The magic number defining the array size is a candidate for a symbolic constant (ES.45).

src/fileformats/ocd_file_export.cpp
lpechacek: Isn't `combined_symbol` always `Symbol::Combined`?

src/fileformats/ocd_file_export.cpp
```diff
@@ -2140,9 +2197,12 @@ void OcdFileExport::exportCombinedSymbol(OcdFile<Format>& file, const CombinedSy
 			return;
 		case Symbol::Line:
 			{
-				auto copy = duplicate(static_cast<const LineSymbol&>(*parts[0]));
+				auto const* main_line = static_cast<const LineSymbol*>(parts[0]);
```
lpechacek, line 2200: AFAIU, this code will never kick in. A few lines above we decrease the number of parts from two to one and the test for `combined_symbol` type is still `Symbol::Combined`. Therefore the switch goes through the `::Combined` branch and the flow goes to `exportGenericCombinedSymbol()`.

lpechacek (changes requested):
The export path needs more love. Overall, the patch set is going in the right direction, I'd say. Thanks!

src/fileformats/ocd_file_import.cpp
```diff
@@ -1525,6 +1529,38 @@ void OcdFileImport::setupLineSymbolPointSymbols(OcdFileImport::OcdImportedLineSy
...
+}
+
+void OcdFileImport::mergeSecondarySymbol(CombinedSymbol* full_line, LineSymbol* main_line, PointSymbol* secondary_symbol)
+{
+	auto* secondary_line = new OcdImportedLineSymbol();
+	for (auto i = 0u; i < full_line->number_components; ++i)
+		secondary_line->setNumberComponent(i, full_line->getNumberComponent(i));
+	secondary_line->setName(full_line->getName() + tr(" - secondary line"));
+	secondary_line->setMidSymbol(secondary_symbol);
+	secondary_line->setShowAtLeastOneSymbol(false);
```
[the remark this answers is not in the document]
dg0yt: This particular discrepancy is intentional with regard to the MTBO symbol. For short lines , you may get
~~~
|\|
|\/\|
~~~
but not
~~~
|X|
~~~
But I have no idea how it is rendered in the OCAD software, or what particular aspect of #1506 I should look at.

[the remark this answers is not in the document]
lpechacek: I see. The mixed mode makes sense for this particular case.

Ad setting the "at least one mid symbol", see the following comparison:
image

On the left, there is an example rendering of fixed length lines with "at least one mid symbol" set to ON. The inclined brown line indicates the position of a dash point and the number on the left shows the distance of the dash point from the beginning of the line.

As seen in the picture, Mapper squeezes the mid symbol into tiny spaces, which is the aspect of #1506 I intended to raise.

With "at least one mid symbol" set to OFF, the picture is as follows:
image

There are still differences in rendering of the ruined wall and ruined fence, but the other symbols' appearance is similar on both sides. For that reason, I'm in favor of setting "at least one mid symbol" to OFF **on .ocd import** for ISOM maps. The other map standards would have to be evaluated separately.

In the context of this pull request, I advise staying compatible with the current behavior ("at least one" == ON) so that we don't open too many cans of worms at once.

PR #1960, opened 2021-05-21
Show/Hide map parts

### LINKED ISSUE #1959: Feature implementation: show/hide map parts
As discussed in #1121 it would be beneficial to show and hide individual layers.
I implemented that feature, but used a simple GUI approach by just adding a map part dependent 'Hide current part' or 'Show current part' to the Map parts toolbar (as mentioned by @Symbian9 in #1121).

The visibility is stored in a new attribute:
<part name="Part1" visibility="100"><objects count="393">
<part name="Part2" visibility="0"><objects count="146">
Current versions of Mapper and Purple Pen do open these kind of files properly.
I used the values 0 and 100 in order to provide a migration path if somewhen later this becomes an true opacity value.

Example by using my favourite sample map and after having moved some objects to another map part:
ShowBothParts
The Map parts toolbar offers the 'Hide current part' as the current part is visible and thus can be hidden.

Map parts toolbar after part was hidden:
Part1Hidden

The change considers opening previous maps as well as considering the undo feature on map parts.

dl3sdo:
Pull request adds the possibility to hide/show map parts.


dg0yt:
Let's have a call later tonight.

dg0yt:
Apart from other issues, the original approach broke the rendering of symbol icons. This, and the addition of a new attribute in all files, were breaking the tests after the first commit.

mpickering:
What's the status on this PR? It would be a nice feature to be merged.

dg0yt:
> What's the status on this PR? It would be a nice feature to be merged.

Since initial submission, a serious problem was resolved (workload on critical path for drawing). But I think it still needs more work and testing.  You don't want to slow down the mobile app, or be left in the forest with the desired map part hidden.

dl3sdo:
For Android the capability to show and hide parts is currently missing. I could imagine that hidden map parts are shown in light gray in popup menu where you can choose the active part. Maybe it's possible to use a long press to switch the visibility of a map part. Another option could be that you switch the visibility if you 'select' the already active part. Selecting an inactive part just makes that part active as today but does not change the visibility.

carlogico:
I played with this a bit on MacOS, and it seems to work ok. I think this would be an extremely useful feature. How can I help to get this merged soon?

PR #1964, opened 2021-06-02
OCDImportExport: import and export of Hatch areas and Baseline view state

dl3sdo:
Mapper currently saves the View settings of Hatch areas and Baseline view to .omap.
With this change Mapper exports and imports the View settings to .ocd (Baseline view is denoted as Keyline mode and is present for OCAD 11 and newer, however not for all editions and not in the OCAD viewer).


dg0yt:
Thanks Matthias!  I made a few small changes. I will squash the commits on merge.

PR #1967, opened 2021-06-08
Adjusting size of symbol 530 in ISSprOM

### LINKED ISSUE #1920: 530 Prominent man-made feature – ring 1.2 mm instead of 1.0
I am using version 0.9.4
and I just realized symbol 530 Prominent man-made feature – in the Mapper default  ISSPROM symbol set is 1.2 mm and it should be 1.0 mm (like cairn or tree - see pictures: black circle is bigger now in Mapper)

manmade

man2

I can't believe no-one ever noticed it, so maybe I am doing something wrong?
please someone else check!
thank you

michael-77:
Correcting the inner radius of the prominent man-made feature - ring to match ISSprOM

Ref: https://orienteering.sport/iof/mapping/

Resolves #1920


dg0yt:
Thanks for the PR @michael-77! I added the same change to the source symbol set. This should resolve the failed checks.

michael-77:
Thank you. I couldn't work out why those had failed.

Issue #1677, opened 2020-07-13
Losing "Hatch/Structure oriented to north when rotating the map" symbol property in Ocad files.


andzs:
### Steps to reproduce
1. Open attached Ocad12 Sample.ocd file in OOM. [Sample.zip](https://github.com/OpenOrienteering/mapper/files/4915345/Sample.zip)

2. Rotate map (Map->Rotate) in OOM for 15 degrees for example. In OOM everything works fine. Appearance of symbols on map
is correct after rotation.
3.  Save this map back in Ocad 12 format as Sample_rotated.ocd
4. Open map Sample_rotated.ocd in Ocad12 and rotate it back (Map->Transform->Rotate Map) by -15 degreees and uncheck "Rotate objects with symbols oriented to North".

### Actual behaviour
Resulting map is oriented back, but symbol appearance is wrong. Pattern of marshes and vegetation symbols is rotated because
symbol definitions are lacking "Hatch/Structure oriented to north when rotating the map" property.

### Expected behaviour
Editing Ocad files in OOM and saving back in ocd format should retain "Hatch/Structure oriented to north when rotating the map" property of symbols.

### Configuration
Mapper Version: 0.9.3
Operating System: OS Name: Microsoft Windows 10 Pro OS Version: 10.0.18362 N/A Build 18362

lpechacek:
Confirmed. Mapper makes objects rotatable. This is a bug in the importer routines. Thanks for the report!

lpechacek:
I've got a fix and now I'm investigating why the defect was not caught by automated testing.

For the record, simplified test routine:
1. Open the attached Ocad12 Sample.ocd file in OOM.
2. Select one of the objects and try to rotate its pattern with Rotate pattern tool

Actual behavior:
The pattern happily rotates.

Expected behavior:
The Rotate pattern tool is disabled and no pattern rotation is possible.

lpechacek:
I'm reopening the issue as my original fix was wrong. The explanation is in https://github.com/OpenOrienteering/mapper/issues/1926#issuecomment-811033479.

lpechacek:
My thoughts on this issue as of today.

* The "area pattern rotation" means two different things. The Mapper interpretation is "the angle between the pattern and the magnetic north shall be ZERO".
* The alternate interpretation is "the angle between the pattern and the magnetic north shall be CONSTANT."

[...]

PR #1975, opened 2021-07-21
OcdFileImport: Handle objects with rotated patterns

lpechacek:
This pull request is a softer attempt at fixing GH-1926. The logic of OCD "hatch/structure is orientated to north" feels somewhat ad-hoc to me even after another look.

This patch makes Mapper recognize two categories of the possible three options supported by .ocd format:
1) areas with a non-zero pattern angle against the north, the pattern angle changes upon map rotation
2) areas with zero pattern angle against the north, the pattern is never rotated

The third possibility (areas with non-zero pattern angle against the north, the pattern angle **is not changing** upon map rotation) gets translated to 1) and an import warning is emitted.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -93,6 +93,27 @@ static QTextCodec* codecFromSettings()
 }	
 
 
+// Fixup for OCD area objects that have their hatch/structure "oriented to the
+// north," but they are in fact rotated. Mapper does not have this category, so
+// we remove the rotation lock in this case.
+
+void fixupPatternRotation(OcdFileImport* importer, Object* object)
+{
+	if (object->getSymbol()->getType() != Symbol::Area)
+		return;
+	
+	auto* object_symbol = const_cast<AreaSymbol*>(object->getSymbol()->asArea());
+	if (object_symbol->getNumFillPatterns()
+	    && !object_symbol->hasRotatableFillPattern()
+	    && !qFuzzyIsNull(object->getRotation()))
+	{
+		for (auto n = 0; n < object_symbol->getNumFillPatterns(); ++n)
+			object_symbol->getFillPattern(n).setRotatable(true);
+		importer->addSymbolWarning(object_symbol,
+		                           OcdFileImport::tr("Removing rotation lock due to existence of an object with a rotated pattern."));
+	}
+}
+
 }  // namespace
 
 
@@ -855,8 +876,11 @@ void OcdFileImport::importObjects(const OcdFile<Ocd::FormatV8>& file)
 	{
 		if (ocd_object.entry->symbol)
 		{
-			if (auto object = importObject(*ocd_object.entity, part))
+			if (auto* object = importObject(*ocd_object.entity, part))
+			{
 				part->addObject(object, part->getNumObjects());
+				fixupPatternRotation(this, object);
+			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


lpechacek:
No build warnings and the patch looks good to me. We can certainly talk about whether the pattern north orientation can be called "rotation lock" and similar, but I think that the general idea about how to handle the conceptual discrepancy between the two mapping applications is clearly expressed in the patch.

The aim of this PR is to supersede #1931.

lpechacek:
@dg0yt I'd be grateful for your review.

dg0yt (commented):
The PR fixes the broken wineyards patterns in our existing maps.
However, it doesn't work for using an OCD symbol set for a new map - the patterns cannot be rotated without editing the symbol first.

One old test map had a rotated undergrowth pattern. This case is covered well by the warning message when I review them carefully. But on the other hand, it allows rotation were I don't want it.

With these comments, do you think the net effect is positive for this heuristics, in contrast to the original behaviour restored by #1931?

lpechacek:
Thanks, Kai, for the review. The delay in my response was caused by my effort to find peace with the OCAD interpretation of the symbol north orientation. ISOM says that most of the map symbols are oriented to magnetic north, therefore also oriented against the paper edges. I thought that the Mapper implementation is the only possible interpretation of the standard but it isn't...

With the new knowledge about how OCAD handles the symbol north orientation, I tend to agree that making all area patterns rotatable on import makes Mapper match the OCAD user experience. Also, a large number of non-compliant maps out there will smoothly pass the import, ensuring user satisfaction.

On the concept level, going back to the original behavior, we not only open one more way to breaching ISOM but we are re-introducing an asymmetry in the export/import path. Mapper exports non-rotatable area patterns as "north oriented" to .ocd format (see `OcdFileExport::exportAreaSymbolCommon()`). But the import path was dropping the north orientation flag, losing our own information that can be well represented even in the foreign file format.

For me, the decision between the two pull requests is a decision between user comfort (original behavior) and stricter ISOM adherence and consistency (status quo). I understand that with perfection we are repelling users, and with comfort we are littering the codebase. That said, I'm leaning slightly towards masking the OCAD map issues in the import path for the sake of increased Mapper popularity.

P.S. Please note that the map provided in the context of issue #1926 has all symbols marked as oriented to the north. That means that its symbol set needs fixing. I've examined a few other .ocd maps I have at hand, and the area north orientation flag is set correctly in their symbol sets and some of them even imported without triggering the new warning.

dg0yt:
I have a tendency to go with #1931 for now, and perhaps add an explicit attribute for the OCD feature sooner or later.

Some parts of export/import will just remain lossy. But as long as you the original format, it should be possible to restore the original symbol set if needed after a round trip of editing.

lpechacek:
Fine with me. I'll write an explanatory comment in #1677 and mark it as WONTFIX. Thanks!

lpechacek:
Ummm, life is full of surprises. While formulating the comment for the aforementioned issue about area hatches, I noticed that Mapper (silently) removes the rotation lock on point symbols when there is a rotated object (https://github.com/OpenOrienteering/mapper/blob/20065ec47cca0b28191ddcfa8ed7d353dbab9d02/src/fileformats/ocd_file_import.cpp#L1932-L1939). It is the same principle as in this proposal.

This rotation lock removal behavior was first implemented in commit 2cce659cf0a04 without explanation *why* is this needed. We know the why now, though. Where to go from here? Align the behavior, either way, across all symbol types?

lpechacek:
https://github.com/lpechacek/mapper/commit/e8a3925ab0457b691c76755c7fb0534f23108996 is my final say on the object rotation topic. The export path needs review for the area, point, and text objects.

PR #1994, opened 2021-10-23
Add support for proj.h, proj_api.h is removed in PROJ 8.

sebastic:
As reported in [Debian Bug #983254](https://bugs.debian.org/983254), mapper fails to build with PROJ 8 due to the removal of `proj_api.h`.

This is fixed by updating `FindPROJ4.cmake` to also support `proj.h`.


dg0yt:
The CMake Find Module approach hits a wall with static linkage, given the varying transitive usage requirements of modern PROJ (sqlite3, libtiff (-> liblzma, libjpeg, zlib), curl (-> openssl)). I offer a different approach in #2006: Use either PROJ config (exported from CMake builds) or proj.pc pkgconfig (exported from autotools build). So the responsibility to declare transitive usage requirements is entirely encapsulated in in the PROJ package.

AFAICT from OBS this works across a range of Linux distribution and releases from Debian 9 (CMake 3.7, PROJ 4.9, autotools) until next releases (for configuration and build).

sebastic:
PROJ 9 scheduled for release in March 2022 will switch to CMake, but should also still install `proj.pc` for projects using pkg-config.

The changes from this PR have been included in the Debian package to fix the build failure in the mean time.

dg0yt:
> PROJ 9 scheduled for release in March 2022 will switch to CMake, but should also still install `proj.pc` for projects using pkg-config.

PROJ 9 will drop autotools. The CMake build system is already available. But we discussed this before.

dg0yt:
Solved with #2006 instead.

PR #2004, opened 2021-11-14
Issue 650 - Move Parallel Tool

lpechacek:
Although the Move Parallel tool does not behave perfectly in every situation (shape distortion on small curves and bevel-join-like corner cuts on sharp corners), it works fine on "sensible" shapes. This is still an improvement over the current state where the best approximation of this tool is the Line Edit tool or tricks with helper double lines.


dl3sdo:
One small observation: although the Move Parallel tool can only be activated once an appropriate object has been selected, it stays activated once the object becomes unavailable (e.g., has been moved to another map part, has been deleted or it's underlying symbol has been deleted). That's not a big issue since you can deactivate the tool by choosing another tool.

lpechacek:
> One small observation: ...

Thanks for testing! I've aligned the new tool behavior with the other tools on this front.

src/core/virtual_path.cpp
```diff
@@ -364,6 +366,27 @@ ClosestPathCoord VirtualPath::findClosestPointTo(
...
 			result.distance_squared = dist_sq;
+			// Here we create predictable behavior for corners and handle 
+			// the ends. Although corners do not have a tangent in 
+			// the mathematical sense, the humans intuitively expect the 
+			// "tangent" to go across the corner and divide the plane into the
+			// inner and outer parts.
+			bool in_ok = false;
+			bool out_ok = false;
+			auto const incoming = calculateIncomingTangent(path_coord.index, in_ok);
+			auto const outgoing = calculateOutgoingTangent(path_coord.index, out_ok);
```
dg0yt, line 377: I'm sceptical about putting this into the loop if it is only for the corner cases (literally). When the search starts at the wrong end of the path, it is called and discarded very often. And for many cases, the tangent isn't used at all. I would say that this is the reason why the tangent is not part of this function and not a member of `ClosestPathCoord`.

IIUC the alternative is to call `VirtualPath::calculateTangent(size_type i)` for the result. This function also takes into accout a minimal length. But I don't if it does the right thing for this tool.

dg0yt (commented):
Thank you for the PR. I'm sorry that I couldn't test the feature yet.

src/core/virtual_path.cpp
lpechacek: Ack and thanks for the remark. On a closer look, I can see that `calculate*Tangent()` is not a good fit for an inner loop. `VirtualPath::calculateTangent()` cannot be used as is, because we calculate an "average tangent" on the corner by summing up the two vectors. However, `VirtualPath::calculateTangent()` already has the `backward` parameter so we can introduce a "bidirectional" mode of operation. Opinions?

dg0yt: I'm not sure if I can follow. The `VirtualPath::calculateTangent(size_type i)` overload I mentioned looks very much like your implementation. It just a little bit more condensed, and it normalized the tangent vector lengths before combining them.

lpechacek: @dg0yt, now I see what you mean. I didn't notice the method overload and looked at `VirtualPath::calculateIncomingTangent(size_type i, bool backward, bool& ok) const` all the time. `VirtualPath::calculateTangent(size_type i)` is fifty lines earlier in the code and separated from the rest of the `calculate*Tangent()` methods by `calculateTangentScaling()`.

I confirm that `VirtualPath::calculateTangent(size_type i)` seems to perform the function I need. Thanks for the hint!

lpechacek: Patch updated.

PR #2006, opened 2021-12-14
Find PROJ via cmake config or via pkg-config

dg0yt:
Drop the custom find module for PROJ. Cmake config or pkg-config data is provided from PROJ's build system. This covers PROJ 4.9 to 8.2, dynamic and static linkage, including transitive dependencies.

Alternative to  #1994.


dg0yt:
azp /run

Issue #2008, opened 2021-12-17
NAD83 changes in PROJ 8.2.0


dg0yt:
With https://github.com/OSGeo/PROJ/commit/85f496b26702bf6566296072795e821d70156c09, PROJ 8.2.0 imports an EPSG database change which adds a "null transformation between NAD83(2011) and WGS 84", impacting a number of tests in PROJ's test suite. I guess that this also triggers the new test regressions we see in linux distributions:
- [Fedora Rawhide](https://build.opensuse.org/package/live_build_log/home:dg0yt:branches:home:dg0yt:staging/openorienteering-mapper/Fedora_Rawhide/x86_64)
- [openSUSE Tumbleweed](https://build.opensuse.org/package/live_build_log/home:dg0yt:branches:home:dg0yt:staging/openorienteering-mapper/openSUSE_Tumbleweed/x86_64)
- [Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=998827)
  @sebastic I couldn't reproduce the error with GDAL 3.4 + PROJ 8.1.1, so that regession's attribution to GDAL 3.4 seems to be wrong.

In fact, the sets of failing and succeeding tests now match the situation before PROJ 6.
I don't see a quick remedy other than updating the expected failures in the test suite.

The changed behaviour of PROJ will impact users on the next generation of linux distributions.
It won't impact our windows, macos or android binaries as long as we don't build with PROJ 8.2.
We should consider upgrade paths to a better solution with Mapper 1.0.

PR #2009, opened 2021-12-17
TemplateTest: Update expected fails for PROJ 8.2.0

dg0yt:
Task: GH-2008


dg0yt:
Tested successfully on OBS.

Issue #2011, opened 2021-12-21
New shortcuts


mlerjen:
`###` Steps to reproduce
1. Shortcuts are a cool thing

### Actual behaviour
1. Not every function has a shortcut.

### Expected behaviour
1. Add X for cut-away
2. Consider adding shortcuts to Android functions as well. The basic set would contain shortcut for zoom in, zoom out, center map to gps position and open/close template dialog. Being able to trigger shortcuts with a minimal external BT keyboard would smoothen the Android workflow very much.

### Configuration
Mapper Version: 0.9.5
Operating System: Windows&Android

puzzlepaint:
Regarding point 1.: Here is a PR that sets X as a shortcut for "Cut away from area": #2013
(Of course, ideally the shortcuts should really be configurable in the program.)

dg0yt:
> ideally the shortcuts should really be configurable in the program.

Still waiting for a usable shortcut editor.

ollesmaps:
How far is the option to give voice commands in Android? I imagine this would be ideal for shortcuts. It could streamline the process even more.

mlerjen:
I realize that X was added as a short-cut for "Cut away from area", while the original request was to apply it to "Cut away", which since then was renamed to "Erase area".  ...

PR #2013, opened 2021-12-22
Shortcuts: Add 'X' as a shortcut for 'Cut away from area'

puzzlepaint:
Adds 'X' as a shortcut for 'Cut away from area', as suggested in #2011.


dg0yt (approved):
Thanks @puzzlepaint.

PR #2014, opened 2021-12-25
map_dialog_scale: Disable checkbox for local CRS

dl3sdo:
For a local CRS, the 'Change map scale' dialog allows to check or uncheck the "Adjust georeferencing reference point" checkbox although this does not make sense (therefore the "Georeferencing reference point" checkbox is disabled).
The commit simply disables the "Adjust georeferencing reference point" checkbox for local CRS.

### DIFF
--- a/src/gui/map/map_dialog_scale.cpp
+++ b/src/gui/map/map_dialog_scale.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2019-2020 Kai Pastor
+ *    Copyright 2019-2021 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -138,7 +138,7 @@ void ScaleMapDialog::updateWidgets()
 {
 	other_x_edit->setEnabled(center_other_radio->isChecked());
 	other_y_edit->setEnabled(center_other_radio->isChecked());
-	adjust_georeferencing_check->setEnabled(!center_georef_radio->isChecked());
+	adjust_georeferencing_check->setEnabled(center_georef_radio->isEnabled() && !center_georef_radio->isChecked());
 }
 
 void ScaleMapDialog::okClicked()


dl3sdo:
Example for the forest sample map:
MapScaleFix

pkturner:
The "Adjust georeferencing reference point" checkbox is actually meaningful for "Local" georeferencing.

IMO it's a setting that's difficult to understand even when the georeferencing is the usual "Geospatial". The natural expectation when changing the scale is for map objects to keep their geographic coordinates unchanged. The reference point has a location among the map objects and a location in geographic coordinates. For that to change as a side effect of changing the scale would be weird. But that is what un-checking either of these boxes does: when the box is un-checked it says to keep the object and reference point at the same map coordinates even though the map scale has shrunk or expanded.

Both the "Scale map object positions" and "Adjust georeferencing reference point" are the defaults, and IMO users would be happier almost all of the time if they were hidden.

In the "Local" case those checkboxes still both have an effect, though the meaning depends on how much the user relies on the local coordinates, particularly on their units of meters. Just look at the effect in the Map::changeScale method. The 'scale_georeferencing' parameter determines whether the MapRefPoint gets moved. The projected (or local) reference point is not moved, so there _is_ an effect.

dl3sdo:
@pkturner: thank you for the explanation. Yes, I had overseen that the 'Map coordinates' reference point' will be changed depending on that checkbox.
Would it be better to rename it?
AlternativeText

pkturner:
@dl3sdo I like that change to the text of the checkbox.

dl3sdo:
@pkturner: I propose to also change the Scaling center radiobutton as well as to do the equivalent changes in the Rotate Map Dialog:
ScaleRotateNewDialog

I noticed the difference to the 'Map coordinate system origin' above.
As a native speaker, what would you propose?
Use 'Map coordinates system origin' and 'Map coordinates reference point' instead and thus treat it as a notion of its own?
Using of genitives s' is not common in Mapper.

I will then close this pull request and open a new one.

pkturner:
> Use 'Map coordinates system origin' and 'Map coordinates reference point' instead and thus treat it as a notion of its own?
Using of genitives s' is not common in Mapper.

The author made an effort to keep the radio button text short. "Coordinate system" is the familiar term, so "coordinates system" would be awkward if not incorrect.

"Map coordinates' reference point" is short for "Reference point, expressed in map coordinates". I would change it to "Map reference point".

There's also "Adjust map coordinates' reference point" among the options. In that case it's not so simple to change it to "Adjust map reference point" because in the most common case, the user probably doesn't view this as a change to the reference point, at all. The option "Adjust map coordinates' reference point" refers to the mechanism which preserves the georeferencing of map objects, by scaling the coordinates of the map reference point. Consider changing it to "Adjust coordinates of map reference point".

dl3sdo:
Scott, thank you for your proposal, which I implemented in #2029

dg0yt:
> There's also "Adjust map coordinates' reference point" among the options.
> ...
> Consider changing it to "Adjust coordinates of map reference point".

Note that the original label explicitly refered to the map coordinate system, as opposed to the projected or geographic coordinate system. If we remove the coordinate system reference, I would shorten it to:
"Adjust map reference point".

dl3sdo:
@Kai: Since I also had a slight preference for the shortened "Adjust map reference point" I changed and force pushed it accordingly.

Issue #2016, opened 2021-12-28
Map based on OpenStreetMap has its georeferencing changed


pkturner:
### Steps to reproduce
1. Create a large map by import from an OpenStreetMap export.
2. Open the map with Mapper built with PROJ >= 7.2.
3. View geographic coordinates of objects. View GPX tracks.

### Actual behaviour
The objects' geographic coordinates are shifted by up to 5 meters, as are GPX track templates.

### Expected behaviour
The map objects are positioned as in the original OpenStreetMap, and alignment of GPX tracks is unchanged.

### Configuration
Mapper Version: 0.9.5

Operating System: Linux with PROJ 7.2.1 and 8.2.0. Android and Mac (PROJ 7.1.1 I think)

### Investigation
Mapper uses an orthographic projection when creating a map by importing from OpenStreetMap data. Prior to 7.2, PROJ used a spherical projection for +proj=ortho. Then at 7.2 it added support for a true ellipsoidal projection, changing the meaning of the specification strings that had been used by Mapper's maps.

A user who encounters this effect can fix it by inserting "+f=0" into the specification string of the map.

A future release of Mapper that uses PROJ 7.1.1 would do better to include "+f=0" in the specification string when setting up an orthographic projection, because that would provide forward compatibility for any maps created. On the other hand, if the next Mapper release uses PROJ 7.2 or better, then the specification string could be left as is. The map would be created with an ellipsoidal projection which is superior, being isotropic at its reference point.

### More Background
I noticed the problem due to field checking a point 5 km west of the map's reference point. The latitude of the map is 36° N. At that latitude, the spherical orthographic projection is distorted in horizontal scale by 0.12% and in vertical scale by 0.3%. The ellipsoidal orthographic projection used by PROJ 7.2 does not have that distortion.

PR #2020, opened 2022-01-07
Keep orthographic projections consistent

pkturner:
PROJ is changing the way it handles '+proj=ortho' in spec strings, as of version 7.2. Previously, that projection performed a spheroidal calculation, even with the WGS84 datum. With the change, PROJ uses the precise ellipsoidal shape when projecting.

I encountered and reported this in #2016, because I build Mapper on Linux where PROJ has been at version 7.2 for several months. The new PROJ alters the geographic coordinates of map objects, changing their alignment to georeferenced templates.

This PR includes a regression test of a GPX track template that fails with the latest PROJ due to the track becoming misaligned with the map. It also addresses the problem by inserting '+f=0' into every spec string that uses '+proj=ortho'.

The drawback to this way of keeping Mapper consistent is that it eliminates the possibility of using an ellipsoidal orthographic projection with Mapper, for the time being. The ellipsoidal version is superior, but probably does not matter for orienteering because the spheroidal projection distorts the scale by less than 1%. It looks to me that if this PR is accepted, then to regain support for ellipsoidal orthographic projections would require a new version of the XML file format, and a bigger code change than this one.

### DIFF
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -626,6 +626,7 @@ void Georeferencing::load(QXmlStreamReader& xml, bool load_scale_only)
 						if (language != literal::proj_4)
 							throw FileFormatException(tr("Unknown CRS specification language: %1").arg(language));
 						projected_crs_spec = xml.readElementText();
+						ensureOrthoIsSpheroidal(projected_crs_spec);
 					}
 					else if (xml.name() == literal::parameter)
 					{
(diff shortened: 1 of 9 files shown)


pkturner:
#2080 reported georeferencing problems that arose in the context of the spheroidal orthographic projection. As a result, I have become unhappy with this PR, because it entrenches Mapper's use of the spheroidal projection.

pkturner:
I developed PR #2092, which I prefer to this.

pkturner:
I developed PR #2092, which I prefer to this.

PR #2024, opened 2022-01-09
Preserve GPX track accuracy by accepting any authority

pkturner:
This is a possible alternative solution to #2008, rather than revising the expected template test failures as in #2009.

By setting up the PROJ transformation with an added option to permit _any_ authority, the tracks that relate to NAD83-based maps are placed more accurately than when processed via GDAL. With this change, those test tracks are positioned the same as they are by the Mapper executables in the 0.9.5 release.

### DIFF
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -383,7 +383,14 @@ ProjTransform::ProjTransform(const QString& crs_spec)
 	// Cf. https://github.com/OSGeo/PROJ/pull/1573
 	crs_spec_utf8.replace("+datum=potsdam", "+ellps=bessel +nadgrids=@BETA2007.gsb");
 #endif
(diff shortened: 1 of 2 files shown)


dg0yt (approved):
Thank you very much for this solution. 
I'm still curious about possible side effects on other contexts.

pkturner:
@dg0yt 
> I'm still curious about possible side effects on other contexts.

I have similar thoughts. I tried various combinations with `cs2cs --authority any`, without gaining further insight.  The reason I believe this change is appropriate is that at approximately 8.0.0, PROJ began to respect the transformation's `authority` property when creating a pipeline between CRSes, which seems to be the immediate cause of the #2008 issue. 

It's annoying that Mapper can't rely on PROJ to be stable in the transformations that it finds. As a result, Mapper can't completely avoid changes in alignment of templates, etc. from one release to another.

I also don't understand why EPSG doesn't have a good datum transformation between recent realizations of WGS84 and NAD83, comparable to ESRI's Helmert transformation.

PR #2025, opened 2022-01-16
text_symbol_settings: Fix Custom tabulator positions output

dl3sdo:
Custom tabulator positions in most cases were not shown with a precision of 3 digits.
This was caused by using 'g' in the toString() conversion which considers also the number before the decimal point.
Using 'f' instead fixes the output.
FontTabsFix


dl3sdo:
@lpechacek: may I ask you for a review and approval?

lpechacek:
Approval might be a strong word here, because I'm not the maintainer, but I can certainly review the code change.

lpechacek:
@dl3sdo the code change is correct from my POV. The format specifier is well chosen. One thing needs to be changed about the patch though. There must be an empty line between the patch title and the comment so that Git can properly extract the title, e.g. in `git log --oneline`. It also became common in OO to label the patch with the class name that it is changing than the file name and the text should be maximum 72 characters wide but that are true nits.

A properly formatted message would look like this:
```
TextSymbolSettings: Fix Custom tabulator positions output                
                                                                         
Custom tabulator positions in most cases were not shown with a precision 
of 3 digits. This was caused by using 'g' in the toString() conversion   
which considers also the number before the decimal point. Using 'f'      
instead fixes the output.                                                
```

In my testing, I found two areas for improvement in the code. You may look into them, if you want to further enhance the code.
1) The .ocd format unit of measure is **hundredths** of millimeter, not thousandths like in Mapper. So it makes sense to show only two decimal places.
2) The maximum number of tabulators is 32 (`Tabs: array[0..31] of longint;` in the original documentation). However, Mapper happily exceeds the number in the UI and does not emit any warning on export. Even worse, it writes `nTabs` larger than 32 into the file.

lpechacek (approved):

dl3sdo:
For the 2nd issue detected by you, I just created pull request #2045 
In addition it fixes an issue where large tabulator values were erroneously exported (conversion to qint16 instead of quint32).
It proves that it was very beneficial to ask you for a review.

dl3sdo:
I consider this PR ready to be merged.

PR #2027, opened 2022-01-18
GeoreferencingDialog: Hide 'Show reference point in' for local CRS

dl3sdo:
When switching between local and geospatial CRS links to OpenStreetMap | World of O Maps were shown even if they were not clickable (for local CRS).
This could also happen for a local CRS when using the Pick on Map and then pressing Reset.
Besides that there was an orphan 'Show reference point in:' label for local CRS.
The commit now uses setVisbible() for both, the label and the links, instead of just using setEnabled() on the links.

When setting up the dialog there was a doubled call edit_layout->addRow(show_refpoint_label, link_label); which lead to adding an empty row. This has been fixed by replacing it by adding a SpacerItem instead.

The last change corrects the link to OpenStreetMap.
mlat and mlon coordinates are now created by 'f', 6 instead of 'g', 10 which is the appropriate format for defining the precision after the decimal point.
BTW: 6 digits are equivalent to 11cm (or better).

### DIFF
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -188,7 +188,7 @@ GeoreferencingDialog::GeoreferencingDialog(
 	geographic_ref_layout->addWidget(geographic_datum_label, 0);
 	
 	show_refpoint_label = new QLabel(tr("Show reference point in:"));
-	link_label = new QLabel();
+	link_label = new QLabel(QLatin1String("-"));
 	link_label->setOpenExternalLinks(true);
 	
 	keep_projected_radio = new QRadioButton(tr("Projected coordinates"));
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -49,6 +49,7 @@
 #include <QSignalBlocker>
 #include <QSize>
 #include <QSpacerItem>
+#include <QString>
 #include <QStringList>
 #include <QStringRef>
 #include <QTimer>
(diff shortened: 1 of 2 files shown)


pkturner:
These changes look good to me.

dl3sdo:
Rebase and split of previous commit to separate independent changes

src/gui/georeferencing_dialog.cpp
```diff
@@ -251,7 +251,7 @@ GeoreferencingDialog::GeoreferencingDialog(
 	edit_layout->addRow(tr("Map coordinates:"), map_ref_layout);
 	edit_layout->addRow(projected_ref_label, projected_ref_layout);
 	edit_layout->addRow(tr("Geographic coordinates:"), geographic_ref_layout);
-	edit_layout->addRow(show_refpoint_label, link_label);
+	edit_layout->addItem(Util::SpacerItem::create(this));
```
dg0yt, line 254: I would prefer to not add a spacer.

dl3sdo: I agree, I compared the look of both variants and also prefer the one without the SpacerItem.
In addition I added a commit that fixes an annoying warning due to comparing int and unsigned int.

dl3sdo:
#2246 resulted in a merge conflict and the need to update a commit message.

dl3sdo:
Rebased (also due to 0a648fd) and squashed into two commits. I consider this PR ready to be merged.

dg0yt:
I still prefer to just disable that item, not to remove it from the dialog.

dl3sdo:
Kai, I'm fine with your approach of disabling instead of removing the item.
Please proceed with this PR.

PR #2029, opened 2022-01-20
ScaleRotateDialogs: Change notion 'georeferencing reference point'

dl3sdo:
In the 'Change map scale' dialog and in the 'Rotate map' dialog,
the scaling/rotation center can be choosen to be the 'Georeferencing reference point', even for local CRS.
This is confusing since the reference point is based on map coordinates (e.g., in mm) and not on geographic coordinates.
In the 'Options' sections below the term 'Adjust georeferencing reference point' also only refers to the
coordinates of the map reference point.
This commit implements Scott Turner's (@pkturner) proposal to use 'Map reference point' and 'Adjust coordinates of map reference point' instead.
It supersedes #2014


dl3sdo:
Rebase and split of previous commit to separate changes for each affected file.

dl3sdo:
@pkturner: It seems that the initial commit oversaw the StretchMapDialog dialog that uses the same wordings. Please check, thank you.

pkturner (approved):
This change in terminology from "georeferencing reference point" to "map reference point" is good all around.

I added related (though independent) issue #2310.

Issue #2012, opened 2021-12-22
FILEFORMAT_ASSERT() on text object export


lpechacek:
https://github.com/OpenOrienteering/mapper/blob/e247f7270fc36ac9c62abe2f179a891d1df98616/src/fileformats/ocd_file_export.cpp#L2622

At this moment, we merely know that it triggers but don't understand the exact circumstances. The reporter explicitly mentioned that the map contains text objects with varying alignment. However, it is unclear why `text_format_mapping` does not have the symbol+alignment combination.

lpechacek:
Unsurprisingly, changing `object`'s `h_align` attribute to an invalid value in an .omap file triggers the export assert. I don't know, however, if that is the root cause of the original report. I'm investigating further.

```
<?xml version="1.0" encoding="UTF-8"?>
<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="9">
<notes></notes>
<georeferencing scale="4000"><projected_crs id="Local"/></georeferencing>
<colors count="0">
</colors>
<barrier version="6" required="0.6.0">
<symbols count="1">
<symbol type="8" id="0" code="1" name="x"><text_symbol icon_text="" rotatable="true"><font family="Arial" size="4000"/><text color="-900" line_spacing="1" paragraph_spacing="0" character_spacing="0"/></text_symbol></symbol>
</symbols>
<parts count="1" current="0">
<part name="default part"><objects count="1">
<object type="4" symbol="0" h_align="3" v_align="2"><coords count="1">-82334 -48780;</coords><text>I like texting</text></object>
</objects></part>
</parts>
<templates count="0" first_front_template="0">
<defaults use_meters_per_pixel="true" meters_per_pixel="0" dpi="0" scale="0"/>
</templates>
<view>
<grid color="#80646464" display="0" alignment="0" additional_rotation="0" unit="1" h_spacing="500" v_spacing="500" h_offset="0" v_offset="0" snapping_enabled="true"/>
<map_view zoom="3.74293" position_x="-62348" position_y="-44393"><map opacity="1" visible="true"/><templates count="0"/></map_view>
</view>
</barrier>
</map>
```

lpechacek:
This is the murder weapon:
```
<object type="4" symbol="-4" h_align="0" v_align="0"><coords count="1">10720 -82640;</coords><text>b</text></object>
```

[...]

PR #2036, opened 2022-02-03
Random fixes in .ocd export

lpechacek:
Random fixes that were made during the analysis and fixing of #2012. Take them or leave them.

### DIFF
--- a/src/fileformats/ocd_file_export.cpp
+++ b/src/fileformats/ocd_file_export.cpp
@@ -1814,9 +1814,9 @@ void OcdFileExport::exportTextSymbol(OcdFile<Format>& file, const TextSymbol* te
 {
 	auto symbol_number = symbol_numbers.at(text_symbol);
 	
-	text_format_mapping.push_back({ text_symbol, TextObject::AlignLeft, 0, symbol_number });
-	text_format_mapping.push_back({ text_symbol, TextObject::AlignHCenter, 0, symbol_number });
-	text_format_mapping.push_back({ text_symbol, TextObject::AlignRight, 0, symbol_number });
+	for (auto alignment : {TextObject::AlignLeft, TextObject::AlignHCenter, TextObject::AlignRight})
+		text_format_mapping.push_back({ text_symbol, alignment, 0, symbol_number });
+
 	auto text_format = text_format_mapping.end() - 3;
 	auto count = [text_format](const auto* object) {
 		auto alignment = static_cast<const TextObject*>(object)->getHorizontalAlignment();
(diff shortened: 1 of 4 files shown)


dl3sdo:
@lpechacek: thank you for the fixes. At least the changes in OcdFileExport and Map make absolutely sense for me. I personally like the "for (auto alignment : {TextObject::AlignLeft, TextObject::AlignHCenter, TextObject::AlignRight})" loop. I'm not familiar with color palettes.

However, I encountered a number of crashes which I can not reliably reproduce (and thus not associate to your fix).
What I did:
- load the forest sample.ocd (which I attach here)
- add two text objects (one with left aligment, one with right alignment)
- try to save as another ocd12 file
- sometimes Mapper crashes and it seems that the crash sometimes happened while the file save dialog was still open.
[forest sample_hatched.zip](https://github.com/OpenOrienteering/mapper/files/8051223/forest.sample_hatched.zip)

dl3sdo:
Again a crash:
Gtk:ERROR:../../../../gtk/gtkrbtree.c:471:_gtk_rbtree_insert_after: assertion failed: (_gtk_rbtree_is_nil (tree->root))

Indeed, the crashes always occur while the save dialog is open (I just wait for some time without either saving or cancelling).
In addition I changed the autosave interval to 1min, which seems to have an effect.

dl3sdo:
The crash is actually independent of this commit, as it happens also on the current master branch.
That's how was reproducible at least for me:

1. set autosave interval to 1min
2. load the forest sample_hatched.ocd (which I attached here)
3. add two text objects (one with left aligment, one with right alignment)
4. go to 'Save as' and choose OCAD version 12
5. select one of the existing ocd files which then appears in the name field
6. activate the name field and observe how the cursor is blinking
7. suddenly it stops blinking
8. select another existing ocd file
9. sometimes Mapper then crashes, otherwise repeat from step 6

It might be unrelated to Mapper and related to QT instead.
OS: Ubuntu 20.04.3 LTS
Mapper runs in a Debug environment (but the crash happened also in Release environment).
However: after I switched from Debug to Release and back to Debug I cannot reproduce the crash.

Conclusion: these mysterious crashes shall not prevent from accepting these commits.

lpechacek:
Many thanks, @dl3sdo, for the review and testing! Also, the detailed replication instructions help a lot. I tried the same procedure on my Ubuntu 20.04.3 test box, but no crash occurred.

In any case, The hint about GTK suggests that the crash may not have been triggered by Mapper code directly, but it's instead something in GTK, Qt, or in their interaction. If you capture a full stack trace of the crash and core (generated with GDB), it might be interesting material for the Ubuntu developers.

src/fileformats/ocd_file_export.cpp
```diff
@@ -1814,9 +1814,9 @@ void OcdFileExport::exportTextSymbol(OcdFile<Format>& file, const TextSymbol* te
 {
 	auto symbol_number = symbol_numbers.at(text_symbol);
 	
-	text_format_mapping.push_back({ text_symbol, TextObject::AlignLeft, 0, symbol_number });
-	text_format_mapping.push_back({ text_symbol, TextObject::AlignHCenter, 0, symbol_number });
-	text_format_mapping.push_back({ text_symbol, TextObject::AlignRight, 0, symbol_number });
+	for (auto alignment : {TextObject::AlignLeft, TextObject::AlignHCenter, TextObject::AlignRight})
+		text_format_mapping.push_back({ text_symbol, alignment, 0, symbol_number });
+
```
dg0yt, line 1819: I'm unsure if this change is really beneficial. With the extra line, it is really harder to understand that `text_format_mapping.end() - 3` below refers to the first entry which was added here.

dg0yt (commented):
I really like the changes.

src/fileformats/ocd_file_export.cpp
lpechacek: The pull request comment has the "take them or leave them" suffix. I know that our views on coding do not align in many areas.

Here I made the change to express the fact that it is only the alignment changing in the format. Other parameters are constant. The previous form requires attentive reading and comparison of the three lines for differences to come to the same conclusion. Try that on Friday afternoon after a week of coding and maybe your stomach will tell you which form is more digestible.

Ad the `text_format` iterator, yes, it's a somewhat strange element here. Its initialization nicely aligns with its nature but I didn't tackle this bit.

src/fileformats/ocd_file_export.cpp
```diff
@@ -1838,35 +1838,24 @@ void OcdFileExport::exportTextSymbol(OcdFile<Format>& file, const TextSymbol* te
...
 	text_format = text_format_mapping.end() - 3;
 	if (text_format->count == 0)
 		text_format->alignment = TextObject::AlignHCenter;  // default if unused: centered
-	auto ocd_symbol = exportTextSymbol<typename Format::TextSymbol>(text_symbol, text_format->symbol_number, text_format->alignment);
-	FILEFORMAT_ASSERT(!ocd_symbol.isEmpty());
-	file.symbols().insert(ocd_symbol);
-	
+	export_single_symbol(text_format->symbol_number, text_format->alignment);
 	++text_format;
```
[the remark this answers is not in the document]
lpechacek: Likely yes. I saw `text_format` as the iterator pointing beyond the last element (which is common in C) while it's meant to point at the last element. In that sense, the increment would be part of the following for loop. I think that it demonstrates that I didn't get the `text_format` thingie or didn't manage to watch the details all the time. I had perhaps burned my energy on the code deduplication itself before reaching this point.

PR #2041, opened 2022-02-18
OcdFileExport: Fix map notes export to .ocd version 8

dl3sdo:
If the size of the map notes exceeded 32768 the map notes were truncated when saving as OCAD 8.
However, instead of truncating to 32767 (the terminating 0 will be appended) it was truncated to 23767.
When testing the fix with OCAD12 it turned out that it was necessary to truncate to 32766, otherwise OCAD12 would not open the map.

### DIFF
--- a/src/fileformats/ocd_file_export.cpp
+++ b/src/fileformats/ocd_file_export.cpp
@@ -810,11 +810,11 @@ void OcdFileExport::exportSetup(OcdFile<Ocd::FormatV8>& file)
 		if (!notes.isEmpty())
 		{
 			auto size = notes.size() + 1;
-			if (size > 32768)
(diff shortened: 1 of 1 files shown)


dl3sdo:
I attach a map (.omap) with 32776 bytes of map notes that I used for testing.
[MapNotes_32776.zip](https://github.com/OpenOrienteering/mapper/files/8097377/MapNotes_32776.zip)

dg0yt (approved):
LGTM. Not tested.

dl3sdo:
I modified the commit message to better meet the guidelines for commit messages.

dg0yt:
Thanks @dl3sdo !

PR #2042, opened 2022-02-20
Update README.md

mlerjen:
Update of the README to comply with the current IOF Specifications. Also adding ISSkiOM and ISMTBOM.

### DIFF
--- a/symbol sets/README.md
+++ b/symbol sets/README.md
@@ -4,18 +4,31 @@ The symbol sets to be edited are stored in the src subdirectory in the verbose
 CMake) an executable (test) `symbol_set_t` which updates compressed files from
 a list of known files in the src directory, thus automating the following steps.
 
-The ISOM 1:10000 symbol set can be created from the 1:15000 version by
-following these steps:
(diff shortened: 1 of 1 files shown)


dl3sdo:
Thank you for the pull request. I went through your proposed changes and made some corrections (e.g., fixing some typos and doing some rewording (I'm sorry but the word 'shear' is unknown to me in that context). You can find my changes of your changes here:
https://github.com/dl3sdo/mapper/tree/readme

pkturner:
'Shear' doesn't fit for me, either. I think it's a misspelling of 'sheer', meaning 'pure' or 'uncomplicated'.

mlerjen:
Yes. I ment sheer. But this pull request seems obsolete anyway. @dg0yt mentioned elseswhere that we only should define masters in the according base scales in the ./src directory while the version for other scales would be somehow generated automatically.

dg0yt (changes requested):
This PR has commits for unrelated aspects. I will split it for individual review and merge. We can have a talk how to avoid this next time.

dg0yt:
The README changes were merged already. The symbol set is the only change left.

PR #2043, opened 2022-02-22
OcdFileImport: Fix import of Rectangle symbols and objects

dl3sdo:
The 'GridFlags' field consists of 3 bits: 1=Grid On, 2=Numbering On, 4=Numbered from the bottom
The symbol import previously only considered Grid On and mapped 'Numbered from the bottom' to Bit 1 instead.
When importing Rectangle objects the 'Numbered from the bottom' property was applied just the other way around.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2013-2021 Kai Pastor
+ *    Copyright 2013-2022 Kai Pastor
  *
  *    Some parts taken from file_format_oc*d8{.h,_p.h,cpp} which are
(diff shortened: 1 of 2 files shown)


dl3sdo:
Please find attached an .ocd12 sample file and a screenshot how it appears in OCAD12:
OCAD_Example
[RectSymbols2_ocd12.zip](https://github.com/OpenOrienteering/mapper/files/8119990/RectSymbols2_ocd12.zip)

lpechacek:
Close/reopen to trigger CI run.

dg0yt (approved):
I didn't test myself, but the code changes LGTM.

dg0yt:
We should give names to the flags, but I don't want to deal with this now.
Thanks @dl3sdo.

PR #2045, opened 2022-03-02
OcdFileExport: Fix export of custom tabulator positions

dl3sdo:
When exporting more than 32 custom tabulator positions the num_tabs
field is now limited to the 32 tabs that are actually only exported and
a warning messages is issued (issue was detected by lpechacek).
In addition the export now converts the tabulator position to a 32 bit
value since the convertSize function that was used before converted it
to a 16 bit value instead (which was indeed only relevant for tab
values larger than 327,67mm).

### DIFF
--- a/src/fileformats/ocd_file_export.cpp
+++ b/src/fileformats/ocd_file_export.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2016-2021 Kai Pastor
+ *    Copyright 2016-2022 Kai Pastor
  *
  *    Some parts taken from file_format_oc*d8{.h,_p.h,cpp} which are
  *    Copyright 2012 Pete Curtis
@@ -1948,9 +1948,19 @@ void OcdFileExport::setupTextSymbolSpecial(const TextSymbol* text_symbol, OcdTex
 	ocd_text_special.line_below_width = decltype(ocd_text_special.line_below_width)(convertSize(qRound(1000 * text_symbol->getLineBelowWidth())));
 	ocd_text_special.line_below_offset = decltype(ocd_text_special.line_below_offset)(convertSize(qRound(1000 * text_symbol->getLineBelowDistance())));
 	
-	ocd_text_special.num_tabs = text_symbol->getNumCustomTabs();
-	auto last_tab = std::min(ocd_text_special.num_tabs, decltype(ocd_text_special.num_tabs)(std::extent<typename std::remove_pointer<decltype(ocd_text_special.tab_pos)>::type>::value));
-	for (auto i = 0u; i < last_tab; ++i)
+	auto const num_tabs = text_symbol->getNumCustomTabs();
+	auto const max_tabs = int(std::extent<typename std::remove_pointer<decltype(ocd_text_special.tab_pos)>::type>::value);
+	if (num_tabs <= max_tabs)
+	{
+		ocd_text_special.num_tabs = num_tabs;
+	}
+	else
+	{
+		ocd_text_special.num_tabs = max_tabs;
+		addWarning(::OpenOrienteering::OcdFileExport::tr("In text symbol %1: exporting only %2 custom tabulator positions")
+		           .arg(text_symbol->getPlainTextName()).arg(max_tabs));
+	}
+	for (auto i = 0u; i < ocd_text_special.num_tabs; ++i)
 		ocd_text_special.tab_pos[i] = convertSize(text_symbol->getCustomTab(i));
 }
 
--- a/src/fileformats/ocd_file_export.cpp
+++ b/src/fileformats/ocd_file_export.cpp
@@ -1961,7 +1961,7 @@ void OcdFileExport::setupTextSymbolSpecial(const TextSymbol* text_symbol, OcdTex
 		           .arg(text_symbol->getPlainTextName()).arg(max_tabs));
 	}
 	for (auto i = 0u; i < ocd_text_special.num_tabs; ++i)
-		ocd_text_special.tab_pos[i] = convertSize(text_symbol->getCustomTab(i));
+		ocd_text_special.tab_pos[i] = convertSize(qint64(text_symbol->getCustomTab(i)));
 }
 
 
--- a/src/fileformats/ocd_types_v8.h
+++ b/src/fileformats/ocd_types_v8.h
@@ -331,7 +331,7 @@ namespace Ocd
 		quint16 indent_first_line;
 		quint16 indent_other_lines;
 		quint16 num_tabs;
-		quint32 tab_pos[32];
+		qint32 tab_pos[32];
 		quint16 line_below_on;
 		quint16 line_below_color;
 		quint16 line_below_width;


dl3sdo:
Please find attached an .omap example that contains 34 custom tabulator positions (of which some exceed 327,67mm).
[ExportTextTabsSample_34TabsWithLargeValues.zip](https://github.com/OpenOrienteering/mapper/files/8172571/ExportTextTabsSample_34TabsWithLargeValues.zip)

dl3sdo:
@lpechacek: may I ask you again for a review since you brought this issue up?

src/fileformats/ocd_file_export.cpp
```diff
@@ -1949,9 +1949,15 @@ void OcdFileExport::setupTextSymbolSpecial(const TextSymbol* text_symbol, OcdTex
...
-		ocd_text_special.tab_pos[i] = convertSize(text_symbol->getCustomTab(i));
+	auto max_tabs = decltype(ocd_text_special.num_tabs)(std::extent<typename std::remove_pointer<decltype(ocd_text_special.tab_pos)>::type>::value);
+	if (ocd_text_special.num_tabs > max_tabs)
+	{
+		ocd_text_special.num_tabs = max_tabs;
+		addWarning(::OpenOrienteering::OcdFileExport::tr("In text symbol %1: exporting only %2 custom tabulator positions")
+		           .arg(text_symbol->getPlainTextName()).arg(max_tabs));
+	}
+	for (auto i = 0u; i < ocd_text_special.num_tabs; ++i)
+		ocd_text_special.tab_pos[i] = quint32((text_symbol->getCustomTab(i) + 5) / 10);
```
lpechacek, line 1960: Two comments on the change. First, the trivial one: every commit should change just one thing. Even if that's a single-line change, or even a single character. The presence of the phrase "in addition" in the commit message suggests that the commit contains multiple changes.

The first change is the `num_tabs` trimming. The second one is the `convertSize()` overload selection problem. I'm fine with the first change but the second tries to walk around a can of worms. I understand the motivation but when we are at it, the code should be cleaned up.

The trouble with `convertSize()` is multifaceted. Let me list my findings:
1) In the file format documentation, the tab positions are [LongInt](http://web.archive.org/web/20200728124010/http://docwiki.embarcadero.com/Libraries/Sydney/en/System.Longint) type. It corresponds to `qint32` in the Qt world. But the Mapper definition is an unsigned integer at `src/fileformats/ocd_types_v8.h:334`. That is the smaller issue.
2) `convertSize()` has two overloads and both of them convert the data type to a smaller one. I don't get the point. I'd expect the data type size to remain the same after division by 10. In any case, triggering the selection of the `qint64 -> qint32` function variant is a better course of action than open-coding the conversion.

My recommendation is, therefore, to split the patch into two - one with the addition of warning and the second addressing the type mess. HTH

lpechacek (changes requested):

src/fileformats/ocd_file_export.cpp
dl3sdo: Thank you for your review and proposals Libor.

I already stumbled across the issue with longint being mapped to quint32 in the ocd_types_v8.h (which does not matter due to the actual value range in this special case).
However this affects more mappings like LineSpace, IndentFirst or IndentOther which are SmallInt but mapped to quint16.

The convertSize() has two overloads but the convertSize(qint64 size) seems not to used at all.
Since the conversion from quint32 to qint32 is only needed once it seemed to me to be the easiest way to directly convert it.
AFAIK it's not possible to overload qint16 convertSize(qint32 size) by qint32 convertSize(qint32 size), since the input parameter/type pattern must be different. I don't know right know whether a convertSize((int64)..) cast would solve it, but that looks somehow weird.
OcdFileImport uses class functions and templates for the convertOcdString functionality. Would that be a path to follow?

lpechacek: My preference would be towards a function template as well, Matthias. And at the exact spot where we need 32-bit precision, I'd go with explicit instantiation (`convertSize<quint32>(text_symbol->getCustomTab(i))`).

However, @dg0yt can have a much better idea about how to eliminate the narrowing type conversion. Kai, any comments on this topic, please?

lpechacek:
Apologies for the long comment.

Thanks, Matthias, for the patches. Regarding the type conversion, I took a longer look at it and I can see that it's not exactly an easy topic. Here are my observations and a hint about the fix.

1) There are multiple `convert*()` functions in the ocd_file_export code. Two of them convert directly to the target type (`convertSize()` and `convertColor()`). The last one (`convertRotation()`) produces an integer and the type conversion is done through a cast.
2) It is only the custom tabs positions where the type don't match. You can identify the spot by wrapping the assignment arguments into curly braces to make the compiler warn about narrowing conversions (`sed -i 's/ = convert\(\(Color\|Size\|Rotation\).*\);/ = { convert\1 };/' src/fileformats/ocd_file_export.cpp`).
3) We already have inconsistent approaches to the type selection in the code. While your patch aligns the structure between the import and export code, it does not address the inconsistency among the usage of the conversion functions.

I can see large room for improvement. Here are my proposals what we can do next:

P1) Just make a minimal fix (use type cast for the overload selection: `convertSize(qint64(text_symbol->getCustomTab(i)));`), possibly fix the `tab_pos` type (should be `qint32`). This addresses the immediate problem and can remain the only action here.

P2) Rework the code to be explicit about the conversion. I can see two options.

P2.1) Provide template functions for `convert*()` and select the return type on each value conversion spot. The usage would be as follows:
`ocd_text_special.tab_pos[i] = convertSize<qint32>(text_symbol->getCustomTab(i));`
and at another place
`ocd_object.angle = convertRotation<decltype(ocd_object.angle)>(path->getPatternRotation());`
The drawback is the necessity of duplicating the target variable name in many cases because the compiler cannot infer the function return type in the assignment, to my knowledge. The benefit is that the assignments stay as they are.

P2.2) Provide template functions with an output parameter for the conversion. The usage would be:
`convertSize(text_symbol->getCustomTab(I), ocd_text_special.tab_pos[i]);`
and 
`convertRotationpath->getPatternRotation(), ocd_object.angle);`.
In this case, the compiler sees clearly the target type and there is little room for code-level inconsistency. The drawback is that the code might be more difficult to read without the explicit assignments.

With both P2.1 and P2.2 we can also perform run-time checks on the values and catch overflows.

My preference is towards P2.2, or doing P1+P2.2. While it may make the code look slightly weird, I believe that it will make it more robust. @dg0yt, @dl3sdo, opinions?

dg0yt:
I want to comment only briefly on the proposals:
- Focus this PR only on the fix of tab position export. This is Libor's "P1" IIUC.
- Don't turn free functions into (non-static) member function (unless needed externally).
- Don't add templates unless they really add value. 
  In the header, they are expensive for every usage of that header (or even pointless if only declared).
- With regard to https://github.com/OpenOrienteering/mapper/commit/11d447bbdb9a207be2b387fb23338297e94e3374: 
  Don't turn pure functions into "functions with output parameters" unless needed.

`convertSize` came from the legacy implementation of OCAD format 8. I guess the signature were designed for the needs at that time. 
What it does do right is to tag a certain type of conversion (unit of measurement). It also seems to do the right thing for most uses. There is nothing wrong with using explicit type conversions together with the function call in a few cases.

I would agree there is room for improvement. (At least they should go into an anonymous namespace, as an implementation detail internal to this translation unit.) But before jumping into templates functions, I would think about replacing the trivial integer types by struct types that signal that actual unit of measurement. Anything that helps the compiler to easily select the the proper conversion function. And there are not so many that we couldn't define them explicitly, without templates.

dg0yt (changes requested):
Focus this PR only on the fix of tab position export.

lpechacek:
> But before jumping into templates functions, I would think about replacing the trivial integer types by struct types that signal that actual unit of measurement. Anything that helps the compiler to easily select the proper conversion function. And there are not so many that we couldn't define them explicitly, without templates.

 Thanks for the comment! I like the proposal. I'll record the conclusion in the refactoring project.

@dl3sdo, please implement a minimal change for the correct `convertSize()` overload selection as hinted in P1. Adjust the commit message accordingly, i.e. drop the part about class methods. Then I think the change will be complete. Thanks!

dl3sdo:
Libor, Kai, thank you for your review and feedback. I fully agree to the proposal to just address and solve the current issue and to postpone a general rework.
We should keep in mind that there are other inconsistencies between .ocd types and their mapping to Mapper types (see https://github.com/OpenOrienteering/mapper/pull/2045#discussion_r820615157)

src/fileformats/ocd_file_export.cpp
```diff
@@ -1949,9 +1949,15 @@ void OcdFileExport::setupTextSymbolSpecial(const TextSymbol* text_symbol, OcdTex
...
-	auto last_tab = std::min(ocd_text_special.num_tabs, decltype(ocd_text_special.num_tabs)(std::extent<typename std::remove_pointer<decltype(ocd_text_special.tab_pos)>::type>::value));
-	for (auto i = 0u; i < last_tab; ++i)
-		ocd_text_special.tab_pos[i] = convertSize(text_symbol->getCustomTab(i));
+	auto max_tabs = decltype(ocd_text_special.num_tabs)(std::extent<typename std::remove_pointer<decltype(ocd_text_special.tab_pos)>::type>::value);
+	if (ocd_text_special.num_tabs > max_tabs)
+	{
+		ocd_text_special.num_tabs = max_tabs;
+		addWarning(::OpenOrienteering::OcdFileExport::tr("In text symbol %1: exporting only %2 custom tabulator positions")
+		           .arg(text_symbol->getPlainTextName()).arg(max_tabs));
+	}
```
dg0yt, line 1958: I was wondering why it looks so complicated, why I have difficulties to understand the type conversion implications.
As a general pattern with potential value domain mismatches, I would probably approach it like this:
~~~C++
auto const num_tabs = text_symbol->getNumCustomTabs();  //  original type retained: int
auto const max_tabs = std::extent<typename std::remove_pointer<decltype(ocd_text_special.tab_pos)>::type>::value);  // original type retained: std::size_t
if (num_tabs <= max_tabs)  // Safe enough here? Read below...
{
     	ocd_text_special.num_tabs = num_tabs  //  safe type conversion now
}
else
{
     	ocd_text_special.num_tabs = max_tabs  //  safe type conversion by data structure design
        addWarning(...)
}
for (...)
~~~

Pitfall:
The problem with `num_tabs <= max_tabs` is that you normally shouldn't mix signed and unsigned types: You may end up with negative values from the signed type being reinterpreted as (large) positive values for the unsigned type. (Cf. [example for Arithmetic comparison operators](https://en.cppreference.com/w/cpp/language/operator_comparison).)
Here, it shouldn't be an issue because both variables should only contain small non-negative values.
But there might still be a compiler warning. To resolve this issue, I would probably just wrap the `std::extent(...)` with `int(...)`.

lpechacek: > The problem with num_tabs <= max_tabs is that you normally shouldn't mix signed and unsigned types: You may end up with negative values from the signed type being reinterpreted as (large) positive values for the unsigned type. (Cf. [example for Arithmetic comparison operators](https://en.cppreference.com/w/cpp/language/operator_comparison).)

I'm glad that we repeat the C-language basics here for the new contributors. Yes, converting unsigned types to signed ones, or vice versa, and then attempting comparison or arithmetic on them leads to surprising effects. The reason, in my opinion, is that C-is a very low-level language that exposes the CPU features to the programmer and assumes that the programmer understands what he/she is working with. Rude people tend to say that C is not a programming language but rather a macro assembler. I consider this statement illustrative of the situation. 

We got to keep in mind that C++ maintains this tradition. Thankfully, the C++ creators try to give us tools to avoid such pitfalls, for example through the concept of narrowing conversion.

> Here, it shouldn't be an issue because both variables should only contain small non-negative values.

Ack. `num_tabs` will likely be non-negative. It would require a bit of effort (i.e. create over 65k custom tab positions) to make it overflow. `max_tabs` is fixed at 32.

> But there might still be a compiler warning. To resolve this issue, I would probably just wrap the std::extent(...) with int(...).

If we really had to fix the (almost-nonexistent) issue, how about adjusting the return type of `TextSymbol::getNumCustomTabs()`? I can see that the first narrowing conversion is happening right there.

I also suggest that we move the discussion elsewhere, say into the Discussions section or into a private developer-only thread to avoid scaring away our new contributors and prevent spamming the inboxes of people who are watching the GH project.

dg0yt: > > But there might still be a compiler warning. To resolve this issue, I would probably just wrap the std::extent(...) with int(...).
> 
> If we really had to fix the (almost-nonexistent) issue, how about adjusting the return type of `TextSymbol::getNumCustomTabs()`? I can see that the first narrowing conversion is happening right there.

I prefer to avoid unsigned integers.

> I also suggest that we move the discussion elsewhere, say into the Discussions section or into a private developer-only thread to avoid scaring away our new contributors and prevent spamming the inboxes of people who are watching the GH project.

I don't agree here. My comments are for this PR and for @dl3sdo. I would try to approach other contributor's differently.
You are effectively asking that such PRs would need to be made to my personal fork instead.

(FTR this is only part of my communication with @dl3sdo.)

lpechacek: > I don't agree here. My comments are for this PR and for @dl3sdo. I would try to approach other contributor's differently. You are effectively asking that such PRs would need to be made to my personal fork instead.

I was perhaps unclear in my message. I meant that there are 48 people watching this thread and only five of them contributed code in the last year. Given that this discussion shifted from the immediate bug fix towards clarification on the coding style, I thought we could stop spamming the remaining 43 people's inboxes with a topic that is truly developer-only.

> (FTR this is only part of my communication with @dl3sdo.)

That looks like another reason for taking the conversation aside. And if the message was intended only for Matthias, it truly helps to start the text with a salutation.

I'll refrain from further comments here. From my POV, Matthias did a good job here and the .ocd custom tab issue is resolved. Thanks!

lpechacek (approved):

dg0yt (approved):

PR #2055, opened 2022-03-29
OcdFileImport: Handle rotated objects with rotation-locked symbols

lpechacek:
This pull request embodies my ideas expressed in https://github.com/OpenOrienteering/mapper/pull/1975#issuecomment-890860795 and subsequent comments. In my opinion, we should maintain the symbol orientation bit whenever possible and issue warnings at places where it's not possible.


lpechacek:
> I noticed that you kept the import warning constant ("...an object with a rotated pattern"), but how about differentiation: "Removing rotation lock due to existence of a rotated text object." "Removing rotation lock due to existence of a rotated point object."

Matthis, the message is prefixed with a leader indicating the symbol type and symbol number in `addSymbolWarning()`. Therefore the whole import warning looks like this:

```
The map import generated warnings.
In point symbol 116.0 'Pit': Removing rotation lock due to existence of an object with a rotated pattern.
```

You still hit the nail on the head with the need for message adjustments. The messages uniformly talk about "patterns" but the word applies only to areas while with point objects and texts we need a different word.

lpechacek:
I've adjusted the texts and dropped the unrelated code cleanup from the patch.

image


Here comes a testing file if there is interest in trying out the change.
[ocad2018-rotated-non-rotatable-symbols.zip](https://github.com/OpenOrienteering/mapper/files/8659681/ocad2018-rotated-non-rotatable-symbols.zip)

PR #2056, opened 2022-04-02
Symbol: Fix symbol sorting by primary color

dl3sdo:
The result of sorting symbols by primary color changed for every
execution.
The 'colors' vector was claimed to contain "RGB values ordered by
color priority", however, the vector was unsorted.
Sorting the vector leads to a stable behaviour when sorting symbols
by primary color.

### DIFF
--- a/src/core/symbols/symbol.cpp
+++ b/src/core/symbols/symbol.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2022 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -956,6 +956,7 @@ Symbol::lessByColor::lessByColor(const Map* map)
 	colors.reserve(std::size_t(map->getNumColors()));
 	for (int i = 0; i < map->getNumColors(); ++i)
 		colors.push_back(QRgb(*map->getColor(i)));
+	sort(begin(colors), end(colors));
 }
 
 


lpechacek:
At first glance, I thought this is a straightforward fix to an existing issue. On a closer inspection, however, I can see that the symbol sorting machinery is much more complicated than a single call to `std::sort()`.

`Symbol::lessByColor::lessByColor(const Map* map)`, where you placed your fix, is a constructor for `struct Symbol::lessByColor`. The actual vector sorting is supposed to happen in `SymbolRenderWidget::sort(T compare)` and `Map::sortSymbols(T compare)`, however.

It seems that far more debugging effort will have to be invested into identifying the root cause.

dg0yt:
The documentation actually says that "this comparison uses the lowest priority which exists for a particular RGB color in the map, i.e. it groups colors by RGB values (and orders the result by priority)."

dg0yt:
> The result of sorting symbols by primary color changed for every
> execution.

@dl3sdo Is there a user-facing example, or is this an interpretation of "the functor becomes invalid as soon as colors are changed"?

It is not meant to sort by RGB. 

> `Symbol::lessByColor::lessByColor(const Map* map)`, where you placed your fix, is a constructor for `struct Symbol::lessByColor`. The actual vector sorting is supposed to happen in `SymbolRenderWidget::sort(T compare)` and `Map::sortSymbols(T compare)`, however.

Adding to that, `Symbol::lessByColor` a functor. Its main work is done in `operator Symbol::lessByColor::operator()(const Symbol* s1, const Symbol* s2)`, https://github.com/OpenOrienteering/mapper/blob/8597a440dce374a99d05df6423d4d77f4a5065e1/src/core/symbols/symbol.cpp#L962-L985

dl3sdo:
I think that the changing results are not surprising:
The current colors are stored unsorted in the 'color' vector when sorting starts. Then the colors are sorted against that snapshot.
The next time sorting is triggered the now different unsorted current colors are stored in the vector and sorting is done against this different snapshot.
Sorting the 'color' vector provides an equal snapshot and thus leads to equal results for subsequent sorting operations.

dg0yt:
Thanks for the screenshots!  It seems to do what it promises to do (primary order), with some quirks (secondary order).

- We only sort symbols here. The colors, and the order of the color list, remains unchanged.
- There is only operator "less than".
- > The current colors are stored unsorted in the 'color' vector when sorting starts.
 
  I think we start with a copy of what is already sorted by color priority.
- We failed to implement a "stable sort", i.e. if two symbols have the same primary color, their original order is not preserved.
  This should be fixed.
- For some symbols, the primary color is dominated by elements which are partially covererd by other elements (e.g. paved area with boundary line). 
  This might need some tweaking.

...

> I really don't understand the comment in the source code: "the functor becomes invalid as soon as colors are changed".

The functor takes a copy of the color list. This copy is invalid when the color list in the map is changed.

dl3sdo:
>* We failed to implement a "stable sort", i.e. if two symbols have the same primary color, their original order is not preserved.
This should be fixed.

I assume that we already use stable_sort (BTW: is it ok to define this in map.h instead in map.cpp?):
```
template<typename T> 
void Map::sortSymbols(T compare)
{
	std::stable_sort(symbols.begin(), symbols.end(), compare);
	// TODO: emit symbolChanged(pos, symbol); ? s/b same choice as for moveSymbol()
	setSymbolsDirty();
}
```
>* I think we start with a copy of what is already sorted by color priority.

I'm not sure about this, otherwise adding the 'sort' would have made no difference.

dg0yt:
> I assume that we already use stable_sort 

Looks like a good plan. But our operator `Symbol::lessByColor` didn't work properly. A test-first approach helped to create #2056.

> (BTW: is it ok to define this in map.h instead in map.cpp?):

`Map::sortSymbols` is a template function, a *generic* definition. It doesn't generate any *specific* code until it is used with a concrete type (instantiation). That's why the generic definition must be placed in the header in the standard case. Here, we don't risk "code bloat" because the function is small and used in very few places.

lpechacek:
And the first thing last, thanks for reporting the issue, @dl3sdo! 🙂

PR #2057, opened 2022-04-05
Fix symbol sorting by primary color

dg0yt:
"Sort by primary color" didn't return a stable result, but local order could vary with every execution.
This change adds a test for the functor which is responsible for sorting, and fixes that functor:
`'a < b` implies `!(b < a)`.
Replaces GH-2056.


dg0yt:
CC @dl3sdo

dl3sdo:
As expected, fix provided by @dg0yt provides stable results.

PR #2063, opened 2022-05-01
OcdParameterStreamReader: Documentation, more state

dg0yt:
@dl3sdo A few more changes which I couldn't finish before this afternoon. Would have been much harder without the unit test. 

Apart from documentation, this change ensures that `param_string.indexOf(QLatin1Char('\t') ...)` is called only once per tab character. All complexity is in `readNext()`.

### DIFF
--- a/src/fileformats/ocd_parameter_stream_reader.cpp
+++ b/src/fileformats/ocd_parameter_stream_reader.cpp
@@ -20,6 +20,8 @@
 
 #include "ocd_parameter_stream_reader.h"
 
+#include <cmath>
+
 #include <QtGlobal>
 #include <QChar>
 #include <QLatin1Char>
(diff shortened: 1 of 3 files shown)


dl3sdo:
I very much like your additional changes, adding the 'next' element makes use of synergy.
I noticed that in `return param_string.midRef(pos, next - pos);` the `next - pos` will become smaller than -1. Although that is not explicitly documented in QT's documentation, it works as if being -1, so using `qMax(next - pos, -1)` is unnecessary.

dg0yt:
> `return param_string.midRef(pos, next - pos);` the `next - pos` will become smaller than -1

Good catch. This wasn't intented. Qt only tests with `>= -1`, so we better don't rely on -2 being treated equally. 
https://code.woboq.org/qt5/qtbase/tests/auto/corelib/text/qstring/tst_qstring.cpp.html#_ZN11tst_QString6midRefEv

dl3sdo:
I went through the code again and it looks good to me, so I propose to merge it into master. Thank you for your rework Kai.

PR #2066, opened 2022-05-07
ObjectQueryParser: Object query fixes

dl3sdo:
There are two bugs in object_query.cpp:
- for special object queries an out_of_range access occurs which leads to a crash in the Debug version.
- the fromEscaped() functions fails for sequences of escaped characters (to verify the correct behaviour a test was added to ObjectQueryTest).

The last change is the removal of an unnecessary if condition.


dg0yt:
Mni tnx!

PR #2067, opened 2022-05-07
GdalSettingsPage: Delete 'Value' cell

dl3sdo:
When a duplicated 'Parameter' is entered the behaviour depends whether
the row is the last row or not.
If it's not the last row, then the complete row is removed, i.e. also
the 'Value' content is deleted.
If it's the last row then just the cell contents shall be deleted.
However, due to a copy-and-paste error only the 'Parameter' cell
content was deleted (twice).


dg0yt:
Tnx!

Issue #1004, opened 2017-11-08
Toggle colors // define colors as helping colors // draft mode


mlerjen:
**#background**
For better field mapping (avoid illegible maps/dense mapping) the use of modified (field)-symbols is a useful trick. 
![image](https://user-images.githubusercontent.com/15611653/32553538-7a7c5032-c497-11e7-882d-aff5798e7870.png) 
Modified symbols may contain additional information on the symbol as "minimal gap", "minimal lenght", "centre to centre radius" to be in accordance with certain screen symbols (f.ex. ISOM2017-212. fig.2) or to visualize the radius of an object oriented to north in case the map is about to be rotated due to shifts in declination. 
![image](https://user-images.githubusercontent.com/15611653/32554624-b71c0e30-c49a-11e7-947f-92fdf548ee80.png)

This feature request aims to create a workflow making it easy to switch back and forward between the "field map"//draft and the "final map".

**#feature request, two different options**
1) adding the option to toggle a color on/off  in the color window dialog. This would be the less sophisticated solution. The mapper would then be able to switch off all helper colors in this dialog.
2) adding the option to define a color as helper or draft color, similar to the definition of a "helper symbol". In consequence creating a "draft" mode and the possibility to toggle between "final map"- and "draft" - mode. Helper symbols and -colors only would be visible in the "draft"-mode.

Symbian9:
> create a workflow making it easy to switch back and forward between the "field map"//draft and the "final map".

Currently Mapper has few ways at this moment:
* By enabling "Baseline view" (F3) and/or "Hatch areas" (F2);
* Using "Symbols -> Load CRT file...": reqire custom `draft.omap` symbol set (need design it firstly) and custom CRT like `ISOM2017-draft.crt` & `draft-ISOM2017.crt`

But adding other way maybe would be better.

Zerbembasqwibo:
Is it possible to produce such help lines as a different symbol, and use "combined symbols" when in field? One adjustment is needed in OOM: if you turn off a symbol, it wil also be turned off where it's used as part of a combined symbol.

OK there will be very many to turn on and off, but in other suggestions discussion is going on to group symbols, maby then it will be easy to turn off all help symbols together.

ollesmaps:
I usually do not have problem to see when point/ line symbols are cluttered with my own eyes. With minimum areas that is a different story. A feature that notifies in real time when an area is bellow set minimum would be helpful.

Issue #2032, opened 2022-01-31
Updat ISOM2017-2 Symbol set


eolmapper:
Updat ISOM2017-2 Symbol set with 23.01.2022 errata.
[PDF  ISOM2017-2](https://1drv.ms/b/s!As4LDA11gDVmgt8CiJdF8MdoLshsLg)

Symbian9:
Duplicate of https://github.com/OpenOrienteering/mapper/issues/1037

Symbian9:
Close it.

Symbian9:
Related:

- https://github.com/OpenOrienteering/mapper/issues?q=is%3Aissue+is%3Aopen+ISOM
  - https://github.com/OpenOrienteering/mapper/issues/1184

eolmapper:
Errata is the word used in the specification in the English version.
I also find that the mapp comission has a chaotic management of these updates, but it's not my fault.

Symbian9:
>  but it's not my fault.

My previous [comment](https://github.com/OpenOrienteering/mapper/issues/2032#issuecomment-1027754015) directed to @krticka (IOF MC member), not to you.

Symbian9:
> Errata is the word used in the specification in the English version.

"Errata" and "Updat" are not English words we use on issues tracker.

As I see Errata even not a common word, as it also *translated* as "changes"

![Screenshot_2022-02-02_12-24-25](https://user-images.githubusercontent.com/13188580/152136139-ad3a2874-6335-4b4a-ae50-a7b1edae33e2.png)

pkturner:
That's an interesting quotation from the Chicago Manual of Style. It arises from the context of paper publishing, and the quoted description is very different from a "changelog". By the way, I have seen the word "errata" many times in English publications.

An international committee has a process it must abide by. It's not surprising that they would stretch the meaning of _errata_ to include fixes that are made two years after publication.

mlerjen:
Here is an updated version with minor corrections on 103, 309, 416, 418 and 507.

[ISOM v4.2022_15000-2.zip](https://github.com/OpenOrienteering/mapper/files/9392448/ISOM.v4.2022_15000-2.zip)

mlerjen:
[ISOM 2017-R5_15000.zip](https://github.com/OpenOrienteering/mapper/files/11229926/ISOM.2017-R5_15000.zip)

bujke018:
[...]

PR #2084, opened 2022-08-22
MeasureWidget: Warn about min size in combined symbols

lpechacek:
This enhancement is especially useful when working with @mlerjen's "halo" symbol set (mentioned multiple times, e.g. in #1004). Please consider merging.

### DIFF
--- a/src/gui/widgets/measure_widget.cpp
+++ b/src/gui/widgets/measure_widget.cpp
@@ -1,6 +1,7 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
  *    Copyright 2012-2015 Kai Pastor
+ *    Copyright 2022 Libor Pecháček
  *
  *    This file is part of OpenOrienteering.
  *
@@ -21,6 +22,8 @@
 
 #include "measure_widget.h"
 
+#include <queue>
+
 #include <QLocale>
 #include <QScroller>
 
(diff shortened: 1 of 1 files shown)


lpechacek:
> @mlerjen's "halo" symbol set (mentioned multiple times, e.g. in #1004)

I've just realized that it's actually the symbol set contributed in #2032.

dl3sdo (changes requested):
Libor, thank you for considering combined area symbols. Your commit works fine except in one case: if a new symbol part is just added to the combined symbol  without specifying its type, Mapper will crash due to dereferencing a nullptr. Adding an additional check solves this issue:
`
if (part)
{
	if (part->getType() == Symbol::Area)
	{
		auto symbol_min_area = 0.001 * part->asArea()->getMinimumArea();
		if (symbol_min_area < minimum_area)
			minimum_area = symbol_min_area;
	}
	else if (part->getType() == Symbol::Combined)
	{
		to_be_examined.push(part->asCombined());
	}
}
`
See https://github.com/dl3sdo/mapper/commit/c1b9ed828e863fc66ece0e73e633eb88530efbe5

lpechacek:
> Libor, thank you for considering combined area symbols. Your commit works fine except in one case: if a new symbol part is just added to the combined symbol without specifying its type, Mapper will crash due to dereferencing a nullptr. Adding an additional check solves this issue: `if (part) { if (part->getType() == Symbol::Area) ........ }` 

Thanks, Matthias, for testing the commit and the fix! I agree with the approach, however, I'd avoid the additional indentation level induced by `if (part) ...` but rather skip the part in the loop with a `continue`. Otherwise, I agree with all the changes and I've merged them into my commit.

dl3sdo (approved):

PR #2088, opened 2022-09-10
GeoreferencingDialog: Fix declination lookup

### LINKED ISSUE #2087: Magnetic declination lookup button does not work (NOAA added a key)
### Steps to reproduce
1. New file with georeferencing UTM 32N 44.6641 10.7893
2. Declination lookup button
3. Browser result: Bad request, check your key

### Configuration
Mapper Version: 0.9.5

NOAA added a key for the request to be valid so the Mapper path is no longer OK

viewing the html page source of 
https://www.ngdc.noaa.gov/geomag/calculators/declinationForm.shtml
I found this variable to be added to the path
key=zNEw7

so using the previous coordinates:

PATH coming from Mapper (not working anymore):
https://www.ngdc.noaa.gov/geomag-web/calculators/calculateDeclination?lat1=44.6641&lon1=10.7893&startYear=2022&startMonth=9&startDay=9&resultFormat=html

PATH with the same coordinates and the new key -  that is actually working:
https://www.ngdc.noaa.gov/geomag-web/calculators/calculateDeclination?key=zNEw7&lat1=44.6641&lon1=10.7893&startYear=2022&startMonth=9&startDay=9&resultFormat=html

On their page https://www.ngdc.noaa.gov/geomag/calculators/magcalc.shtml  it is written
"Registration is required to access this service. Please register using the API registration link on the top right"
But right now it is possible to use the service without registration and this key=zNEw7 comes out automatically when you push the Calculate button

I tried to get all info I could and I hope this is useful and can give a hint to start debugging, but bear with me I'm no expert so I have no idea if the key is fixed or changing or whatever :-) and even if adding this key to our path is allowed.

dl3sdo:
The declination lookup at www.ngdc.noaa.gov now requests an additional parameter 'key'.
Closes #2087

### DIFF
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -34,6 +34,7 @@
 #include <QDesktopServices>  // IWYU pragma: keep
 #include <QDialogButtonBox>
 #include <QDoubleSpinBox>
+#include <QInputDialog>
 #include <QFlags>
 #include <QFormLayout>
 #include <QHBoxLayout>
@@ -435,20 +436,21 @@ void GeoreferencingDialog::requestDeclination(bool no_confirm)
 	QUrl service_url(user_url + QLatin1String("calculators/calculateDeclination"));
 	LatLon latlon(georef->getGeographicRefPoint());
 	
+	QString key;
 	if (!no_confirm)
 	{
-		int result = QMessageBox::question(this, tr("Online declination lookup"),
-		  tr("The magnetic declination for the reference point %1° %2° will now be retrieved from <a href=\"%3\">%3</a>. Do you want to continue?").
-		    arg(latlon.latitude()).arg(latlon.longitude()).arg(user_url),
-		  QMessageBox::Yes | QMessageBox::No,
-		  QMessageBox::Yes );
-		if (result != QMessageBox::Yes)
+		bool ok;
+		key = QInputDialog::getText(this, tr("Online declination lookup"),
+									tr("The magnetic declination for the reference point %1° %2°<br>will now be retrieved from <a href=\"%3\">%3</a>.<br><br>Enter access key:").
+									  arg(latlon.latitude()).arg(latlon.longitude()).arg(user_url),
+									QLineEdit::Normal, QString{}, &ok);
+		if (!ok || key.isEmpty())
 			return;
 	}
 	
 	QUrlQuery query;
 	QDate today = QDate::currentDate();
-	query.addQueryItem(QString::fromLatin1("key"), QString::fromLatin1("zNEw7"));
+	query.addQueryItem(QString::fromLatin1("key"), key);
 	query.addQueryItem(QString::fromLatin1("lat1"), QString::number(latlon.latitude()));
 	query.addQueryItem(QString::fromLatin1("lon1"), QString::number(latlon.longitude()));
 	query.addQueryItem(QString::fromLatin1("startYear"), QString::number(today.year()));
@@ -836,7 +838,7 @@ void GeoreferencingDialog::declinationReplyFinished(QNetworkReply* reply)
 		QMessageBox::Retry | QMessageBox::Close,
 		QMessageBox::Close );
 	if (result == QMessageBox::Retry)
-		requestDeclination(true);
+		requestDeclination(false);	//  NOTE: if a wrong key was entered it makes no sense to repeat without having the chance to enter a new one
 #else
 	Q_UNUSED(reply)
 #endif
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2022 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -448,6 +448,7 @@ void GeoreferencingDialog::requestDeclination(bool no_confirm)
 	
 	QUrlQuery query;
 	QDate today = QDate::currentDate();
+	query.addQueryItem(QString::fromLatin1("key"), QString::fromLatin1("zNEw7"));
 	query.addQueryItem(QString::fromLatin1("lat1"), QString::number(latlon.latitude()));
 	query.addQueryItem(QString::fromLatin1("lon1"), QString::number(latlon.longitude()));
 	query.addQueryItem(QString::fromLatin1("startYear"), QString::number(today.year()));
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -34,7 +34,6 @@
 #include <QDesktopServices>  // IWYU pragma: keep
 #include <QDialogButtonBox>
 #include <QDoubleSpinBox>
-#include <QInputDialog>
 #include <QFlags>
 #include <QFormLayout>
 #include <QHBoxLayout>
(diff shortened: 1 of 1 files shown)


src/gui/georeferencing_dialog.cpp
```diff
@@ -448,6 +448,7 @@ void GeoreferencingDialog::requestDeclination(bool no_confirm)
 	
 	QUrlQuery query;
 	QDate today = QDate::currentDate();
+	query.addQueryItem(QString::fromLatin1("key"), QString::fromLatin1("zNEw7"));
```
dg0yt, line 451: We cannot hardcode a key of unknown ownership. Either we get a key for our app, or the user needs to provide an individual key.

dl3sdo: As we have not yet got a key for Mapper I implemented the necessity for the user to enter a key.

lpechacek: > As we have not yet got a key for Mapper I implemented the necessity for the user to enter a key. I intentionally didn't clean up the code by removing the 'no_confirm' parameter.

I checked with the NOAA admins about the use of the key and they confirmed that we can register Mapper as an application in their system and put the access key into the sources. I've been warned that the key might expire. If that happens, I'll receive a notification about the event.

So, I've just filled in the registration form and got the key. Incidentally, it's the same string as the one above. I think we can merge the first patch in this pull request then.

As a convenience extra, there can be a mechanism to ask the user for an alternate key in case the current key expires and the lookups start failing.

dl3sdo: @lpechacek: Thank you for getting in touch with NOAA.

I propose not to implement the mechanism to ask the user for an alternate key but instead add the key to the settings. Thus, when the key changes the user has to enter the new key only one time in the settings and is done (although declination lookup is something that won't be frequently used by a single user).

What is your opinion?

lpechacek: Sounds good to me in general. However, if we build the default key into the application and the key is in the settings, at the time it changes, we will have to update the settings. That is, we might have to discern between the default application key and the user provided key going forward. That feels like a complex task to me but if you like the exercise, go ahead.

As for me, I'd rather keep our (code contributors') life simple, build the key into the application, and _possibly_ give the user an option to provide an alternate key in case the current one is expired and the user has not yet upgraded the app. That is, the key would always be stored in the app and the key update would be bound to an app update. If key update turn out to be frequent event, we can adopt a more complicated mechanism.

pkturner: > the user has to enter the new key only one time in the settings and is done

As @lpechacek notes, once the user has updated a setting, any further updates would be the user's responsibility.

> keep our (code contributors') life simple, build the key into the application

Wouldn't this put pressure on the maintainers to create a new release any time the hard-coded key expires?

It would be preferable if a world magnetic model were built in to Mapper, because the Internet service that Mapper has used in the past is susceptible to breaking change. A brief search turned up some closely related public domain software for the World Magnetic Model (WMM), but no library/API that would be convenient to incorporate in Mapper. The WMM appears to be less than half as accurate as the service that Mapper has used in the past. In my opinion the WMM is accurate enough to provide a starting point when first georeferencing a map. Perhaps there is some way that the WMM could be bundled with Mapper.

lpechacek: Thanks, Scott, for sharing your views. Before I respond, let me formulate my position. I've spent a good part of my life with infrastructure software maintenance. That are usually the things that users don't notice and very often either take them for granted or don't recognize their existence. From that point of view, improving the operation of the cut tool is much more important to me than improving the declination lookup. I use the declination lookup only once on a map project while I use the cut tool several hundred times. I hope that illustrates my priorities and clarifies my starting point.

> Wouldn't this put pressure on the maintainers to create a new release any time the hard-coded key expires?

Of course, it would. My assumption is that the key updates will be infrequent. If the assumption proves wrong, we can add code.

> It would be preferable if a world magnetic model were built in to Mapper, because the Internet service that Mapper has used in the past is susceptible to breaking change. A brief search turned up some closely related public domain software for the World Magnetic Model (WMM), but no library/API that would be convenient to incorporate in Mapper. The WMM appears to be less than half as accurate as the service that Mapper has used in the past. In my opinion the WMM is accurate enough to provide a starting point when first georeferencing a map. Perhaps there is some way that the WMM could be bundled with Mapper.

Ummm, when I hear about merging third-party code into Mapper just to save the user three copy-pastes on a project, I'm feeling shivers along my spine. As I wrote earlier, the declination lookup is done only once. And it's fairly easy to fill the numbers into the online calculator and copy the resulting declination back to Mapper. Putting coding effort into improving this operation does not sound to **me** (mind I'm focusing on infrastructure and frequently repeated operations) like a good investment. I see other areas where a similar amount of effort would produce greater value:
* [New Clipper](http://www.angusj.com/clipper2/Docs/Overview.htm) has been released. This library is behind the cool boolean operations in Mapper. Clipper is now version 2, has a different API, and can do [parallel moves on lines](http://www.angusj.com/clipper2/Docs/Units/Clipper.Offset/Classes/ClipperOffset/_Body.htm). The parallel moves can be perhaps leveraged in the Move parallel tool.
* The cut tool on areas can start only at an area border and end at another place on a border. It's not possible to start the cut outside of the border but it's possible to finish the cut in the wrong place. If that happens Mapper just says that the cut was wrong and wastes all the effort invested into the cut drawing. That feels somewhat rude to me, so I propose that Mapper does not allow finishing a cut in the wrong place. Someone also told me that he would appreciate if cuts cut start and end anywhere. I'm not sure how it would work but I'm sharing it here in case it makes sense to anyone else.
* OCAD corner points are not as stupid an idea as I thought originally. While on many dashed lines we can easily do with dash points, the ruined fence symbol remains a challenge. Mapper places the mid symbol at the corner, which is not exactly what I would like to see. Also, regular fences look somewhat cleaner with the mid-symbols further from the corner than we currently have. Implementation of the corner point paradigm (a point that has full-length symbol segments on each side as if the line was cut in two at the place) seems desirable to me.
* Highlight which object will be followed when following a line during drawing. My experience shows that always the wrong object is selected but I can see it only later during drawing which is annoying.

Mind that my proposal is mere **my** proposal and I'm not pushing you in that direction. I'm sharing this information here to illustrate my way of thinking about the software. Please also mind that I'm a UNIX and Linux user for over a quarter century. Do one thing and do it well is my favorite mantra. :)

pkturner: @lpechacek Your comments are appreciated, particularly in regard to directing effort into what is valuable to users.

I love those suggestions for features/areas to work on.

dg0yt: Geomagnetic field calculations are supported by GeographicLib, https://sourceforge.net/projects/geographiclib/.

lpechacek: > @lpechacek Your comments are appreciated, particularly in regard to directing effort into what is valuable to users.

Thanks for the positive feedback, Scott. Let me make one thing clear. I'm by no means saying what is valuable to **users** here. I'm trying to clarify what is **my** priority and why in the hope that it may help you understand my comments.

To be fair, I think I read a comment from Kai that Mapper is supposed to focus on beginners. If that is true, a smooth magnetic declination lookup dialog is a key feature to attract newbies and counts as end-user value. I'm, however, concerned with power users and tend to let the newbies do their homework on their own. I'm not sure if my tendency ever changes but if it does, I'll let you know.

ollesmaps: > * The cut tool on areas can start only at an area border and end at another place on a border. It's not possible to start the cut outside of the border but it's possible to finish the cut in the wrong place. If that happens Mapper just says that the cut was wrong and wastes all the effort invested into the cut drawing. That feels somewhat rude to me, so I propose that Mapper does not allow finishing a cut in the wrong place. Someone also told me that he would appreciate if cuts cut start and end anywhere. I'm not sure how it would work but I'm sharing it here in case it makes sense to anyone else.

It makes sense! Could it be a separate ticket?
I would suggest that cuts start and end (if not on perimeter) at the nearest perimeter point.

lpechacek: > > * The cut tool on areas can start only at an area border and end at another place on a border. It's not possible to start the cut outside of the border but it's possible to finish the cut in the wrong place. If that happens Mapper just says that the cut was wrong and wastes all the effort invested into the cut drawing. That feels somewhat rude to me, so I propose that Mapper does not allow finishing a cut in the wrong place. Someone also told me that he would appreciate if cuts cut start and end anywhere. I'm not sure how it would work but I'm sharing it here in case it makes sense to anyone else.
> 
> It makes sense! Could it be a separate ticket?

Done.

dl3sdo:
Closed in favor of #2344

PR #2091, opened 2022-09-14
PointSymbolEditorWidget: Fix and refactor signal blocking

dl3sdo:
Fix an obvious copy&paste error (duplicate usage of point_outer_width_edit instead of point_outer_color_edit).
Use QSignalBlocker and ScopedMultiSignalsBlocker classes to block signals.

PR #2093, opened 2022-09-18
PointSymbolEditorWidget: Don't duplicate CurveStart property

### LINKED ISSUE #2090: Symbol Editor: Cannot remove coordinate pairs
### Steps to reproduce
1. Defining a point symbol with a line element
2. Adding coordinate pairs
3. Want to remove an unnecessary coordinate pair

### Actual behaviour
At certain coordinate pairs the minus-button is grayed out, for others it is not.
image

### Expected behaviour
Minus button should be always on IMO.

### Configuration
Mapper Version: Master v.20220316.2
Operating System: Windows7

dl3sdo:
When adding a path point by clicking on the '+' button either a default point is appended to the path or an existing path point is duplicated.
However, when duplicating an existing point, also the CurveStart property was duplicated which lead to two successive path points with the CurveStart property being set although the next two path points following a CurveStart path point define the course of the curve and thus can't be a CurveStart path point itself.

Closes #2090 (Symbol Editor: Cannot remove coordinate pairs).

Additional fixes:
- When adding a path point by the '+' button the symbol was not updated. Since the point is normally duplicated this had no visual impact. However, if no row is selected then a default point at (0,0) is created which was not shown in this case.
- When the 'Line closed' checkmark was set and all lines were removed the checkbox remained set but became disabled.
- If a coordinate was changed to a curve start while a coordinate below was selected that would thus become part of the curve, then the button to delete a coordinate was not disabled.
- The 'Center all elements' button is initially disabled if the point symbol contains no elements. The button is enabled after adding an element but was not disabled after deleting all elements.

### DIFF
--- a/code-check-wrapper.sh
+++ b/code-check-wrapper.sh
@@ -85,6 +85,7 @@ for I in \
   overriding_shortcut.cpp \
   paint_on_template \
   point_symbol.cpp \
+  point_symbol_editor_widget.cpp \
   print_widget.cpp \
   renderable.cpp \
   renderable_implementation.cpp \
--- a/src/gui/symbols/point_symbol_editor_widget.cpp
+++ b/src/gui/symbols/point_symbol_editor_widget.cpp
@@ -857,12 +857,10 @@ void PointSymbolEditorWidget::centerCoordsClicked()
 void PointSymbolEditorWidget::updateCoordsTable()
 {
 	auto* object = getCurrentElementObject();
-	int num_rows;
-	if (object->getType() == Object::Point)
-		num_rows = 1;
-	else
+	int num_rows = 1;
+	if (object->getType() == Object::Path)
 	{
-		auto* path = static_cast<PathObject*>(object);
+		auto* path = static_cast<const PathObject*>(object);
 		num_rows = int(path->getCoordinateCount());
 		if (num_rows > 0 && path->parts().front().isClosed())
 			--num_rows;
--- a/src/gui/symbols/point_symbol_editor_widget.cpp
+++ b/src/gui/symbols/point_symbol_editor_widget.cpp
@@ -844,7 +844,11 @@ void PointSymbolEditorWidget::updateCoordsTable()
 		if (num_rows > 0 && path->parts().front().isClosed())
 			--num_rows;
 		if (path->getSymbol()->getType() == Symbol::Line)
+		{
+			const QSignalBlocker blocker(line_closed_check);
+			line_closed_check->setChecked(num_rows > 0 && path->parts().front().isClosed());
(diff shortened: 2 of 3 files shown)


dl3sdo:
Waiting for #2091 to be merged.

dg0yt:
Working on it...

dl3sdo:
I notice that you change the behaviour when deleting a curve point:
Previously, when deleting a curve point, the point and the points defining the curve handles were deleted. Now, only the curve property is removed which allows to remove also curve handle points which results in removing the curve property from the related point above.
There is one flaw: for the minimum curve (e.g. in symbol 708), when deleting any of the first three points, the curve property is removed and only the point is deleted. However, when deleting the last point, then all four points are removed.

I also noticed that when there are only two points left, then the right '-' remains disabled and the user has to remove the remaining minimum line with the left '-'. The same logic applies to areas: the minimum area can only be deleted which the left '-'.

dg0yt:
Well, this part of the symbol editors wasn't complete in the past, and it isn't complete now. I pushed what I had at this moment, for testing/review. I do have some unfinished changes.
 - If there is a table, the `[+]` and `[-]` buttons should maintain table rows IMO, i.e. affect a single row.
   However, this low-level approach means a little bit of fighting against the high-level behavior of `PathObject::deleteCoordinate` which is geared towards the normal graphical editing.
 - "when deleting the last point, then all four points are removed."  
   This is not desired.
 - Having to remove a whole element in order to remove the remaining coordinates makes sense IMO.  
   The incomplete part is making adding an element also add the minimum number of coordinates.

dg0yt:
> * "when deleting the last point, then all four points are removed."
>   This is not desired.

Resolved.

> * Having to remove a whole element in order to remove the remaining coordinates makes sense IMO.
>   The incomplete part is making adding an element also add the minimum number of coordinates.

Resolved.

dl3sdo:
With the latest changes the '+' button will add three empty coordinate points for an empty area symbol (and two for an empty line). The user however is able to add just a single point by clicking into the right drawing area. Apparently nothing bad happens.
It's questionable whether Mapper should enforce at least subsymbols with two or three coordinate points by disabling the OK button otherwise.

In general all issues have been resolved.
Kai, thank you for the rework of this PR.

dl3sdo:
@dg0yt: I followed @mlerjen's proposal to mark the curve points. Feel free to use or drop my commit. As an alternative a different brush could be used for marking I guess.

dg0yt:
This is what I did now. More visible than just italics, and some drive-by improvements.
screenshot
CC @mlerjen

dl3sdo:
I tested the current state and it looks good to me.

Issue #982, opened 2017-10-05
Involuntarily adding stuff on map


ollesmaps:
### Steps to reproduce
It is very easy involuntarily add stuff on map (Android app). You can touch screen by accident or forget that you are in drawing or scriblling mode. And voilà there are objects on map you are not aware of. It is hard to spot these mistakes. (In "manual" you could always go back to mylar and see the original. You can not do that in the app.)

### Expected behaviour
So there is a strong need to avoid mistakes being made. I believe there are more ways how to do it. My suggestion would be:
After some time of no action (10 secs?) automatically switch buttons to Pan map button. This way you omit chances of accidentally adding stuff on map (the time might be optional).

### Configuration
Android App 0.7.0

Zerbembasqwibo:
I have exactly same thoughts. But, maybe better, or another way, to have some kind of check compared to you GPS-position (if used). That means you will be given some kind of warning if you try to draw an object more than X meter from your position.

The "switch to pan-mode" idea is anyway maybe easier and best way? I think 10 sec might be good, but also maybe after drawing point object time could be much shorter or zero? If you accidentally draw line/area object you will probalby see it?

AskoMaatta:
When you select a symbol in OOMapper (Touch Mode User Interface), it has a move function by default on. It's a bit tricky, because you can accidentally use a pen to move, for example, the contour or some area symbols.
It would be great if the move and delete functions could be protected and they could be released in some clever way

Asko M

Zerbembasqwibo:
Also, would be nice to have a warning, if you try to edit something more
far away than X meter (50 default, set by user) from your GPS position.

Purpose; to prevent editing by mistake at field work.


lördag 24 september 2022 skrev AskoMaatta ***@***.***>:
> When you select a symbol in OOMapper (Touch Mode User Interface), it has
a move function by default on. It's a bit tricky, because you can
accidentally use a pen to move, for example, the contour or some area
symbols.
> It would be great if the move and delete functions could be protected and
they could be released in some clever way
>
> Asko M
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you commented.<
[...]

PR #2097, opened 2022-09-24
MapEditorController: Select pan tool on touch UI startup

lpechacek:
The edit mode is selected by default on editor startup. However, in the touch UI, object editing is rarely the first action to be done after opening a map. The user typically pans the map view to the area of interest first and then starts editing.

This code change makes the touch UI start in the view pan mode. The desktop action remains to be editing as the desktop users use mouse middle button for editing.

Discussed in GH-982.


dg0yt:
Thanks!

PR #2098, opened 2022-10-02
LineSymbolSettings: Fix enabling/disabling in line symbol editor

dl3sdo:
The 'mid symbol distance' field and label were not correctly disabled and enabled.
Enabling them by increasing 'Mid symbols per spot' was not possible unless reentering the symbols setting dialog. If they were enabled, removing the mid symbol would not disable them.
Adding both to `mid_symbol_widget_list` solves the issue.

Alternatively it would have been possible to replace in `mid_symbol_distance_edit->setEnabled(mid_symbol_distance_edit->isEnabled() &&` ...
by
`mid_symbol_distance_edit->setEnabled(!symbol->mid_symbol->isEmpty() &&` ...

### DIFF
--- a/src/gui/symbols/line_symbol_settings.cpp
+++ b/src/gui/symbols/line_symbol_settings.cpp
@@ -254,7 +254,6 @@ LineSymbolSettings::LineSymbolSettings(LineSymbol* symbol, SymbolSettingDialog* 
 	mid_symbol_widget_list = {
 	    mid_symbol_placement_label, mid_symbol_placement_combo,
 	    mid_symbol_per_spot_label, mid_symbol_per_spot_edit,
-	    mid_symbol_distance_label, mid_symbol_distance_edit,
 	    show_at_least_one_symbol_check,
 	};
 	
--- a/src/gui/symbols/line_symbol_settings.cpp
+++ b/src/gui/symbols/line_symbol_settings.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2018 Kai Pastor
+ *    Copyright 2012-2019, 2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -40,9 +40,9 @@
 #include <QVariant>
 #include <QWidget>
 
-#include "core/symbols/symbol.h"
 #include "core/symbols/line_symbol.h"
 #include "core/symbols/point_symbol.h"
+#include "core/symbols/symbol.h"
 #include "gui/util_gui.h"
 #include "gui/symbols/point_symbol_editor_widget.h"
 #include "gui/symbols/symbol_setting_dialog.h"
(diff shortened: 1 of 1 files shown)


lpechacek:
It took me some time to study the UI code mechanisms and still I don't understand it in full. Anyway, your code change appears relevant and correct to me.

The `mid_symbol_distance_`* components were removed from `mid_symbol_widget_list` in commit 1f7930a1f10d. There is no compelling reason for the removal given in the commit message and also the code change does not provide an explanation on its own. So, I'd say that the bug is a result of an editing error.

That said, I advise for the following **commit message** changes:
* Mention commit 1f7930a1f10d as the source of the error and declare that this commit restores the original state. A simple "Amends 1f7930a1f10d." with some explanation that we are going back in the history is enough.
* Remove the discussion about alternative fixes from the commit message. Discussion about alternatives should be part of the cover letter (pull request message here at GH) but not the commit itself.

Ad the alternatives, the code would really benefit from further cleanup. Removal of the copy-paste in mid-symbol distance edit might be relevant:
```
	auto const mid_symbol_distance_enabled = !symbol->mid_symbol->isEmpty() && symbol->mid_symbols_per_spot > 1;
	mid_symbol_distance_label->setEnabled(mid_symbol_distance_enabled);
	mid_symbol_distance_edit->setEnabled(mid_symbol_distance_enabled);
```
However, the cleanup should be part of a new commit if we ever want to perform the beautification.

Thanks!

lpechacek (approved):

dl3sdo:
Rebased and squashed. I consider this PR ready to be merged.

src/gui/symbols/line_symbol_settings.cpp
```diff
@@ -252,6 +254,7 @@ LineSymbolSettings::LineSymbolSettings(LineSymbol* symbol, SymbolSettingDialog*
 	mid_symbol_widget_list = {
 	    mid_symbol_placement_label, mid_symbol_placement_combo,
 	    mid_symbol_per_spot_label, mid_symbol_per_spot_edit,
+	    mid_symbol_distance_label, mid_symbol_distance_edit,
```
dg0yt, line 257:
```suggestion
```
This list is for label having the *same* condition ...

src/gui/symbols/line_symbol_settings.cpp
```diff
@@ -717,8 +720,9 @@ void LineSymbolSettings::updateStates()
 	{
 		mid_symbol_widget->setEnabled(!symbol->mid_symbol->isEmpty());
```
dg0yt, line 721: ... here ...

src/gui/symbols/line_symbol_settings.cpp
```diff
@@ -717,8 +720,9 @@ void LineSymbolSettings::updateStates()
...
 	}
-	mid_symbol_distance_label->setEnabled(mid_symbol_distance_label->isEnabled() && symbol->mid_symbols_per_spot > 1);
-	mid_symbol_distance_edit->setEnabled(mid_symbol_distance_edit->isEnabled() && symbol->mid_symbols_per_spot > 1);
+	const bool mid_symbol_distance_enabled = !symbol->mid_symbol->isEmpty() && symbol->mid_symbols_per_spot > 1;
```
dg0yt, line 723: ... but those two labels have a different condition here.

PR #2099, opened 2022-10-13
LineSymbol: Fix dashed line with mid symbols

### LINKED ISSUE #2064: Bezier curve around mid symbol of dashed line
### Steps to reproduce
1. Checking out the ISOM 517 Ruined Fence symbol I noticed a bump around the mid symbol see screencast below. What I do is moving the handle first, than switching to 516 Fence and back to 517.

https://user-images.githubusercontent.com/15611653/166445467-27227b00-4858-46aa-97d4-84873a92cea9.mp4

Just some strange behaviour, isn't it? - BTW it is the same for 514 Ruined wall.

### Configuration
Mapper Version: master v20220316.2
Operating System: Windows 7

dl3sdo:
Dashed lines with mid symbols that are placed in the center of a dash are sometimes disturbed.
This commit ensures that the dash is always drawn the same way, independent of whether mid symbols are present or not.
Fixes #2064


src/core/symbols/line_symbol.cpp
```diff
@@ -803,23 +803,19 @@ void LineSymbol::processContinuousLine(
...
-				orientation = split.tangentVector().angle();
-			mid_symbol->createRenderablesScaled(split.pos, orientation, output);
+				orientation = next_split.tangentVector().angle();
+			mid_symbol->createRenderablesScaled(next_split.pos, orientation, output);
 			
 			if (i > 1)
 			{
 				mid_position += mid_symbol_distance_f;
-				next_split = SplitPathCoord::at(mid_position, split);
-				path.copy(split, next_split, processed_flags, processed_coords);
```
dg0yt, line 821: IIUC this is the original spot which doesn't work as expected: This seems to start at split with the opposite direction.

dg0yt (approved):
I really like the change.

But I still wonder what is wrong with the original approach. Where is the hidden bug?

src/core/symbols/line_symbol.cpp
dl3sdo: @dg0yt: Thank you for reviewing the changes.

I also don't know why this code section doesn't provide the expected results.
After the merge request I spent some time doing more tests on the original code basis but was not yet able to find the root cause.

lpechacek: Many thanks, Matthias, for looking into the issue! I've spent some time playing with the code as well. This part of the code base is very difficult to read, partly due to methods using output parameters. In any case, my understanding is that `processContinuousLine()` only places the mid symbols but the carrier line segment is rendered elsewhere. I might be easily wrong though.

As a result of my playing, this seems to be the minimal fix to the rendering glitch.
```
diff --git a/src/core/symbols/line_symbol.cpp b/src/core/symbols/line_symbol.cpp
index 36b62be6d278..3265d408aa5e 100644
--- a/src/core/symbols/line_symbol.cpp
+++ b/src/core/symbols/line_symbol.cpp
@@ -803,7 +803,6 @@ void LineSymbol::processContinuousLine(
        {
                auto mid_position = (split.clen + effective_end_length - mid_symbols_length) / 2;
                auto next_split = SplitPathCoord::at(mid_position, split);
-               path.copy(split, next_split, processed_flags, processed_coords);
                split = next_split;
                
                auto orientation = qreal(0);
@@ -818,13 +817,12 @@ void LineSymbol::processContinuousLine(
                        {
                                mid_position += mid_symbol_distance_f;
                                next_split = SplitPathCoord::at(mid_position, split);
-                               path.copy(split, next_split, processed_flags, processed_coords);
                                split = next_split;
                        }
                }
        }
        
-       path.copy(split, end, processed_flags, processed_coords);
+       path.copy(start, end, processed_flags, processed_coords);
        
        processed_flags.back().setGapPoint(true);
        Q_ASSERT(!processed_flags[processed_flags.size()-2].isCurveStart());
```

Not messing with the line segment makes it render correctly. :-\

src/core/symbols/line_symbol.cpp
```diff
@@ -803,23 +803,19 @@ void LineSymbol::processContinuousLine(
 	{
 		auto mid_position = (split.clen + effective_end_length - mid_symbols_length) / 2;
 		auto next_split = SplitPathCoord::at(mid_position, split);
-		path.copy(split, next_split, processed_flags, processed_coords);
-		split = next_split;
 		
 		auto orientation = qreal(0);
-		bool mid_symbol_rotatable = bool(mid_symbol) && mid_symbol->isRotatable();
```
lpechacek, line 810: Removal of the test for `nullptr` makes me ask how come the `mid_symbol` pointer won't be NULL?

dl3sdo: I removed that check since it's already done some lines above: `set_mid_symbols = set_mid_symbols && mid_symbol && ...`

src/core/symbols/line_symbol.cpp
dl3sdo: Yes, processContinuousLine() places the line segment of a dash (but not its renderables) and adds the mid symbols to the list of renderables.
Maybe, the original idea was to divide the dash into smaller parts to have defined coord points to place the mid symbols. The proposed fix assumes that the coord points to place the mid symbols are still laying on the (curved) path, even if that (curved) dash is now processed in one piece.

src/core/symbols/line_symbol.cpp
lpechacek: Ah, yes. I overlooked that test. Thanks for the clarification and sorry for the noise.

dl3sdo:
@dg0yt @lpechacek: Do you prefer correcting commits (and then squash when merging) or should I force-push instead?

lpechacek:
> @dg0yt @lpechacek: Do you prefer correcting commits (and then squash when merging) or should I force-push instead?

As far as I'm concerned, squash on merge might be relevant in this particular case. In any case, Kai is the current code repo owner, so the final decision is with him.

dl3sdo:
The check was aborted due to https://github.com/actions/virtual-environments/issues/6002
Thus I intend to squash and force-push in order to have a successfully checked PR.

dl3sdo:
Kai, maybe you can fix a crash while working on that issue:
Use symbol 517 Ruined fence, change Mid symbols settings to
Center of gaps, 3 mid symbols per spot and increase mid symbol distance.
The symbol editor will crash Mapper while increasing to 0.30mm.
ASSERT: "p >= 0.0f" in file path_coord.cpp, line 471

dg0yt:
Hm...

dl3sdo:
Indeed, the Release version does not crash.

dg0yt:
And here, it fires when the first mid symbol is to be place before the gap.

Issue #2101, opened 2022-10-19
Incorrect rendering of area cross-hatch pattern


lpechacek:
### Steps to reproduce
1. Create a large area object with a cross-hatch pattern (e.g. 709 Out of bounds).
2. Zoom into the area so that parts of the area are out of view.

### Actual behaviour
Some of the lines in the top-left - bottom-right direction are missing while all the bottom-left - top-right lines are present. The issue affects screen rendering and exports as mentioned in #2100.

### Expected behaviour
Both directions should render correctly.

### Illustration
A map from 20th IARU World ARDF Championships Bulletin 4. Kudos to @mlerjen for the pointer.
![image](https://user-images.githubusercontent.com/5584406/196687258-2cc211da-e7c3-4767-a319-e8a70d9a66bd.png)

### Configuration
Mapper Version: v0.9.5
Operating System: all

dl3sdo:
When reproducing that issue I noticed some interesting things:
It's easy to reproduce when using the forest_sample map and adding an object of (710 Dangerous area). As soon as a part of the object is out of view on the left side (and only there), then parts of the lines in the top-left - bottom-right are missing.
The same is true for single hatched objects that are rotated (e.g., 411 Forest runnable in one direction).

However, when creating a new map and doing the same with (e.g., 410 Vegetation: fight, normal running in one direction) by making the Fill rotation adjustable per object, everything looks fine. But when I copy that symbol to the forest_sample map then the error occurs. This also happens with cross-hatched patterns.

lpechacek:
Thanks, Matthias, for your involvement here.

> However, when creating a new map and doing the same with (e.g., 410 Vegetation: fight, normal running in one direction) by making the Fill rotation adjustable per object, everything looks fine. But when I copy that symbol to the forest_sample map then the error occurs. This also happens with cross-hatched patterns.

Interestingly enough, I can replicate the issue easily after making the undergrowth symbol rotatable. Maybe I just chose a different object shape in my testing.

[Screencast from 2022-10-21 13-17-48.webm](https://user-images.githubusercontent.com/5584406/197184066-49c3f150-131e-4faf-b615-1745d26277b1.webm)

PR #2102, opened 2022-10-19
Fix rendering of cross-hatched areas

lpechacek:
See issue #2101 for details.

### DIFF
--- a/src/core/renderables/renderable_implementation.cpp
+++ b/src/core/renderables/renderable_implementation.cpp
@@ -287,7 +287,7 @@ LineRenderable::LineRenderable(const LineSymbol* symbol, QPointF first, QPointF 
 	right.normalize();
 	right *= half_line_width;
 	
-	extent.setTopLeft(first + right);
+	rectIncludeSafe(extent, first + right);
 	rectInclude(extent, first - right);
 	rectInclude(extent, second - right);
 	rectInclude(extent, second + right);
--- a/src/core/renderables/renderable_implementation.cpp
+++ b/src/core/renderables/renderable_implementation.cpp
@@ -283,14 +283,13 @@ LineRenderable::LineRenderable(const LineSymbol* symbol, QPointF first, QPointF 
 {
 	qreal half_line_width = (color_priority < 0) ? 0 : line_width/2;
 	
-	auto right = MapCoordF(second - first).perpRight();
-	right.normalize();
-	right *= half_line_width;
-	
-	rectIncludeSafe(extent, first + right);
-	rectInclude(extent, first - right);
-	rectInclude(extent, second - right);
-	rectInclude(extent, second + right);
+	auto margin = MapCoordF(second - first).perpRight();
+	margin.normalize();
+	margin.rx() = qAbs(margin.x()) * half_line_width;
+	margin.ry() = qAbs(margin.y()) * half_line_width;
+	extent = QRectF(first, second)
+	             .normalized()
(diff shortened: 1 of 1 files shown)


dl3sdo:
The fix looks fine to me. I just tested it and it provides the expected results.

dg0yt:
Thanks to both of you for picking up this issue.

`setTopLeft` was really wrong because it "may change the size". I think it should have been `moveTopLeft`. 
IMO `includeRect` does too much to the `extent` which is an initial, invalid `QRectF` at this point. 
(There is also `safeIncludeRect`, just for cases where the rect maybe initially invalid. So many helpers which even aren't pure functions...)

lpechacek:
Answering out of order...

> IMO `includeRect` does too much to the `extent` which is an initial, invalid `QRectF` at this point. (There is also `safeIncludeRect`, just for cases where the rect maybe initially invalid. So many helpers which even aren't pure functions...)

Thanks, Kai, for the review. `includeRect` is plain wrong in my patch as it causes too large extents for the lines. `includeRectSafe` does the right job in the freshly constructed `extent`.

> `setTopLeft` was really wrong because it "may change the size". I think it should have been `moveTopLeft`. 

The use of `moveTopLeft` would be a solution. Anyway, we would have wither to determine the minimum on the x and y coordinates, including the line width, or blindly set the rectangle origin and then call `includeRect`:

```
	extent.moveTopLeft(first);
	rectInclude(extent, first - right);
	rectInclude(extent, first + right);
	rectInclude(extent, second - right);
	rectInclude(extent, second + right);
```

My approach would be completely different, however. We've got a straightforward calculation here - two line endpoints, determine the bounding rectangle. My choice would be a complete rewrite here for the sake of comprehension and efficiency. The only disadvantage I can see is that we are working with individual coordinates rather than with vectors which is error prone. I don't see a vector-based solution at the moment though.

```
diff --git a/src/core/renderables/renderable_implementation.cpp b/src/core/renderables/renderable_implementation.cpp
index 32f6f78d9201..ab849f4fc977 100644
--- a/src/core/renderables/renderable_implementation.cpp
+++ b/src/core/renderables/renderable_implementation.cpp
@@ -281,17 +281,15 @@ LineRenderable::LineRenderable(const LineSymbol* symbol, QPointF first, QPointF
  , cap_style(Qt::FlatCap)
  , join_style(Qt::MiterJoin)
 {
-	qreal half_line_width = (color_priority < 0) ? 0 : line_width/2;
-	
 	auto right = MapCoordF(second - first).perpRight();
 	right.normalize();
-	right *= half_line_width;
-	
-	extent.setTopLeft(first + right);
-	rectInclude(extent, first - right);
-	rectInclude(extent, second - right);
-	rectInclude(extent, second + right);
-	
+	right *= (color_priority < 0) ? 0 : line_width;
+	auto const origin = QPointF { std::min(first.x(), second.x()) - right.x()/2,
+	                              std::min(first.y(), second.y()) - right.y()/2 };
+	auto const size = QSizeF { std::abs(first.x() - second.x()) + right.x(),
+	                           std::abs(first.y() - second.y()) + right.y() };
+
+	extent = QRectF { origin, size };
 	path.moveTo(first);
 	path.lineTo(second);
 }
```

lpechacek:
> My choice would be a complete rewrite here for the sake of comprehension and efficiency.

https://github.com/LibreMapper/mapper/commit/c3f86a75934a5f8fc2ae051f7fb66d2774b3b11e

PR #2110, opened 2022-11-14
MapEditorController: Show details for map part operations

dl3sdo:
When merging or removing a map part the user is not informed about the number of affected objects.
This easy change tells the user how many objects are subject of merging or removing a map part.

### DIFF
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2021 Kai Pastor
+ *    Copyright 2012-2022 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -3850,12 +3850,14 @@ void MapEditorController::addMapPart()
 void MapEditorController::removeMapPart()
 {
 	auto* part = map->getCurrentPart();
+	auto i = part->getNumObjects();
 	
 	QMessageBox::StandardButton button =
 	        QMessageBox::question(
 	            window,
                 tr("Remove current part"),
-                tr("Do you want to remove map part \"%1\" and all its objects?").arg(part->getName()),
+                i ? tr("Do you want to remove map part \"%1\" and its %2 objects?").arg(part->getName()).arg(i) :
+                    tr("Do you want to remove empty map part \"%1\"?").arg(part->getName()),
                 QMessageBox::Yes | QMessageBox::No );
 	
 	if (button == QMessageBox::Yes)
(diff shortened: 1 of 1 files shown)


dl3sdo:
Examples:
MergeMapPart
RemoveMapPart
RemoveEmptyMapPart

lpechacek (approved):

dl3sdo:
@pkturner: Indeed it's an option to remove the popup window if empty map parts shall be removed or merged.
I added another commit so that there is the option to take both or pick the first one only.
@lpechacek,@dg0yt: what is your preference?

lpechacek:
> @lpechacek,@dg0yt: what is your preference?

Given that the map part operations can be undone, I'm all fine with skipping the confirmation on empty parts.

Code-wise, I'm going to review the second patch soon. I've got a remark right away though. The variable holding the number of map parts (`i`) appears in more places after the change. It should get a more descriptive name.

dl3sdo:
> > @lpechacek,@dg0yt: what is your preference?
> 
> Given that the map part operations can be undone, I'm all fine with skipping the confirmation on empty parts.
> 
> Code-wise, I'm going to review the second patch soon. I've got a remark right away though. The variable holding the number of map parts (`i`) appears in more places after the change. It should get a more descriptive name.

I agree and will change `i` to `num_objects` once your review is finished.

src/gui/map/map_editor.cpp
```diff
@@ -3850,42 +3850,41 @@ void MapEditorController::addMapPart()
...
-			auto* combined_step = new CombinedUndoStep(map);
-			combined_step->push(add_step);
-			combined_step->push(undo_step);
-			undo_step = combined_step;
+			--num_objects;
+			auto* object = part->getObject(num_objects);
+			add_step->addObject(num_objects, object);
+			part->releaseObject(object);
 		}
+		while (num_objects > 0);
```
lpechacek, line 3878: The loop modifies the `num_objects` variable. While it's fine as is, it may cause trouble in future code maintenance, if the coder does not spot the value change and assumes the original `num_objects` value. I suggest that the value is left intact even after the part removal.

```suggestion
		for (auto i = 0; i < num_objects; ++i)
		{
			auto* object = part->getObject(i);
			add_step->addObject(i, object);
			part->releaseObject(object);
		}
```

Yes, the new code moves the objects into the undo in the reverse order, compared to the original code.

lpechacek (approved):
LGTM. The comment about the `num_objects` change is more a nitpicking than a code issue. Take it and re-rest the change if you like the change, or leave it as is. Thanks!

src/gui/map/map_editor.cpp
dl3sdo: Libor, thank you for your thourough review.

I basically agree with your proposal, but for a different reason, since the function itself is so small that I don't see the risk that someone would oversee the modification of the num_objects variable (which I otherwise would make a const one).
But I really like the usage of a for loop instead of a while loop.

However, I was really surprised that moving the objects into the undo in the reverse order is not an arbitrary choice. Instead, Mapper would crash when undoing the map part removal!
As an alternative I propose to use `for (auto i = num_objects - 1; i >= 0; --i)` to preserve the original order.
If you are fine with it, just give me the thumbs up.

PR #2113, opened 2022-12-05
Create ISOM 2017-2v5_15000.xmap

mlerjen:
Also here. New symbol set according to the color table and the most recent revision.

### DIFF
--- a/symbol sets/15000/ISOM 2017-2v5_15000.xmap
+++ b/symbol sets/15000/ISOM 2017-2v5_15000.xmap
@@ -1,0 +1,10122 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<map xmlns="http://openorienteering.org/apps/mapper/xml/v2" version="9">
+    <notes></notes>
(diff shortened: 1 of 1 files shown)


MaBreaker:
It seems that this merge does not just implement "minor" corrections to the existing symbol set, but introduces totally different approach to define symbols for multiple types of terrain objects like cultivated land, orchard etc. also many symbol sub numbers are just switched to another meaning compared to original symbol set provided within OOM.

This means that if anyone tries to do "update the latest symbol set" to the existing map created by using original symbol set (previous way of working) it would "broke" the whole map even if the best possible "crt/replacement" mapping is used.

Wouldn't it be better to try to keep "existing" sub numbering and also layout of the fields etc. and just "fix" symbol dimensions and colours there. That would make map updates and also github "change" tracking much easier.

dg0yt:
For review: [Report on colors and symbols](https://github.com/user-attachments/files/15896087/index.html.zip)

Bildschirmfoto vom 2024-06-19 08-05-50

PR #2127, opened 2023-01-17
Show usage of color before deletion

dl3sdo:
When the user decided to delete a color that was used by at least one symbol, a warning was issued: "The map contains symbols with this color. Deleting it will remove the color from these objects!".
This warning was however ambiguous: it was even issued when the affected symbol(s) were not used by any object. In addition, the mixture of "symbols" and "these objects" sounded awkward.

With the first commit the number of affected symbols is counted and put out in the warning message.
In addition the number of objects actually using these affected symbols is counted as well. If there is at least one object then the warning is extended by stating the number of affected objects. In any case the message text is changed to "...remove the color from these symbols".

The second commit allows to select the symbols that will be affected by removing a given color by extending the warning message box by a new button 'Select affected symbols and abort'.
Pressing on this button then selects all of the affected symbols in the symbol window.

Warning message when symbols but no objects are affected:
ColorDeletionSymbols

Warning message when symbols and objects are affected:
ColorDeletionSymbolsObjects
After pressing on the 'Select affected symbols and abort' button:
SelectedSymbols

### DIFF
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2021 Kai Pastor
+ *    Copyright 2012-2023 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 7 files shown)


dl3sdo:
Superseded by #2222

Issue #1639, opened 2020-05-23
Missing pylons on imported powerline (normal nodes can't be converted to dash nodes)


reinisberzins:
### Steps to reproduce
A) Import a powerline from .osm file.
B) Or draw a powerline by hand with non-dash mode (e.g. pressed spacebar accidentally or changed my mind later).

### Actual behaviour
Powerline has no pylons (dash symbols) because all nodes are normal (non-dash) nodes and can't be converted to dash nodes (with pylons showing).

### Expected behaviour
A) Dash symbol should always appear on every node - normal or dash node (unless "Suppress dash symbol at start and end" is checked) because powerlines can't change direction without a pylon and I don't know any no other standard orienteering symbols that use dash symbols (all others use mid symbols only).*
B) Assign dash node to every imported node of powerlines.
C) Future Idea: an option to easily convert all selected nodes to dash nodes or reverse.

### Configuration
Mapper Version: 0.9.3
Operating System: Windows 10 x64

*Actually the current behaviour would be useful when drawing a depression/slope with a new customised symbol of 101 Contour line + dash symbol of slope marker (which doesn't exist in standard symbol sets) where you mostly draw non-dash nodes, but occasionally toggle dash mode and insert a slope marker. But I'm not sure if anyone uses this method.

dg0yt:
> Actually the current behaviour would be useful when drawing a depression/slope with a new customised symbol of 101 Contour line + dash symbol of slope marker (which doesn't exist in standard symbol sets) where you mostly draw non-dash nodes, but occasionally toggle dash mode and insert a slope marker. But I'm not sure if anyone uses this method.

https://github.com/OpenOrienteering/mapper/issues/1035#issuecomment-376235108

dg0yt:
Unfortunately, the pylon property in OSM is on the individual nodes, not on the way that they form, and we don't have rules to handle to that.

For now, you can switch the state of indiviudal nodes with the Edit-points tool, by Space key + Click.

PR #2136, opened 2023-03-18
Auxiliary symbol properties

dl3sdo:
Mapper sometimes needs to handle symbol properties that are not part of Mapper's own set of symbol properties and that need to be applied when importing objects.
Example: horizontal and vertical text alignment is a symbol property in OACD but an object property in Mapper.
Mapper uses auxiliary data structures to temporary store those symbol properties for later applying them on object import.
This change adds a QVariantHash element to the Symbol base class that allows a flexible and type safe storage of auxiliary symbol properties.
These auxiliary symbol properties are neither saved to a map file nor loaded from it.

The second commit applies the auxiliary symbol properties to the DXF import where text object properties like the text itself, alignment and rotation were previously imported by putting them into a string and transporting the string via the description property of the related text symbol.
Besides being an awkward design the approach suffers from multiple deficiencies: the auxiliary string was not removed from the description field, the rotation angle was rounded (e.g., -17.2 became 20) and the way of constructing and parsing the string was unnecessary complex.
Note that this commit replaces #2135.
    
The third commit applies the auxiliary symbol properties to the import of OCAD maps.
Auxiliary symbol properties are replacing the text_halign_map and text_valign_map hashes that were used to store OCAD's symbol properties for horizontal and vertical text alignment for being applied later to text object import.

This PR is a POC to demonstrate a possible way of preserving symbol/object related properties until object creation.

Auxiliary symbol properties could be considered with respect to #2082 and #1639.

### DIFF
--- a/src/core/symbols/symbol.h
+++ b/src/core/symbols/symbol.h
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2023 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -32,6 +32,7 @@
 #include <QtGlobal>
 #include <QFlags>
 #include <QHash>
+#include <QVariant>
 #include <QImage>
 #include <QMetaType>
 #include <QRgb>
@@ -499,6 +500,26 @@ public:
 	 */
 	virtual bool hasRotatableFillPattern() const;
 	
+	/**
+	 * Sets an auxiliary property of this symbol.
+	 */
+	void setAuxiliaryProperty(QString key, QVariant value) { auxiliary_properties.insert(key, value); }
+	
+	/**
+	 * Returns an auxiliary property of this symbol.
+	 */
+	QVariant getAuxiliaryProperty(QString key) const { return auxiliary_properties.value(key); }
+	
+	/**
+	 * Returns an auxiliary property of this symbol or a default value.
+	 */
+	QVariant getAuxiliaryProperty(QString key, QVariant default_value) const { return auxiliary_properties.value(key, default_value); }
+	
+	/**
+	 * Returns a reference to the auxiliary property member.
+	 */
+	const QVariantHash& getAuxiliaryPropertyHash() const { return auxiliary_properties; }
+	
 protected:
 	/**
 	 * Sets the rotatability state of the symbol.
@@ -621,6 +642,7 @@ private:
 	bool is_hidden;           /// \see isHidden()
 	bool is_protected;        /// \see isProtected()
 	bool is_rotatable = false;
+	QVariantHash auxiliary_properties;	/// Auxiliary properties for import of objects (e.g., .ocd, .dxf files). Auxiliary properties are not saved in a map file.
 };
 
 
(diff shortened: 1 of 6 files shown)


src/gdal/ogr_file_format.cpp
```diff
@@ -1176,15 +1176,9 @@ Object* OgrFileImport::importPointGeometry(OGRFeatureH feature, OGRGeometryH geo
...
+	if (symbol->getType() == Symbol::Text && !symbol->getAuxiliaryProperty(QLatin1String("label")).isNull())
 	{
-		const auto& description = symbol->getDescription();
-		auto length = description.length();
-		auto split = description.indexOf(QLatin1Char(' '));
-		FILEFORMAT_ASSERT(split > 0);
-		FILEFORMAT_ASSERT(split < length);
-		
-		auto label = description.right(length - split - 1);
+		auto label = symbol->consumeAuxiliaryProperty(QLatin1String("label")).toString();
```
dg0yt, line 1181: I didn't test now, but I guess that this shows the potential *and* the missing pieces:
Putting the label into the description was a hack, but it helped to assign objects with the same label to a a single symbol. Probably also on a second and third import into the same map. 

This could be covered with an automated test: a test vector file with to identical objects, imported twice. At the end there must be a map with four objects with one symbol.

dg0yt (commented):
Early notes.

src/core/symbols/symbol.h
```diff
@@ -650,6 +676,7 @@ class Symbol
 	bool is_hidden;           /// \see isHidden()
 	bool is_protected;        /// \see isProtected()
 	bool is_rotatable = false;
+	QVariantHash auxiliary_properties;	/// Auxiliary properties for import of objects (e.g., .ocd, .dxf files). Auxiliary properties are not saved in a map file.
```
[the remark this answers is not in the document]
dl3sdo: Done

src/fileformats/ocd_file_import.cpp
```diff
@@ -1922,7 +1922,7 @@ Object* OcdFileImport::importObject(const O& ocd_object, MapPart* part)
 		auto t = new TextObject(symbol);
 		t->setText(getObjectText(ocd_object));
 		t->setRotation(convertAngle(ocd_object.angle));
-		t->setHorizontalAlignment(text_halign_map.value(symbol));
+		t->setHorizontalAlignment((symbol->getAuxiliaryProperty(QLatin1String("HAlign")).value<TextObject::HorizontalAlignment>()));
```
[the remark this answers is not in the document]
dl3sdo: Done, I put the enums in the source files. I'm not sure whether the anonymous namespaces around are superfluous.

dl3sdo:
Test file:
[DXFTextImport.zip](https://github.com/user-attachments/files/18632356/DXFTextImport.zip)

src/gdal/ogr_file_format.cpp
dl3sdo: Due to my understanding of OgrFileImport::getSymbolForLabel() objects with the same color and font size are assigned to a single symbol, the label as well as anchor and angle are always overwritten.
My proposal would not change the current behaviour, IMO.

dl3sdo:
Adding an automated test as proposed by @dg0yt shows an interesting behaviour (maybe a bug in my version of OGR):
The import of the angle as done by `auto angle = OGR_ST_GetParamDbl(tool, OGRSTLabelAngle, &is_null);`
will return

- -1.12 for -1.1234
- 1.12 for 1.1234
- 17.2 for -17.234
- 53.2 for 53.234
- -348.0 for -348.456

src/gdal/ogr_file_format.cpp
dl3sdo: Done

dl3sdo:
I propose to merge #2421 first.

PR #2142, opened 2023-03-23
XmlElementReader: Fix file format error messages

dl3sdo:
When opening an .omap file, the error messages being shown if the actual number of object coordinates doesn't match the number inside the file didn't contain both numbers since these were not passed by arg().


dl3sdo:
@lpechacek, @dg0yt, please review, tnx.

dg0yt (approved):
Thanks Matthias

PR #2144, opened 2023-03-26
OcdFileImport: Handle emtpy area holes

dl3sdo:
There are .ocd files that erroneously contain an empty area hole, i.e., there are two subsequent points with the Ocd::OcdPoint32::FlagHole property.
If such a malformed .ocd file is imported Mapper will crash if the area object containing the empty hole is selected. 
To fix this, the first of these .ocd points needs to be ignored. Since setPointFlags() applies the HolePoint property to the last point instead of the first point of the next part, the consequence is that instead of removing the first point of these two .ocd points, the second point of the two Mapper points with the HolePoint property needs to be removed.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2013-2022 Kai Pastor
+ *    Copyright 2013-2023 Kai Pastor
  *
  *    Some parts taken from file_format_oc*d8{.h,_p.h,cpp} which are
  *    Copyright 2012 Pete Curtis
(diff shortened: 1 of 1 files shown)


dl3sdo (commented):
Removing empty area holes should not be restricted to just one but be extended to multiple subsequent ones by replacing the `if` statement by a `while` loop.

src/fileformats/ocd_file_import.cpp
```diff
@@ -2153,6 +2153,17 @@ void OcdFileImport::fillPathCoords(OcdImportedPathObject *object, bool is_area,
 		size_t start = 0;
 		for (size_t i = 0; i < object->coords.size(); ++i)
 		{
+			// There are .ocd files that erroneously contain an empty area hole, i.e., there are two subsequent points 
+			// with the Ocd::OcdPoint32::FlagHole property. To fix this, the first of these points needs to be ignored.
+			// Since setPointFlags() applies the HolePoint property to the last point instead of the first point of the next part,
+			// the consequence is that instead of removing the first point of these two .ocd points,
+			// the second point of the two Mapper points with the HolePoint property needs to be removed.
+			// Using a loop ensures that multiple subsequent empty area holes are removed.
+			while (is_area && i < object->coords.size() - 1 && object->coords[i].isHolePoint() && object->coords[i+1].isHolePoint())
```
dg0yt, line 2162: I think it might be better to use a separate loop (or STL algorithm) then to nest loops.

dg0yt (commented):
Maybe I need some real or synthetical test data here.

dl3sdo:
Kai, I'll send you files via email.

dl3sdo:
Superseded by #2224 and #2225

PR #2148, opened 2023-04-07
OCD format enhancements

lpechacek:
This patch set is related to various OCD format support enhancements. The first two commits complement existing changes in the repo. The next two add new functionality for the sake of compatibility. I'd be happy if they were merged into the mainline Mapper code.


dl3sdo:
Libor, thank you for the PR.
I am happy to see that you started to address the topic of object timestamps which I really want to have in Mapper for a number of reasons.
Thus I would prefer to go even one step further and not add the timestamps to the objects' tags but instead as object properties and also save them in omap file format directly.
This would not imply that the full support of timestamp updates needs to be provided immediately.

dl3sdo (approved):
Yes, the value should be set to -1 as my analysis of .ocd12 files shows as well.

PR #2154, opened 2023-04-27
Consider dash symbols in combined symbols

dl3sdo:
When drawing line objects for symbols with dash symbols, no dash points were set automatically if the symbol was a combined symbol.
When editing line objects for symbols with dash symbols, additional points were not added as dash points if the symbol was a combined symbol.
These deviations were known and marked as TODO.
This commit treats line symbols with dash symbols and combined symbols containing line symbols with dash symbols equally.

### DIFF
--- a/src/core/symbols/combined_symbol.cpp
+++ b/src/core/symbols/combined_symbol.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2020, 2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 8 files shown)


src/tools/tool_helpers.cpp
```diff
@@ -746,4 +746,42 @@ void AzimuthInfoHelper::draw(QPainter* painter, const MapWidget* widget, const M
 }
 
 
+/**
+ * Helper function to determine whether a symbol that is either a line symbol
+ * or contains line symbols as part of a combined symbol contains a dash symbol.
+ * Function is used by DrawPathTool::updateDashPointDrawing() and 
+ * EditPointTool::addDashPointDefault() in relation to setting and changing dash points.
+ */
+bool symbolContainsDashSymbol(const Symbol* symbol)
```
dg0yt, line 755: I think this should become a virtual member function `Symbol::containsDashSymbol`, with suitable overrides in subclasses.

dl3sdo:
Rebased to solve merge conflicts.
@dg0yt: please review.

dl3sdo:
@dg0yt: I agree to your proposed changes, thank you.

PR #2156, opened 2023-05-17
Make QPainterTest pass with Qt 5.15.9

mitya57:
https://bugreports.qt.io/browse/QTBUG-100327 was fixed in 5.15.9, so now we have a good result from the beginning and don't need ImageTransparencyFixup.

This fixes a test failure which can be seen in [this log](https://launchpadlibrarian.net/666512825/buildlog_ubuntu-mantic-amd64.openorienteering-mapper_0.9.5-3build8_BUILDING.txt.gz).

### DIFF
--- a/src/core/image_transparency_fixup.h
+++ b/src/core/image_transparency_fixup.h
@@ -26,8 +26,12 @@ namespace OpenOrienteering {
 
 
 /**
+ * Repairs a pixel composing issue aka QTBUG-100327.
+ * 
  * ImageTransparencyFixup repairs a particular issue with composing
- * transparent pixels.
+ * transparent pixels with Qt5 < 5.15.9 and Qt6 < 6.2.4, tracked
+ * upstream as https://bugreports.qt.io/browse/QTBUG-100327, and
+ * does nothing otherwise.
  * 
  * QPainter::CompositionMode_Multiply and QPainter::CompositionMode_Darken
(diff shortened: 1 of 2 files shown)


dg0yt:
Thank you very much for this PR.

The test served its purpose when it found the Qt fixed :tada: However, the solution needs to be modified:
- IMO, only `QEXPECT_FAIL` shall be guarded in the test.
  (Qt is no longer expected to fail, but everything else is expected to work unchanged.)
- The `ImageTransparencyFixup` operation should be turned into a no-op for the fixed versions of Qt.
  (It would remain "called" in source code, but optimized away by the compiler. Fixup details remain encapsulated in `ImageTransparencyFixup`.)
- It would be nice if the versions conditions should already be adjusted to correctly handle the versions of Qt 6.

If you don't feel comfortable doing these changes, we can take over.

mitya57:
Okay, updated. As you said that only `QEXPECT_FAIL` should be guarded, I removed the assertion that checked for wrong behavior. If you want to keep it, I can move it inside the guarded block before the `QEXPECT_FAIL` line.

mitya57:
Does this PR look good to you now?

dg0yt:
Looks good on https://build.opensuse.org/package/show/home:dg0yt/openorienteering-mapper.

dg0yt:
Thanks!

PR #2157, opened 2023-05-22
JSON based Mapper configuration

### LINKED ISSUE #1961: Dash point by on default
(I had this problem before, #1728)

### Steps to reproduce
1. Draw a line - dash point "purple dot" is on.
2. Draw a second line - switch off dash point by pressing space bar
3. Draw a third line - dash point still off. Normal points are set.
4. Draw another line object.
5.  Draw a fourth line with the first object - dash point is on again.
image

### Actual behaviour
So dash point is the default. It can be switched off as long as you stick with drawing features of the same symbol, but jumps back to default = dash point as soon you draw something else.

### Expected behaviour
The dash point as the exception should be opt-in, but not default.

### Configuration
Mapper Version: 0.9.5
Operating System: windows 7

### LINKED ISSUE #2087: Magnetic declination lookup button does not work (NOAA added a key)

dl3sdo:
In addition to the configuration of Mapper through the 'Settings' GUI where a dedicated control element is needed for each setting, this commit implements a lightweight mechanism by using a JSON configuration file that allows to easily configure a variety of settings.
The 'Settings' GUI is extended by the possibility to select and remember the JSON configuration to be used as default configuration.
An additional dialog allows to load and save JSON configuration and provides a simple, i.e., not context-sensitive, text editor.
The validity of the JSON format is checked and deviations to Mapper's default JSON configuration can be shown and resolved.

As examples for its usage it closes #2087 and closes #1961, thus superseding #2088 and #2154.

JSON_Configuration_Dialog

Note: this is currently just a POC and there are some TODOs:
- when the 'Settings' dialog is reset or canceled, also the JSON settings should be reset
- the code needs proper comments and maybe some refactoring
- a test should be added

However, before continuing I need feedback whether this kind of JSON based configuration is desired at all.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,6 @@
 #
 #    Copyright 2012-2014 Thomas Schöps
-#    Copyright 2012-2018 Kai Pastor
+#    Copyright 2012-2023 Kai Pastor
 #    
 #    This file is part of OpenOrienteering.
 # 
@@ -204,6 +204,7 @@ set(Mapper_Common_SRCS
   gui/widgets/template_list_widget.cpp
   gui/widgets/text_browser.cpp
   gui/widgets/toast.cpp
+  gui/widgets/json_config_widget.cpp
   
   templates/paint_on_template_feature.cpp
   templates/paint_on_template_tool.cpp
@@ -266,6 +267,7 @@ set(Mapper_Common_SRCS
   util/translation_util.cpp
   util/util.cpp
   util/xml_stream_util.cpp
+  util/json_config.cpp
 )
 
 # Extra header to be shown in the IDE or to be translated
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2023 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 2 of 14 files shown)


jmacura:
I think this is just great. If more config options can find its way down here, that would be awesome. Also custom colors for scribbling might perhaps be stored in here (#1738)?
That way you can switch to another OO Mapper instance / device and keep all your preferences in a single file 🥳

PR #2158, opened 2023-05-23
Persist OCD color numbers

lpechacek:
This is a working implementation addressing issue 1245. The introduction of a stable color id uncovered some room for improvement in the native Mapper format and the new tests discovered bugs and limitations in the OCD format implementation. The first four patches introduce the stable color id required for smooth interoperability with the other software. The next four ones address defects and limitations in the OCD support routines.

There is certainly number of points for discussion in the patch set which I'm leaving for a possible review round.


lpechacek:
Marking the pull request as draft. There are problems in symbol import - the newly created colors can have duplicate id's which in turn breaks OCD export. I'll update the patch set soon.

lpechacek:
FWIW, this implementation has served me well for a few months. I'm not saying it's perfect. There are many spots in the code where improvement might be desirable. On the other hand, this exercise has uncovered a few defects. The fixes may be cherry-picked on its own. That said, I'm removing the draft status.

dl3sdo:
Libor, thank you for solving that issue.
At a first glance I noticed that you don't use a color_id value of 0. On the other hand, your checks don't exclude 0 (e.g., `if (color_id >= 0` or `return (id >= 0`). I consider that not an issue indeed.

In addition I noticed that the Registration black color changed its id from 1 to 45 during the course of import from and export to .ocd (OCAD12), although 1 is not used in the exported map.

lpechacek:
> At a first glance I noticed that you don't use a color_id value of 0. On the other hand, your checks don't exclude 0 (e.g., `if (color_id >= 0` or `return (id >= 0`). I consider that not an issue indeed.

Ack. I noticed in some OCD files that zero is used as an ID. The Viewer also allows that value. But start at one is more common and it's also aligned with the Pascal view of the world. That's why zero is allowed but not preferred.

> In addition I noticed that the Registration black color changed its id from 1 to 45 during the course of import from and export to .ocd (OCAD12), although 1 is not used in the exported map.

Yes, likely a bug in the code. :) I'll have a look.

lpechacek:
> Yes, likely a bug in the code. :) I'll have a look.

Fixed and rebased the patch set. The thing is that spot colors were assigned new id's in the OCD import path. The export path saw the low id numbers as used and searched for a new id for the registration black. Thanks again for testing the change reporting your findings!

Issue #2167, opened 2023-07-06
Vikazimuth compatibly


dg0yt:
Hi,

We have continued to improve the link between OOM and Vikazimut for creating routes from OOM.
We've managed the differences between the IOF and OOM formats in the XML file, but the map scale is missing from your XML.
Is it possible to add this tag to the XML?
RaceCourseData>
  Map>
      Scale>10000 /Scale>
  /Map>
...
/RaceCourseData>

- Second request, would it be possible to create multiple races in one XML?

- If you can make it easier to create the definition table, it would remove a big drawback to creating courses under OOM.

_Originally posted by @Eric-P-Kmz in https://github.com/OpenOrienteering/mapper/issues/1883#issuecomment-1619142674_

app4soft:
Duplicate of https://github.com/OpenOrienteering/mapper/issues/2166

Issue #2166, opened 2023-07-05
XML format for Vikazimut


Eric-P-Kmz:
### Steps to reproduce
We have continued to improve the link between OOM and Vikazimut for creating routes from OOM.
1 - We've managed the differences between the IOF and OOM formats in the XML file, but the map scale is missing from your XML.

 2 - Would it be possible to create multiple races in one XML?

 3 -  If you can make it easier to create the definition table, it would remove a big drawback to creating courses under OOM.

### Actual behaviour
2 - Only one race in XML

3 - No GUI to build the definition table

### Expected behaviour
1 - Is it possible to add this tag to the XML?
RaceCourseData>
Map>
Scale>10000 /Scale>
/Map>
...
/RaceCourseData>

2 - Create multiple <Course> tags in the XML file
3 - Create a GUI for the definition table

### Configuration
Mapper Version: 0.9.5
Operating System: Windows

dl3sdo:
@Eric-P-Kmz: I added a commit that adds the map scale to the XML. It was also necessary to add the top left and bottom right positions to get an IOF compliant XML.
You can use one of the created artifacts from the Azure Pipeline to test the XML output with Vikazimut.

Eric-P-Kmz:
Hello,

We have just tested this new version. No more errors. I'm preparing a video to show how to create a course with OOM and export it to Vikazimut.
Thanks

dl3sdo:
Regarding topic 2 (multiple races in one XML):
You propose that it should be possible to create multiple, e.g., 'Simple Orienteering Course' objects and export them as multiple <RaceCourseData> elements of the same <Event> ?

Eric-P-Kmz:
Yes, that's right. You'd have to be able to name 'Simple Orienteering Course' objects to do that .
 OCAD XML exemple
Course>
      Name>Chasse</Name>
      Length>2630</Length>
      Climb>0</Climb>
      CourseControl type="Start">
        <Control>S1</Control>
      </CourseControl>
      <CourseControl type="Control">
        <Control>31</Control>
        <LegLength>34</LegLength>
      </CourseControl>
      CourseControl type="Finish">
        Control>F1</Control>
        LegLength>63</LegLength>
      /CourseControl>
    /Course>

    Course>
      Name>Vert - facile</Name>
      Length>1300</Length>
      Climb>0</Climb>
      CourseControl type="Start">
        Control>S1</Control>
      /CourseControl>
      CourseControl type="Control">
[...]

PR #2168, opened 2023-07-06
IofCourseExport: Add 'Map' element

dl3sdo:
The 'Map' element is added to the IOF 3.0 compliant xml export of course data.

This commit partly solves #2167, #2166.

### DIFF
--- a/src/fileformats/iof_course_export.cpp
+++ b/src/fileformats/iof_course_export.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2021 Kai Pastor
+ *    Copyright 2021, 2025 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -22,6 +22,8 @@
 #include <Qt>
 #include <QDateTime>
 #include <QLatin1String>
+#include <QPointF>
+#include <QRectF>
 #include <QString>
 #include <QXmlStreamWriter>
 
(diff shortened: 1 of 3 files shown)


src/fileformats/iof_course_export.cpp
```diff
@@ -97,11 +98,36 @@ void IofCourseExport::writeXml(const PathObject& object)
...
 
+void IofCourseExport::writeMap()
+{
+	XmlElementWriter map_data(*xml, QLatin1String("Map"));
+	xml->writeTextElement(QLatin1String("Scale"), QString::number(map->getScaleDenominator()));
+	
+	// IOF.xsd proposes for 'MapPositionTopLeft': "The position of the map's top left corner given in the map's coordinate system, usually (0, 0)."
+	// With this the bottom right position is determined by the (negative) size of the map extent.
+	// Alternatively use real coordinates from the map_extent structure below.
+	// NOTE: templates are not considered when determining the map extent.
```
dg0yt, line 115: Hm, the current file can be just the course, with the actual map being a template.

dl3sdo: That's true. With the current demand for the scale information, the map positions came in as well but I doubt that anyone or any app would really use them right now.
However, I am thinking about whether to change the map positions to use the real map positions instead shifting the top left to (0,0).
Unfortunately I found no example for 'Map' settings on [Github](https://github.com/international-orienteering-federation/datastandard-v3/tree/master)

dl3sdo: @dg0yt: I now consider templates (if visible) and changed the map positions to use the real map positions instead.

src/fileformats/iof_course_export.cpp
```diff
@@ -97,11 +99,35 @@ void IofCourseExport::writeXml(const PathObject& object)
...
+		map_position_topleft.writeAttribute(QLatin1String("x"), QString::number(map_extent.topLeft().x()));
+		map_position_topleft.writeAttribute(QLatin1String("y"), QString::number(map_extent.topLeft().y()));
+		map_position_topleft.writeAttribute(QLatin1String("unit"), QLatin1String("mm"));	// optional attribute, default is 'mm'
+	}
+	{
+		XmlElementWriter map_position_bottomright(*xml, QLatin1String("MapPositionBottomRight"));
+		map_position_bottomright.writeAttribute(QLatin1String("x"), QString::number(map_extent.bottomRight().x()));
+		map_position_bottomright.writeAttribute(QLatin1String("y"), QString::number(map_extent.bottomRight().y()));
+		map_position_bottomright.writeAttribute(QLatin1String("unit"), QLatin1String("mm"));	// optional attribute, default is 'mm'
+	}
```
dg0yt, line 128: Is the extent information useful at all when not also providing the `Image` element?

dl3sdo: > Is the extent information useful at all when not also providing the `Image` element?

Good question, but I assume yes, since the 'Image' element is not mandatory and the PR works for the issue reporter due to #2166

PR #2179, opened 2023-09-06
DrawPathTool: Highlight object that is going to be followed

lpechacek:
In complex map situations it is unclear which object will be followed when the user uses the Shift modifier key to follow lines or area object boundaries. This is a proposal how to make the object selection more obvious.

Note the dashed object boundary highlight in the video.

[Screencast from 2023-09-06 15-49-27.webm](https://github.com/OpenOrienteering/mapper/assets/5584406/64c26fdf-0430-4f8f-bbf5-dfa23eb4ee8c)

### DIFF
--- a/src/tools/draw_path_tool.cpp
+++ b/src/tools/draw_path_tool.cpp
@@ -51,6 +51,7 @@
 #include "core/path_coord.h"
 #include "core/virtual_coord_vector.h"
 #include "core/virtual_path.h"
+#include "core/symbols/combined_symbol.h"
 #include "core/symbols/line_symbol.h"
 #include "core/symbols/symbol.h"
 #include "core/objects/object.h"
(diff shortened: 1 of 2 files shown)


mlerjen:
The clarification which object you follow is welcomed. To make it really useful it should be combined with the possibilty to "loop" through the possible objects on the actual click location. Thus a repeated crtl & click on the (+/-) same spot with no "drag" added,  should result in this "loop".

(e.g. while "following" a line shared by multiple objects, I sometimes somehow am able to switch the object to follow and sometimes  I don't. The logic behind seems unclear. With the highlighting of the object to follow as @lpechacek proposes, this could make the logic clear, but anyway. I'd like to have the choice which object to follow at any given point )

lpechacek:
Thanks for the comment, @mlerjen . I'm happy that my contributions still spark interest in the user community. :)

> To make it really useful it should be combined with the possibilty to "loop" through the possible objects on the actual click location.

Agreed. Only after adding the "followed object highlight" I realized that it's always the closest object that is followed. Before the new feature addition I thought that Mapper cycles between the available objects but I just don't see it. So, implementation of the object cycling (in fact a borrow from the edit point tool) is a logical next step. I'm, however not sure if I'm to implement it as this deficiency doesn't [itch](http://www.catb.org/esr/writings/homesteading/cathedral-bazaar/ar01s02.html) me that strong.

> Thus a repeated crtl & click on the (+/-) same spot with no "drag" added, should result in this "loop".

And the repeated click should not add extra vertices on the path. The extra points added by the draw path tool bother me quite a lot, starting with the extra point added on the first click with follow path. IMO, the draw path tool should automatically merge new points added by clicking within click distance on the new path, for example when starting a new path or switching between followed objects.

eolmapper:
I tested the PR, it’s a very good idea!
On the other hand, I find that the highlight with the red dash line is not always clearly visible and brings a confusion with the red editing object line.
It may be necessary to change the color or type of symbol. 
Why not dots instead of the dash line?

lpechacek:
@eolmapper , thanks for the feedback! I looked into the dots version of the object highlight. The draft version is in #2204. Let me know if that highlight method feels better.

PR #2181, opened 2023-09-10
Color edit dialog - show symbol usage

lpechacek:
Currently it's somewhat difficult to identify symbols that use a certain color. This change set adds a list showing the color usage in symbols ~~and the possibility to trigger symbol editing dialog from the color dialog~~.

Use case - color table clean up after map/symbol set import.

[Screencast from 2023-09-10 19-52-49.webm](https://github.com/OpenOrienteering/mapper/assets/5584406/ddcb82c0-df19-4da7-8e94-f13a993e3b51)

### DIFF
--- a/src/gui/color_dialog.cpp
+++ b/src/gui/color_dialog.cpp
@@ -51,10 +51,13 @@
 #include <QTabWidget>
 #include <QVariant>
 #include <QWidget>
+#include <QListWidget>
 
 #include "core/map.h"
 #include "core/map_color.h"
+#include "core/symbols/symbol.h"
 #include "gui/util_gui.h"
+#include "gui/symbols/symbol_setting_dialog.h"
 #include "gui/widgets/color_dropdown.h"
 #include "util/backports.h"  // IWYU pragma: keep
 #include "util/util.h"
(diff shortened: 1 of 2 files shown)


dl3sdo:
I tested the PR and it works excellently.

src/gui/color_dialog.cpp
```diff
@@ -839,5 +871,23 @@ void ColorDialog::rgbValueChanged()
 	}
 }
 
+void ColorDialog::colorUsageItemClicked(QListWidgetItem* clicked_item)
+{
+	auto item_data = clicked_item->data(Qt::UserRole);
+	Q_ASSERT(item_data.canConvert<const Symbol*>());
+	auto* symbol = const_cast<Symbol*>(item_data.value<const Symbol*>());
+
+	auto* map = const_cast<Map*>(&this->map);
```
dg0yt, line 880: These two instances of `const_cast` are a hint: Originally, `ColorDialog` is designed to not change the map. The click-through feature now enables direct changes (to the symbols). IMO we either have to change the relationship to the map, or we don't add this click-through feature at the moment. I tend to prefer the latter.

(And changing the symbols implies the need to update the color usage list.)

lpechacek: Thanks for the review, Kai, I acknowledge the approach and I've updated the pull request.

dg0yt:
Thanks!

PR #2184, opened 2023-10-03
ObjectQuery: Find objects based on symbol number

### LINKED ISSUE #2183: Find object by code number
### Steps to reproduce
1.Draw some object with symbol 123
2. Find Object with the expression SYMBOL 123 (Like it's describe in the user manual)

### Actual behaviour
Nothing happens. the objects are not find

Its work fine with the name of the symbol but it will be easily if it worked with the symbol code.
I have tried different syntax variations but without success

dl3sdo:
Extend the 'Find' functionality to search for objects with a specific symbol number as described by the Mapper manual (e.g.,'SYMBOL 123').

Closes #2183

### DIFF
--- a/src/core/objects/object_query.cpp
+++ b/src/core/objects/object_query.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2016 Mitchell Krome
- *    Copyright 2017-2022 Kai Pastor
+ *    Copyright 2017-2023 Kai Pastor
  *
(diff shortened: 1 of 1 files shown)


dg0yt:
The fix is much simpler: #2200.
But tnx!

dl3sdo:
@dg0yt: your fix is not just simpler but it uses the already present mechanisms (I actually had overseen that SYMBOL was a known keyword).

PR #2204, opened 2024-02-03
DrawPathTool: Highlight object that is going to be followed (dots version)

lpechacek:
This is a variant of PR#2179 that uses dots for the highlight. The code complexity is higher than with the dashes and the patch would need a bit of tweaking before a merge. The pull request is in place to make the implementation effort public and give users the opportunity to test the change.

### DIFF
--- a/src/tools/draw_path_tool.cpp
+++ b/src/tools/draw_path_tool.cpp
@@ -51,6 +51,7 @@
 #include "core/path_coord.h"
 #include "core/virtual_coord_vector.h"
 #include "core/virtual_path.h"
+#include "core/symbols/combined_symbol.h"
 #include "core/symbols/line_symbol.h"
 #include "core/symbols/symbol.h"
 #include "core/objects/object.h"
@@ -64,6 +65,8 @@
 #include "util/util.h"
 #include "undo/object_undo.h"
 
+#include <core/symbols/point_symbol.h>
+
 
 namespace OpenOrienteering {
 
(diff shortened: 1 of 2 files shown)


dl3sdo:
Libor, thank you for the variant. I personally think that the dots should be larger as they are sometimes hard to see.

I encountered some crashes with different error messages:
home/matthias/oom/mapper/src/core/renderables/renderable.cpp:273:33: runtime error: member call on address 0x5601d8d480c0 which does not point to an object of type 'Object'
/home/matthias/oom/mapper/src/core/renderables/renderable.cpp:270:24: runtime error: member call on address 0x5601d9470e10 which does not point to an object of type 'Symbol'
/home/matthias/oom/mapper/src/core/symbols/symbol.h:471:33: runtime error: member access within address 0x5601d9470e10 which does not point to an object of type 'Symbol'
0x5601d9470e10: note: object is of type 'QPinchGesture'

To reproduce the crash:
Start following a line, release the Shift key but then close the line by pressing the Return key.
The highlighting dots appear again when moving the cursor close to the line.
Start again by pressing the Shift Key and pressing the left mouse button twice (and start dragging).

https://github.com/OpenOrienteering/mapper/assets/14225666/7c6ca0a7-991b-4f59-b511-c7cda18c7c06

lpechacek:
Thanks, Matthias, for the thorough testing! I've replicated the issue and drafted a fix.

Circumventing `finishFollowing()` is the key to replicating the crash. It can be done by finishing the line with Enter key **while holding the left mouse button**. Another possibility is to abort drawing with Escape key while holding LMB again. The bellow patch makes both the code paths call `finishFollowing()` where we remove the renderables.

Ad the dot size, I agree that they are somewhat small. I also received feedback that they scale with changing magnification but that's something I cannot fix quickly.

What do you thing would be the optimal dot size in terms of an ISOM line symbol size? Width of _505 Footpath_, _503 Road_, or another size, please?

```
diff --git a/src/tools/draw_path_tool.cpp b/src/tools/draw_path_tool.cpp
index d1d30ff9429d..09965c9b3d3e 100644
--- a/src/tools/draw_path_tool.cpp
+++ b/src/tools/draw_path_tool.cpp
@@ -925,7 +925,7 @@ void DrawPathTool::finishDrawing()
        }
        
        dragging = false;
-       following = false;
+       finishFollowing();
        setEditingInProgress(false);
        if (!ctrl_pressed)
                angle_helper->setActive(false);
@@ -941,7 +941,7 @@ void DrawPathTool::finishDrawing()
 void DrawPathTool::abortDrawing()
 {
        dragging = false;
-       following = false;
+       finishFollowing();
        setEditingInProgress(false);
        if (!ctrl_pressed)
                angle_helper->setActive(false);
```

dl3sdo:
Thanks, Libor, for the fix which solves the issue for the scenarios described by you and me.

Regarding the dot size I tend to something of at least 503 Road.
I played around and increased the values to
```
dot_distance = 500;
red_dot->setInnerRadius(150);
white_dot->setInnerRadius(180);
```

lpechacek:
> Regarding the dot size I tend to something of at least 503 Road.
> I played around and increased the values to ...

I've worked with this dot size for a few days. Not bad but not good either. Dots independent of the zoom level would be great.

Screenshot from 2024-04-16 20-37-41

Issue #2196, opened 2023-12-17
Inaccurate import of images and tracks that use geographic coordinates


sfroyen:
### Steps to reproduce
1. Create a map using an accurate CRS for your area. (I used EPSG:6342 which is the current NAD83 datum for Colorado).
2. Go out with an accurate GNSS device and add some point and line features that are easily seen in images, using the Android version of OOM.
3. Copy the files from the Android device to the your non-Android OOM.
4. Import the GPX track data from the Android session. Notice that the tracks are offset compared to point and line features that were added in step 2. In my case the offset is just over 1 meter.
5. Use QGIS to capture a georeferenced Google areal image of the area that was used in step 2.
6. Add the image as a georeferenced template in OOM. Notice that the image is offset compared to point and line features that were added in step 2.

### Actual behaviour
Any georeferenced data that use geographic coordinates is offset when added as a template. 

### Expected behaviour
Georeferenced data using geographic coordinates are not offset when added as templates.

Root cause:
It appears that OOM is not specifying a default EPSG for geographic coordinates. Without a default, GDAL/PROJ uses EPSG:4326. This is an ensemble EPSG that assumes all geographic coordinate datums are the same and ignores the shifts that have accumulated over time. I have verified that this is the cause of the shifts by using QGIS to convert the images and tracks from EPSG:7667 (the 2019 datum revision) to EPSG:6342. When importing the converted files as templates, the shifts are gone.

Fix:
Hardcoding EPSG:7667 in the initialization of GDAL/PROJ would almost all of the cases. Adding a configuration setting for the EPSG code would solve any issue with (very) old data (pre 2012).

### Configuration
Mapper Version: 0.95
Operating System: MacOS Sonoma

pkturner:
I agree that EPSG:4326 is poor for geographic coordinates, and that something like EPSG:7667 is an improvement. 

It's been a long time since when I looked at this issue. Initializing PROJ differently did not appear to be an option.

krticka:
I think that accurate GNSS receivers and data will be more and more used, so it seems to me logical to stop using low precision ensemble which is based on dynamic datums. For example if you want to save any data to 4326, QGIS now warns you about limited accuracy. Thinking about this, not sure what is the best solution. What about to add parameter in Mapper settings where you can define WGS84 epoch for importing data?
BTW latest WGS 84 revision should be from 2021 (G2139).
[...]

PR #2205, opened 2024-02-10
Use WGS84 (G2139) as the geographic CRS

### LINKED ISSUE #1264: UTM coordinates in NAD83 differ from WGS84

### LINKED ISSUE #1709: GPX tracks display in different position on desktop vs. Android (regression)

pkturner:
These changes replace the geographic CRS spec string "+proj=latlong +datum=WGS84" with "EPSG:9755" for most Georeferencing purposes. EPSG:9755 is code for the current realization (G2139) of WGS84.

When creating a template with GDAL, OgrFileImport::setSRS will now recognize whether the data SRS (as detected by GDAL) is geographic and uses the ensemble/ballpark WGS84. In that case it changes the data SRS to EPSG:9755. As a result, alignment of tracks is more accurate, and tracks are aligned the same whether or not imported with GDAL.

**map file format**
The old spec string "+proj=latlong +datum=WGS84" continues to be used when Mapper saves to file. In Mapper's file format, this string appears as the map's geographic CRS and as the CRS of some templates. Because the file format is unchanged, this version of Mapper is interoperable with v0.9.5 and earlier.

However, this compatibility has unfortunate consequences.

1. Because the file has <geographic_crs_spec> elements, a Mapper that uses EPSG:9755 should put "EPSG:9755" in the file. The current PR throws away this information.
2. Future users of Mapper may need to work with old maps in a way that preserves the original alignment of tracks and images. It should be possible to distinguish old maps from new.

In the future, I plan to extend this PR with a change to the file format for new maps.

**fixes**
This is a draft, intended to address #2196 (Inaccurate import of images and tracks), along the lines suggested by @sfroyen. It fixes #1709 (Tracks positioned differently with GDAL enabled). It fixes #1264 (UTM coordinates should vary with datum).

There has been plenty of discussion of fixing these problems, along with hopes for a fix that would use PROJ to transform directly from CRS to CRS without an intermediate geographic CRS "pivot". Those efforts have not succeeded. This draft is worth considering because it is a simple change.

### DIFF
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -304,7 +304,7 @@ bool ProjTransform::isGeographic() const
 
 QPointF ProjTransform::forward(const LatLon& lat_lon, bool* ok) const
 {
-	static auto const geographic_crs = ProjTransform(Georeferencing::ballpark_geographic_crs_spec);
+	static auto const geographic_crs = ProjTransform(Georeferencing::gnss_crs_spec);
 	
 	auto point = isGeographic()
 	             ? QPointF{lat_lon.longitude(), lat_lon.latitude()}
@@ -324,7 +324,7 @@ QPointF ProjTransform::forward(const LatLon& lat_lon, bool* ok) const
 
 LatLon ProjTransform::inverse(const QPointF& projected_coords, bool* ok) const
 {
-	static auto const geographic_crs = ProjTransform(Georeferencing::ballpark_geographic_crs_spec);
+	static auto const geographic_crs = ProjTransform(Georeferencing::gnss_crs_spec);
 	
 	double easting = projected_coords.x(), northing = projected_coords.y();
 	if (geographic_crs.isValid())
@@ -376,7 +376,7 @@ ProjTransform::ProjTransform(const QString& crs_spec)
 	if (crs_spec.isEmpty())
 		return;
 	
-	static auto const geographic_crs_spec_utf8 = Georeferencing::ballpark_geographic_crs_spec.toUtf8();
+	static auto const geographic_crs_spec_utf8 = Georeferencing::gnss_crs_spec.toUtf8();
 	
 	auto crs_spec_utf8 = crs_spec.toUtf8();
 #ifdef PROJ_ISSUE_1573
@@ -386,7 +386,7 @@ ProjTransform::ProjTransform(const QString& crs_spec)
 #if defined(ACCEPT_USE_OF_DEPRECATED_PROJ_API_H) || (PROJ_VERSION_MAJOR) < 8
 	pj = proj_create_crs_to_crs(PJ_DEFAULT_CTX, geographic_crs_spec_utf8, crs_spec_utf8, nullptr);
 #else
-	static auto const geographic_crs = crs(Georeferencing::ballpark_geographic_crs_spec);
+	static auto const geographic_crs = crs(Georeferencing::gnss_crs_spec);
 	auto const projected_crs = crs(crs_spec);
 	static const char* const options[] = {"AUTHORITY=any", nullptr};
 	pj = proj_create_crs_to_crs_from_pj(PJ_DEFAULT_CTX, geographic_crs.pj, projected_crs.pj, nullptr, options);
--- a/src/core/georeferencing.cpp
+++ b/src/core/georeferencing.cpp
@@ -304,7 +304,7 @@ bool ProjTransform::isGeographic() const
 
 QPointF ProjTransform::forward(const LatLon& lat_lon, bool* ok) const
 {
-	static auto const geographic_crs = ProjTransform(Georeferencing::geographic_crs_spec);
+	static auto const geographic_crs = ProjTransform(Georeferencing::ballpark_geographic_crs_spec);
 	
 	auto point = isGeographic()
 	             ? QPointF{lat_lon.longitude(), lat_lon.latitude()}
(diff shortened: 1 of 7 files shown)


pkturner:
I added a commit which changes the WGS84 realization to the older G1762 and EPSG:9057. Evidently the version of PROJ that's in the Azure pipeline does not support EPSG:9755. The older geographic CRS provides less accurate transformation to NAD83-based georeferencing, but passes the same tests.

pkturner:
I have uploaded further development on the new-geographic-crs branch. This enhances the change so that when a user opens an existing map in Mapper, that map's templates are aligned just as in the past. If the user wishes to switch to use the more accurate alignment that comes from using EPSG:9057, that is optional. New maps would all use EPSG:9057.

The most useful versions are:

- 8869ce64 - basic fix using EPSG:9057 for all maps
- addecfed - preserves template alignment for old maps
- 68c70f59 - saves CRS spec "EPSG:9057" in map file

pkturner:
This PR is relevant to several open issues.
- #2196
- #1709
- #1264

All of these issues arise from situations when WGS84-based coordinates are converted to NAD83 coordinates, where a transformation is relevant, and leads to inaccuracy if PROJ or GDAL does not apply the transformation. Factors affecting the omission of the transformation could be
- live track vs. loaded as GPX file
- version of PROJ
- setting of whether GPX is imported using GDAL

As far as I know, the aspects of those three issues that remain are due to either

1. GDAL import of data does not apply the needed transformation.
2. Mapper built with PROJ 8 and later requires the fix from #2024. That fix is not present in any downloadable Mapper.

The present PR addresses GDAL import of GPX. It probably also takes care of images that have an associated geographic CRS, that are imported using GDAL.

pkturner:
To provide an option for the user to switch their map to the newly available accuracy of EPSG:9057, Mapper (1) provides a warning when opening an old map, and (2) asks the user to decide between the two options, when opening the Georeferencing dialog. Mapper only asks the user once. Mapper preserves that decision when the map is saved.

So that older Mappers can open the output of new Mapper, `<geographic_crs_spec>+proj=latlong +datum=WGS84</geographic_crs_spec>` will continue to be present in the map file. `<realization_spec>EPSG:9057</realization_spec>` indicates that there's more going on than just the same old ensemble `+proj=latlong +datum=WGS84`.

`<realization_spec/>` indicates that the user has chosen alignment-compatibility.

The EPSG:9057 pivot does not actually benefit non-GDAL tracks, because in Mapper v0.9.5, PROJ already finds a datum transform when converting to EPSG:6342. I tried EPSG:9755, but it failed the automated checks, presumably because the version of PROJ in the superbuild is too old.

pkturner:
One change to the Georeferencing code has a subtle rationale, which I express because in other respects this PR is straightforward.

Splitting off the ProjCRS from ProjTransform is needed because the ProjTransform class already had a split personality.  Either you construct it with `ProjTransform::crs(const QString&)` or you construct it with `ProjTransform(const QString&)`.  In the first it represents a CRS, and cannot be used to transform coordinates.  In the second the `forward` and `inverse` methods are available.

Implementation of this PR forces the split, because the second type of ProjTransform relies on a ProjTransform of the first type, named `geographic_crs`. Until now, `geographic_crs` has been a single object shared by all ProjTransforms. But now, different maps need different `geographic_crs` values, which calls for each ProjTransform of the transforming type to have its own member representing its geographic CRS.

In C++, a class (ProjTransform) can't have a regular data member of the same class.

So within ProjTransform there are distinct purposes, distinct construction, and a distinction in the data required. Splitting out the ProjCRS class is obvious. Now each ProjTransform has its own ProjCRS named `geographic_crs`.

PR #2207, opened 2024-03-10
OcdFileImport: Handle text objects without termination

### LINKED ISSUE #2206: Spurious text in title
### Steps to reproduce
1. I entered the title of a map in OOM using the Write text tool
2. I saved the map and viewed it in Condes, which I am using for course planning
on OOM:
image

### Actual behaviour
The title of the map was displayed but with other text (present elsewhere on the map) added at the end.  
I deleted the title, created a new version and got different spurious text 
On Condes:
image

### Expected behaviour
Just the title text I had entered to display

### Configuration
Mapper Version: 0.9.5
Operating System: W 10
Condes 10.5.7

dl3sdo:
As some .ocd files contain text objects without terminating null character limit the search for a terminating null character to the data area of the string.
Closes #2206

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -1978,9 +1978,16 @@ template< class O >
 QString OcdFileImport::getObjectText(const O& ocd_object) const
 {
 	auto data = reinterpret_cast<const QChar *>(ocd_object.coords + ocd_object.num_items);
+	auto maxlen = int(sizeof(Ocd::OcdPoint32) * ocd_object.num_text / 2);
 	if (data[0] == QLatin1Char{'\r'} && data[1] == QLatin1Char{'\n'})
+	{
 		data += 2;
-	return QString(data);
+		maxlen -= 2;
+	}
+	auto object_text = QString(data, maxlen);	// limit in case of faulty strings without terminating null character
+	if (object_text.indexOf(QLatin1Char{'\0'}) != -1)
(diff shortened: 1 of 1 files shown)


dg0yt:
I understood #2206 as indicating a problem with saving OCD files, not with loading OCD files.
We do fill remain space with zero, but what if we didn't reserve extra space?
https://github.com/OpenOrienteering/mapper/blob/d42dbf49ec4fea4fd9389426b57ac485614bd795/src/fileformats/ocd_file_export.cpp#L3117-L3121

dl3sdo:
Hello Kai,
you are absolutely right. The error is caused by an invalid export in Mapper which can easily be reproduced for every text object with a multiple of 32 characters.
I apologize for not having read the issue description carefully enough where it was stated that the .ocd files had been created by Mapper.
I will create a new PR (which should just contain an additional '+ 1').

lpechacek:
@dl3sdo, while reading through the patch, I noticed that `OcdFileImport::getObjectText()` functionality is implemented twice. The OCD v8 version is correct while the generic version is buggy. I suggest that we reuse the v8 code for all the format versions.

Example implementation is at https://github.com/lpechacek/mapper/commits/issue-2206-garbage-in-text/.

dl3sdo:
Libor, you are absolutely right, it makes little sense to have two implementations.

dg0yt:
Closing in favor of https://github.com/OpenOrienteering/mapper/pull/2219.

PR #2209, opened 2024-03-16
ColorListWidget: Warn about spot color removal

lpechacek:
A color entry may serve as a spot color without direct use in symbols. So far Mapper allowed silent removal of these color entries. This patch adds a warning that triggers when the user tries to remove such a spot color.

~~The problem with disabling the Overprint simulation option while the map is still displayed in overprint mode (as seen in the video further down) is not addressed by this patch. The mechanism for disabling overprint upon the last color deletion seems to be in place but does not work.~~

EDIT: The second patch fixes the problem with disabling the Overprint simulation option presented at the end of the video.

A video documenting the current behavior:
[spot-color-no-warning.webm](https://github.com/OpenOrienteering/mapper/assets/5584406/2ac06646-1295-41da-8b5d-c749642c23b0)

### DIFF
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -2002,6 +2002,7 @@ void MapEditorController::spotColorPresenceChanged(bool has_spot_colors)
 		}
 		else
 		{
+			overprintingSimulation(false);
 			overprinting_simulation_act->setChecked(false);
 			overprinting_simulation_act->setEnabled(false);
 		}
--- a/src/gui/widgets/color_list_widget.cpp
+++ b/src/gui/widgets/color_list_widget.cpp
@@ -224,10 +224,35 @@ void ColorListWidget::deleteColor()
 	if (row < 0) return; // In release mode
 	
 	// Show a warning if the color is used
-	if (map->isColorUsedByASymbol(map->getColor(row)))
+	auto const* color_to_be_removed = map->getColor(row);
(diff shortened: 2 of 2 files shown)


dl3sdo:
Libor, if you want, I would review your PR.
At first glance I stumbled across "Another map color uses....remove the spot color from these map colors!"

lpechacek:
> Libor, if you want, I would review your PR. At first glance I stumbled across "Another map color uses....remove the spot color from these map colors!"

Yes, please! The text, and maybe the code, may benefit from editing. :)

dl3sdo:
Counting and reporting the number of affected colors could be a nice addition.
See https://github.com/OpenOrienteering/mapper/commit/90e2650fdd49e313466f2a593e0e2a43de5202d1
On the other hand, this information might be unimportant for most users and thus not worth the effort (and it still lacks from using 'these' and the hope that the %n and (s) work as intended).

If you keep your originial proposal I would suggest to replace "Another map color uses this one as a spot color."
by "At least one other map color uses this color as a spot color."

lpechacek:
> Counting and reporting the number of affected colors could be a nice addition. 

LGTM. Merged into LibreMapper and squashed into this series. Thanks!

dg0yt:
My suggestion is now in GH-2222.

lpechacek:
> My suggestion is now in [GH-2222](https://github.com/OpenOrienteering/mapper/pull/2222).

Whatever solves the problem... Thanks for the enhancements!

PR #2213, opened 2024-03-30
OcdFileImport: Warn about ignored point symbol in text symbol

dl3sdo:
Text symbols in .ocd files can contain an additional point symbol that is currently silently ignored when importing such files. This change issues a detailed warning when ignoring an additional point symbol.

### DIFF
--- a/src/fileformats/ocd_file_import.cpp
+++ b/src/fileformats/ocd_file_import.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2013-2022 Kai Pastor
+ *    Copyright 2013-2024 Kai Pastor
  *
  *    Some parts taken from file_format_oc*d8{.h,_p.h,cpp} which are
  *    Copyright 2012 Pete Curtis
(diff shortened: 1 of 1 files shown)


dg0yt:
Thanks Matthias. Can you review my change? I probably don't have test data.

lpechacek:
Test data is at https://www.ocad.com/download/samples/Example_TextPointsymbol.ocd

@dl3sdo, I'd vote for a positive message in the warning text. That is replace the "Not importing" with "Removing" or "Skipping". For instance, the warning text "_In text symbol 1.1 'Text ohne Freistellung': **Skipping** unsupported reference to point symbol '1.6'._" feels more fluent to me. Otherwise, the patch looks good to me. Thanks for filling another gap in the import path!

dg0yt:
Yes, a positive wording is an excellent guideline. There should just be no doubt that information is lost.

dl3sdo:
Kai, Libor: thank you for your improvements of this PR.

dg0yt (approved):

lpechacek (approved):

Issue #2122, opened 2022-12-28
Enhancement: Show user what symbols are using a color


ryan-mooore:
When deleting a colour with symbols associated with it, the dialog will appear:

> The map contains symbols with this color. Deleting it will remove the color from these objects! Do you really want to do that?

Firstly notice the inconsistency between _symbols_ and _objects_ in the dialog text.

Additionally, Mapper provides zero indication of which symbols are going to have colors deleted, which I think is essential information to the user. Why even show me a dialog saying that some symbols are going to be affected, if I have no idea which ones?

One solution to this frustration would be to right click on a symbol with an option "_Select all symbols with this color_", for example. In my opinion this is the most elegant solution as it is consistent with other options and right click menus within the program and makes intuitive sense.

dl3sdo:
Indeed, the map rather contains objects and uses symbols.
The map information tool would provide you the option to determine the usage of colors by symbols:
https://github.com/OpenOrienteering/mapper/issues/2018#issuecomment-1016875926

dl3sdo:
Now as draft pull request #2125

ryan-mooore:
That looks good, but I would prefer to have a right-click action additionally.

PR #2222, opened 2024-04-13
Warn on spot color deletion

dg0yt:
Modification of GH-2209, GH-2235

- Don't use more than `overprinting_simulation_act` in `MapEditorController::spotColorPresenceChanged`.
- Add and use iteration functions for colors and symbols (long-standing wishlist item :innocent:)
- Show the color name in the spot color removal message. Feels more comfortable when confirming.

### DIFF
--- a/src/core/map.h
+++ b/src/core/map.h
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012-2014 Thomas Schöps
- *    Copyright 2013-2020 Kai Pastor
+ *    Copyright 2013-2020, 2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -489,6 +489,16 @@ public:
 	 */
 	bool hasAlpha() const;
 	
+	/**
+	 * Applies a const operation on all colors which match a particular condition.
+	 */
+	void applyOnMatchingColors(const std::function<void (const MapColor*)>& operation, const std::function<bool (const MapColor*)>& condition) const;
+	
+	/**
+	 * Applies a const operation on all colors.
+	 */
+	void applyOnAllColors(const std::function<void (const MapColor*)>& operation) const;
+	
 	
 	// Symbols
 	
@@ -578,6 +588,16 @@ public:
 	 * display the symbols indicated by the bitfield because of symbol dependencies.
 	 */
 	void determineSymbolUseClosure(std::vector< bool >& symbol_bitfield) const;
+	
+	/**
+	 * Applies a const operation on all symbols which match a particular condition.
+	 */
+	void applyOnMatchingSymbols(const std::function<void (const Symbol*)>& operation, const std::function<bool (const Symbol*)>& condition) const;
+	
+	/**
+	 * Applies a const operation on all symbols.
+	 */
+	void applyOnAllSymbols(const std::function<void (const Symbol*)>& operation) const;
 	
 	/**
 	 * Returns the scale factor to be used for default symbol icons.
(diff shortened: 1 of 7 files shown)


src/gui/widgets/color_list_widget.cpp
```diff
@@ -224,12 +224,28 @@ void ColorListWidget::deleteColor()
...
+	if (color_to_be_removed->getSpotColorMethod() == MapColor::SpotColor)
+	{
+		if (auto color_in_use = map->countSpotColorUsage(color_to_be_removed))
+		{
+			if (QMessageBox::warning(this, tr("Confirmation"),
+			                         tr("%n map color(s) use spot color \"%1\"."
+			                            " Deleting it will remove the spot color from these map colors."
+			                            " Do you really want to do that?",
+			                            nullptr,
+			                            color_in_use).arg(color_to_be_removed->getName()),
```
dl3sdo, line 243: color_in_use).arg(color_to_be_removed->getSpotColorName()),

src/gui/map/map_editor.cpp
```diff
@@ -2002,7 +2002,8 @@ void MapEditorController::spotColorPresenceChanged(bool has_spot_colors)
 		}
 		else
 		{
-			overprinting_simulation_act->setChecked(false);
+			if(overprinting_simulation_act->isChecked())
```
dl3sdo, line 2005: if (overprinting_simulation_act->isChecked())

dl3sdo (changes requested):

src/gui/map/map_editor.cpp
dg0yt: Oh, I'm struggling with CMake code style ;-)

dg0yt:
TBH I'm unhappy with possibly two dialogs appearing. I think it could make use of the [detailedText](https://doc.qt.io/qt-5/qmessagebox.html#detailedText-prop). This would give more room for text, or maybe even for listing the symbols and colors (https://github.com/OpenOrienteering/mapper/issues/2122).

dl3sdo (approved):

src/core/map.h
```diff
@@ -579,6 +589,16 @@ friend class XMLFileExporter;
 	 */
 	void determineSymbolUseClosure(std::vector< bool >& symbol_bitfield) const;
 	
+	/**
+	 * Applies a const operation on all colors which match a particular condition.
```
dl3sdo, line 593: * Applies a const operation on all symbols which match a particular condition.

src/core/map.h
```diff
@@ -579,6 +589,16 @@ friend class XMLFileExporter;
...
+	 */
+	void applyOnMatchingSymbols(const std::function<void (const Symbol*)>& operation, const std::function<bool (const Symbol*)>& condition) const;
+	
+	/**
+	 * Applies a const operation on all colors.
```
dl3sdo, line 598: Applies a const operation on all colors.

src/core/map.h
dl3sdo, line 593: Applies a const operation on all symbols which match a particular condition.

dl3sdo (changes requested):

src/core/map.h
dg0yt:
```suggestion
	 * Applies a const operation on all symbols which match a particular condition.
```

src/core/map.h
dg0yt:
```suggestion
	 * Applies a const operation on all symbols.
```

dl3sdo (approved):

PR #2224, opened 2024-04-14
OcdFileImport: Revise path coord processing

dg0yt:
Resolves #2144: Importing accepting pairs of hole points, but editor crashing when selecting the export.

This PR moves some single-use helper functions back into the processing loops. This allows optimizing the conditions and adding the desired fix. (The original functions might have been more easy to analyze separately, but passing and modifying `pos` in the call stack added complexity.) 

The PR is broken into a series of commits which allow to track the source code transformation. There is no specific unit test to cover this change.


dl3sdo:
I'm currently writing an unit test.

dl3sdo:
Kai, I had to rebase your PR to base it on the latest version of file_format_t.
I was not sure whether I should force-push to your branch, so I pushed to https://github.com/dl3sdo/mapper/tree/ocd-hole-points

Please check whether the unit test goes into the right direction.

I'm currently only testing area objects and I don't test all the flags. Although I crafted the .ocd coords to be real coordinates I intentionally don't check the x- and y-values of the coordinates after conversion.

dg0yt:
> https://github.com/dl3sdo/mapper/tree/ocd-hole-points

I will pick the commits which add the actual test and work on it separately first.

dg0yt:
Moving to this test data pattern:
~~~C++
	QTest::addColumn<OcdPointsView>("data");
	QTest::addColumn<FlagsView>("expected");
	
	{
		static Ocd::OcdPoint32 ocd_points[] = {
		    // +0,  bezier
		    { C(-1109), C(212) },
		    { C(-1035) | Ocd::OcdPoint32::FlagCtl1, C(302) },
		    { C(-1008) | Ocd::OcdPoint32::FlagCtl2, C(519) },
		    { C(-926), C(437) },
		};
		static int expected_flags[] = {
		    MapCoord::CurveStart, 0, 0, 0, MapCoord::ClosePoint,
		};
		QTest::newRow("bezier") << OcdPointsView(ocd_points) << FlagsView(expected_flags);
	}
~~~
Those view classes are litte more than [ size, pointer ].

src/fileformats/ocd_file_import.cpp
```diff
@@ -2116,43 +2116,51 @@ Object* OcdFileImport::importRectangleObject(const Ocd::OcdPoint32* ocd_points,
...
+				else
+				{
+					// HolePoint needs to be applied to the last point of a part
+					object->coords[i-1].setHolePoint(true);
+				}
+			}
+		}
+		return ++i;
+	});
+	Q_ASSERT(object->coords.size() >= count);
```
dl3sdo, line 2162: Q_ASSERT(object->coords.size() >= static_cast<std::size_t>(count));

dg0yt:
Plan:
- [ ] Test behavior-preserving changes
- [ ] Adjust test to desired changes of behavior
- [ ] Test changes

PR #2225, opened 2024-04-14
Fix "empty hole" quirks

dg0yt:
The crash fixed in GH-2224/GH-2144 had an OCD import aspect (data source) and an map editor aspect (data usage). This PR adresses the second aspect. While it was easy to locate the crash and its trigger, a confusing fact was that the path part data in the map editor wasn't the same as the data observed while debugging the OCD import. It furned out that the initial calculation of map parts in `PathObject::recalculateParts()` (expected data) was overwritten by `PathObject::updatePathCoords()` (actual data).

This PR aligns map part calulation in the second function with the first function, and a adds a basic test. This fixes the crash. 

There is a certain risk that this creates new problems. But the generated data is fairly reasonable now.

### DIFF
--- a/src/core/virtual_path.cpp
+++ b/src/core/virtual_path.cpp
@@ -89,20 +89,17 @@ VirtualCoordVector::size_type PathCoordVector::update(VirtualCoordVector::size_t
 	{	
 		Q_ASSERT(virtual_coords.size() == flags.size());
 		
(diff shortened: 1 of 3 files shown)


dg0yt:
This wasn't meant to be merged so quickly, but...

PR #2231, opened 2024-04-15
Fix spelling of 'if' in comments

dl3sdo:


dg0yt:
Not an error: `iff` means "if and only if".

dl3sdo:
That explains the frequent occurence and why your previous spellchecks skipped those words.

PR #2232, opened 2024-04-16
Various fixes

dl3sdo:
Fix include paths in two .cpp files.

Fix compiler warnings and remove superfluous QString inclusion.

### DIFF
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -88,7 +88,7 @@
 
 namespace OpenOrienteering {
 
-Q_STATIC_ASSERT(Georeferencing::declinationPrecision() == Util::InputProperties<Util::RotationalDegrees>::decimals());
+Q_STATIC_ASSERT(static_cast<int>(Georeferencing::declinationPrecision()) == Util::InputProperties<Util::RotationalDegrees>::decimals());
 
 
 namespace  {
--- a/src/gui/scaling_icon_engine_plugin.cpp
+++ b/src/gui/scaling_icon_engine_plugin.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2020 Kai Pastor
+ *    Copyright 2020-2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -22,7 +22,7 @@
 
 #include <QIconEnginePlugin>
 
-#include "scaling_icon_engine.h"
+#include "gui/scaling_icon_engine.h"
 
 namespace OpenOrienteering {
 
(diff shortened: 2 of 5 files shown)


src/gui/scaling_icon_engine_plugin.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include <QIconEnginePlugin>
 
-#include "scaling_icon_engine.h"
+#include "gui/scaling_icon_engine.h"
```
dg0yt, line 25: I guess this is proposed for some sort of consistency. But for `#include "foo"`, "typical implementations first search the directory where the current file resides". so the original form seems appropriate.

src/undo/undo_manager.cpp
```diff
@@ -41,7 +41,7 @@
 
 namespace OpenOrienteering {
 
-Q_STATIC_ASSERT(UndoManager::max_undo_steps < std::numeric_limits<int>::max());
+Q_STATIC_ASSERT((int)UndoManager::max_undo_steps < std::numeric_limits<int>::max());
```
dg0yt, line 44: That's a type of cast we avoid in modern C++. I can't add a proposal right now. I would need to check which limit and conversion is actually guarded here.

dg0yt (commented):
Just quick feedback.

src/gui/scaling_icon_engine_plugin.cpp
dl3sdo: Yes, since we always prepend the path for any header file which is not the one of the source file itself, I consider that change a matter of style and consistency.

src/undo/undo_manager.cpp
dl3sdo: My initial thought was to remove that check completely. It merely checks that the value 'UndoManager::max_undo_steps' (which is actually 128) is less than the maximum number of an integer.

dl3sdo: Do you prefer
`Q_STATIC_ASSERT(static_cast<int>(UndoManager::max_undo_steps) < std::numeric_limits<int>::max());`
instead?

src/gui/georeferencing_dialog.cpp
```diff
@@ -88,7 +88,7 @@
 
 namespace OpenOrienteering {
 
-Q_STATIC_ASSERT(Georeferencing::declinationPrecision() == Util::InputProperties<Util::RotationalDegrees>::decimals());
+Q_STATIC_ASSERT(static_cast<int>(Georeferencing::declinationPrecision()) == Util::InputProperties<Util::RotationalDegrees>::decimals());
```
dl3sdo, line 91: Also addressed by #2027

src/undo/undo_manager.cpp
```diff
@@ -41,7 +41,7 @@
 
 namespace OpenOrienteering {
 
-Q_STATIC_ASSERT(UndoManager::max_undo_steps < std::numeric_limits<int>::max());
+Q_STATIC_ASSERT(static_cast<int>(UndoManager::max_undo_steps) < std::numeric_limits<int>::max());
```
dg0yt, line 44: max_undo_steps is of `std::size_t` which could be to big for `int`.
But `std::numeric_limits<int>::max()` is never to big for `unsigned int`.

dg0yt:
Taken to https://github.com/OpenOrienteering/mapper/pull/2264.

PR #2235, opened 2024-04-24
Show color usage and warn on spot color deletion

dl3sdo:
When a user wants to delete a color, the warning message was ambiguous by mixing up symbols and objects (see #2127).
This PR merges #2127, #2222, #2209.
It shows one warning dialog if

- the color is used by symbols
- the color is used as a spot color by other colors.

In addition the warning messages specifies the number of symbols, objects and other map colors that will be affected by the color removal.
A more detailed warning is available when removing spot colors.

### DIFF
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012-2014 Thomas Schöps
- *    Copyright 2013-2020 Kai Pastor
+ *    Copyright 2013-2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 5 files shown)


dl3sdo:
Screenshot1

dl3sdo:
Kai, Libor: What do you think? Does this PR meet your functional expectations?

dg0yt:
I couldn't look at it in detail. 
- There is a lot of extra code in a single GUI class method which might be better separated into non-GUI functions with clear interfaces.
- The translatable strings must form whole sentences or paragraphs. My suggestion:
  - General message:
    ~~~c
    tr("Color \"%1\" is used by other elements. Removing the color will change the appearance of these elements. Do you really want to remove it?").arg(color_to_be_removed->getName());
    ~~~
  - Details:
    ~~~c
    if (<used_by_colors>)
    {
        tr("This color is used as a spot color by the following map colors:")
        <list>
    }
    if (<used_by_symbols>)
    {
        tr("This color is used by the following symbols:")
        <list of symbol_name: n objects>
    }
~~~

dl3sdo:
@dg0yt: Following your suggestions:
Screenshot2

dg0yt:
Integrated and merged with #2222. Tnx!

PR #2237, opened 2024-04-26
ColorListWidget: Fix table column resize mode

dl3sdo:
The individual setting 'QHeaderView::Stretch' of the 'Name' column's resize mode was erroneously overwritten by the setting which was intended for the 'Spot color' column.
Since 'QHeaderView::Stretch' results in undesired behaviour the default setting 'QHeaderView::Interactive' is kept.


dl3sdo:
Kai, I tested your commit and it looks good to me.

dl3sdo (commented):
Kai, I tested your commit and it looks good to me.

PR #2245, opened 2024-06-21
OcdFileExport: Harmonize warning messages

dl3sdo:
Warning messages were not consistent regarding whether
- the symbol name is quoted (text symbols were unquoted)
- they end with a dot
- details are separated by a comma or a colon


dl3sdo:
Regarding consistency:
.ocd import has a different style, e.g.
_In line symbol 862.0 'Treppe': Skipped secondary point symbol._

dg0yt:
- This affects translations. Without manual edits, existing translations are no longer used, immediately.
 - `OcdFileImport` has overloads `addSymbolWarning` for (OCD) symbol types which use per-type variants of `"In area symbol %1 '%2': %3"` to decouple the repeated text from the individual details. Maybe this is a pattern to adopt and adopt in other places.

Issue #1042, opened 2018-02-02
"World of O Maps" link is not working


sikmir:
![screenshot1](https://user-images.githubusercontent.com/688044/35748010-8b3d0254-085d-11e8-871b-3059a0e6a475.jpg)

[World of O Maps](http://maps.worldofo.com/?zoom=15&lat=56.0565&lng=40.4398):
> 19/11/2013: Due to changes with Google Maps the overview map is currently disabled and only the search interface works (Note that the advanced search interface lets you search by position). It will be looked at within a few weeks time. Sorry for the inconvenience - please tell me if this was an important function for you at jan [at] kocbach.net!

dg0yt:
Yes, this "a few weeks" since Nov 2013. However, it might be possible to switch to the search form if we provide a rectangle:
http://maps.worldofo.com/searchmaps.php?name=&owner=&country=&region=&sourceid=&year=&minlat=48.5&minlng=8.5&maxlat=49&maxlng=9&over=&mapmaker=&sortby=name

dg0yt:
The new search form doesn't help that much: It will center & zoom to the known maps it finds for the given rectangle, not to the rectangle itself. And it won't show anything if doesn't find a known map at all.

ollesmaps:
Question from Jan Kocbach: "Do you know what this is actually used for? Does anybody use the old map archive? Most want the omap archive, and that is a different link...?"

dg0yt:
I used to find this useful to find nearby maps...

What's "the omap archive"? http://omaps.worldofo.com/? Note that I had to look for the term "omap" on http://worldofo.com/ to find that link - the "Maps" button/menu does link to http://maps.worldofo.com. I also see no way to load a particular location.

ollesmaps:
"Omaps" are maps with routes uploaded by orienteers (usually very good resolution). "Maps" are maps uploaded by federations - usually archives of existing maps - only very few countries have that (usually bad resolution or just an outline without the real map).

ollesmaps:
I believe Jan can supply a link if we specify what we want. I prefer omaps to maps - bigger variety, more uptodate and bigger resolution.

sikmir:
A lot of O-maps (primarly for Russia) can be found on https://nakarte.me, so osm and worldofo links can be replaced with:
`https://nakarte.me/#m=18/%1/%2&l=O/U/R` (layers: O - OpenStreetMap, U - Races, R - O-sport)

[...]

PR #2246, opened 2024-06-21
GeoreferencingDialog: Remove link to WorldofO

dl3sdo:
Since the link to 'World of O' to lookup maps is not working for quite a long time and as this service is apparently no longer provided, remove the link by commenting it out in the code.

### DIFF
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2020 Kai Pastor
+ *    Copyright 2012-2020, 2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -395,6 +395,7 @@ void GeoreferencingDialog::projectionChanged()
 	QString osm_link =
 	  QString::fromLatin1("https://www.openstreetmap.org/?mlat=%1&mlon=%2&zoom=18&layers=M").
 	  arg(latitude, 0, 'g', 10).arg(longitude, 0, 'g', 10);
+#ifdef MAPPER_WITH_WORLDOFO_LINK
 	QString worldofo_link =
 	  QString::fromLatin1("http://maps.worldofo.com/?zoom=15&lat=%1&lng=%2").
 	  arg(latitude).arg(longitude);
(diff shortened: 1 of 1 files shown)


src/gui/georeferencing_dialog.cpp
```diff
@@ -48,13 +48,11 @@
 #include <QPushButton>
 #include <QRadioButton>
 #include <QSignalBlocker>
-#include <QSize>
```
dg0yt, line 51: Still used with size hints.

src/gui/georeferencing_dialog.cpp
```diff
@@ -48,13 +48,11 @@
...
 #include <QSpacerItem>
 #include <QStringRef>
 #include <QTimer>
 #include <QUrl>
 #include <QUrlQuery>
-#include <QVariant>
```
dg0yt, line 57: I guess this is still used, but I can't check now.

src/gui/georeferencing_dialog.cpp
```diff
@@ -395,13 +393,14 @@ void GeoreferencingDialog::projectionChanged()
 	QString osm_link =
 	  QString::fromLatin1("https://www.openstreetmap.org/?mlat=%1&mlon=%2&zoom=18&layers=M").
 	  arg(latitude, 0, 'g', 10).arg(longitude, 0, 'g', 10);
-	QString worldofo_link =
+/*	QString worldofo_link =
```
dg0yt, line 396:
```suggestion
#ifdef MAPPER_WITH_WORLDOFO_LINK
	QString worldofo_link =
```

src/gui/georeferencing_dialog.cpp
```diff
@@ -395,13 +393,14 @@ void GeoreferencingDialog::projectionChanged()
...
 	  QString::fromLatin1("http://maps.worldofo.com/?zoom=15&lat=%1&lng=%2").
 	  arg(latitude).arg(longitude);
 	link_label->setText(
 	  tr("<a href=\"%1\">OpenStreetMap</a> | <a href=\"%2\">World of O Maps</a>").
 	  arg(osm_link, worldofo_link)
-	);
+	);*/
+	link_label->setText(tr("<a href=\"%1\">OpenStreetMap</a>").arg(osm_link));
```
dg0yt, line 403:
```suggestion
	);
#else
	link_label->setText(tr("<a href=\"%1\">OpenStreetMap</a>").arg(osm_link));
#endif
```

dg0yt (commented):
We can probably remove that link completely.

src/gui/georeferencing_dialog.cpp
dl3sdo: You are right

src/gui/georeferencing_dialog.cpp
dl3sdo: You are also right, it's used by Settings::getInstance().getSetting().

dg0yt (approved):

dg0yt:
Resolves https://github.com/OpenOrienteering/mapper/issues/1042.

PR #2255, opened 2024-07-06
GeoreferencingDialog: Fix declination lookup

dl3sdo:
The declination lookup at www.ngdc.noaa.gov requests an additional parameter 'key'. OpenOrienteering Mapper is entitled due to registration to use the key which is also used by the Magnetic Field Calculators' homepage.

This PR is a straightforward version of #2088 and thus might supersede it.

### DIFF
--- a/src/gui/georeferencing_dialog.cpp
+++ b/src/gui/georeferencing_dialog.cpp
@@ -452,6 +452,7 @@ void GeoreferencingDialog::requestDeclination(bool no_confirm)
 	
 	QUrlQuery query;
 	QDate today = QDate::currentDate();
+	query.addQueryItem(QString::fromLatin1("key"), QString::fromLatin1("zNEw7"));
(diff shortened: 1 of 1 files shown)


lpechacek (approved):
Ack on this one. I believe that this straightforward implementation suffices.

FTR, I took a slightly different route in LibreMapper (https://github.com/LibreMapper/mapper/commit/6c2fd6c7475538120f8198282021a6de145bb4c0), giving the user opportunity to enter an alternate key when the declination lookup fails. On the other hand, I doubt that the key will change without an advance notice. That makes the extra code mostly useless if not harmful. Time will show...

dl3sdo:
Additional information:
[International Geomagnetic Reference Field (IGRF)](https://www.ncei.noaa.gov/products/international-geomagnetic-reference-field)

dl3sdo:
Closed in favor of https://github.com/OpenOrienteering/mapper/pull/2344

PR #2261, opened 2024-07-19
MapEditorController: Disable menu item if no symbols are selected

dl3sdo:
The 'Select all objects with selected symbols' menu item below the 'Edit' menu was enabled even if no symbols were selected. The same menu item in the popup symbol menu of the symbol pane however was disabled if no symbols were selected.
Move the enabling or disabling of the menu item to a function which is called by a change of the selected symbols and disable the menu item if no symbols are selected.

### DIFF
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -454,7 +454,6 @@ void MapEditorController::setEditingInProgress(bool value)
 		select_all_act->setEnabled(!editing_in_progress);
 		select_nothing_act->setEnabled(!editing_in_progress);
 		invert_selection_act->setEnabled(!editing_in_progress);
-		select_by_current_symbol_act->setEnabled(!editing_in_progress);
 		find_feature->setEnabled(!editing_in_progress);
 		
 		// Map menu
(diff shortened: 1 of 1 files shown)


lpechacek (approved):
LGTM.

FWIW, I'd drop the menu entry altogether. Why? It *partly* duplicates the functionality of the symbol widget context menu. I'm wondering why one function appears at two places and not the others.

Well, I admit that I'm biased because I've just spent an extra half an hour on testing the code change just to learn that it's not about the context menu "Select by symbol" but the other item about which I didn't know. :)

dg0yt:
Generally there should be just a single action reused across the UI (main menu, toolbar, context menu).

dl3sdo:
I agree that removing it from the Edit menu is more appropriate indeed of improving its appearance, especially since it's still different as the menu item in the symbol widget context menu would even change from 'Select all objects with this symbol' to 'Select all objects with selected symbols' which the Edit menu's version does not (yet).

Thus I will create a new PR that removes this Edit menu entry, ok?

lpechacek:
@dl3sdo I've removed myself from the assignee role. I've tested and reviewed the change but I'm not going to do more (e.g. merge) on it. Thanks!

dl3sdo:
@lpechacek: thank you for spending time on that issue.

dg0yt:
Thanks to all.

dg0yt:
> Thus I will create a new PR that removes this Edit menu entry, ok?

The Edit menu isn't too cluttered ATM.

This particular menu item is probably more discoverable if you already selected some *object in the map* and want to select all objects with the same symbol.

dl3sdo:
> The Edit menu isn't too cluttered ATM.
> 
> This particular menu item is probably more discoverable if you already selected some _object in the map_ and want to select all objects with the same symbol.

Indeed, it works and it's not a bug, so fine for me keeping it.

PR #2264, opened 2024-07-29
Fix compiler warnings

dg0yt:


dl3sdo (commented):
For code-check-wrapper.sh changes:
I could not find `boolean_tool.cpp`
Is `file_format_registry.cpp` missing?
`kml_course_export`, `iof_course_export` should rather have a .cpp but how about replacing them and `simple_course_export.cpp` by `course_export.cpp`

dl3sdo:
Another observation for various fixes:
In kml_course_export.h the
`#ifndef OPENORIENTEERING_KML_EXPORT_EXPORT_H`
should rather be
`#ifndef OPENORIENTEERING_KML_COURSE_EXPORT_H`

dg0yt:
> Is `file_format_registry.cpp` missing?

It is unclear if this is intentational... As mentioned before, we should have an exclusion list instead.

> `kml_course_export`, `iof_course_export` should rather have a .cpp

It is a simple pattern match. No .cpp required. In fact some patterns are intentionally short. We could have `course_export`.

dl3sdo:
> Another observation for various fixes: In kml_course_export.h the `#ifndef OPENORIENTEERING_KML_EXPORT_EXPORT_H` should rather be `#ifndef OPENORIENTEERING_KML_COURSE_EXPORT_H`

Addressed by #2265

PR #2265, opened 2024-07-29
Fix include guards

dl3sdo:
Fix some include guards in fileformats folder.
I noticed that the comment after the final #endif is mostly missing so I'm not going to further fix it.


dg0yt:
> I noticed that the comment after the final #endif is mostly missing so I'm not going to further fix it.

Yeah, this is something for drive-by fixes.

PR #2124, opened 2023-01-10
Import of graphic, layout and image objects

dl3sdo:
Graphic, layout and image objects are stored in .ocd without dedicated symbols.
While graphic objects re-use existing symbol colors, image and layout objects store the object color inside the object.
Mapper imports these special objects by defining auxiliary symbols and colors.
Layout image objects, i.e.,objects that reference externally stored images cannot yet be imported, however, a warning is now issued instead of silently discarding them.

Closes #959
Closes #1231

PR #2266, opened 2024-07-29
Ignore virtual gaps in line objects

dl3sdo:
Line objects in .ocd files can contain virtual gaps, i.e., gaps that interrupt the line but keep the object as a single object.
Mapper does not yet support virtual gaps and thus ignores them when importing .ocd files.
Issue a warning that virtual gaps in line objects are not supported.
The current PR is a WIP as it depends on merging #2224.
If there a multiple objects with virtual gaps, only one warning should be issued. This is being accomplished by the first commit (which is a modified version of the one in #2124).


dl3sdo:
Behaviour with current PR:
VirtualGaps2Imported

dl3sdo:
Prerequisites:
- [X] Merge #2270
- [ ] Adapt #2224 acc. to #2270
- [ ] Merge #2224

PR #2269, opened 2024-07-30
Fix #endif location

dl3sdo:


dl3sdo:
These bugs didn't lead to a compile error since

- `print_progress_dialog.h` is only included once, so the include guard protection never comes into effect. If you unnecessarily include it e.g. in `core/map_printer.h` the compilation fails.
- `print_tool.h` is also included only once. Here the failure occurs if `QT_PRINTSUPPORT_LIB` is not defined. However, in `print_tool.cpp` `print_tool.h` is not included if `QT_PRINTSUPPORT_LIB` is not defined, so the error is masked and never comes into effect.

PR #2275, opened 2024-08-22
CI: Migrate to DownloadPipelineArtifact@2 task

dl3sdo:
React to warning that the DownloadPipelineArtifact@0 task is deprecated and that the latest version of the DownloadPipelineArtifact task shall be used instead.

### DIFF
--- a/ci/setup-common.yml
+++ b/ci/setup-common.yml
@@ -64,7 +64,8 @@ steps:
     buildVersionToDownload: 'specific'
     pipelineId: 2865
 #    pipelineId: $(openorienteering.superbuild.buildId)
-    definition: 1
+    definition: $(openorienteering.superbuild.definitionId)
+#    definition: 1
     artifactName: 'superbuild-$(SUPBERBUILD_IMAGE_NAME)$(OUTPUT_SUFFIX)'
     targetPath: $(SUPERBUILD_INSTALL_DIR)
   displayName: 'Download superbuild artifacts'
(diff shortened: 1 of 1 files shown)


ci/setup-common.yml
```diff
@@ -57,9 +57,13 @@ steps:
...
   displayName: 'Update environment'
 
-- task: DownloadPipelineArtifact@0
+- task: DownloadPipelineArtifact@2
   inputs:
-    pipelineId: $(openorienteering.superbuild.buildId)
+    buildType: 'specific'
+    project: $(System.TeamProjectId)
+    buildVersionToDownload: 'latest'
+#    pipelineId: $(openorienteering.superbuild.buildId)
```
dg0yt, line 65: This is the part which is left to resolve: Pick a particular build, not the `latest`.
(IIRC that's why I didn't update this task yet.)

dl3sdo: If I understand you correct, specifying a particular build instead of the latest one ensures that any yet unstable change of superbuild won't break mapper builds. I now pick the pipeline run that you marked by setting its retention to forever. However, if there is a stable update of superbuild and thus a new pipeline run, then Mapper's setup-common.yml would need to be updated.
BTW: How is the variable openorienteering.superbuild.buildId set? Is this always pointing to the latest build?

dl3sdo:
With the current approach of hardcoding the `pipelineId` this PR needs an update in the course of updating superbuild due to #2284.

dg0yt:
Resolved with #2291. Thanks Matthias.

PR #2277, opened 2024-09-15
ISOM 2017 2 revision 6 update

lpechacek:
This is an evolutionary step of the Mapper native symbol set.


lpechacek:
Too much pain. Giving up.

Issue #2175, opened 2023-08-03
Enhancement: Map color window - allow moving multiple colors at once


lpechacek:
### Steps to reproduce
1. Import multiple symbols which introduce new colors. The new colors are placed at the bottom of the list.
2. Try to move a block of the new colors higher in the table.

### Actual behaviour
Currently the colors have to be moved one by one.

### Expected behaviour
It would be great if multiple colors could be selected at once and moved in the color list.

### Configuration
Mapper Version: 0.9.5
Operating System: all

PR #2289, opened 2024-10-27
ColorListWidget: Remove unused colors

### LINKED ISSUE #381: Purge unused colours button?
**gandym01 reported on Sourceforge [[tickets:#381](https://sourceforge.net/p/oorienteering/tickets/381)]:**

In the colour window, I would appreciate a button that allows the user to purge unused colours.

I know that the _Replace symbols_ tool already partially does this, however I do not know of any convenient way to delete colours that were used by symbols that have since been manually deleted.

gandym

dl3sdo:
Add a button to the Colors window that allows to remove colors which are not used by any symbol. Before the removal a popup window offers to show the list of colors to be removed and asks for a confirmation.

Closes #381, #2073

### DIFF
--- a/src/gui/widgets/color_list_widget.cpp
+++ b/src/gui/widgets/color_list_widget.cpp
@@ -1,6 +1,7 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
  *    Copyright 2012-2018, 2021, 2024 Kai Pastor
+ *    Copyright 2025 Matthias Kühlewein
  *
  *    This file is part of OpenOrienteering.
  *
@@ -21,12 +22,15 @@
 
 #include "color_list_widget.h"
 
+#include <vector>
+
 #include <Qt>
 #include <QtGlobal>
 #include <QAbstractButton>
 #include <QAbstractItemView>
 #include <QAction>
 #include <QApplication>
+#include <QChar>
 #include <QColor>
 #include <QDialog>
 #include <QFlags>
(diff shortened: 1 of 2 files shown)


dl3sdo:
https://github.com/user-attachments/assets/8e7bab23-505f-4b7f-9fab-c4c6a42b5746

dg0yt:
I wonder if it was more useful to have an action to just *select* unused colors. This would allow to see them in context. And the Delete button could remove all selected colors.

dl3sdo:
Currently the color window does not allow to select more than one color. A multi-color selection increases the complexity if the user does not just press the delete button but instead decides to move or add a color: you would need to either disable those buttons or handle it in a special way (move all selected, but add after first selected or refuse to add?). Can multi-color selection be limited to the _select unused_ button action to avoid additional complexities?
And you replace one button (I admit that _Cleanup_ is not the most intuitive label but the shortest one, _Remove unused_ could be a good alternative) by another one.

I doubt that the additional effort is worth the slightly increased gain in giving an easy overview, especially for a feature which is in most cases never or only rarely used.
The implementation itself was easy after implementation of #2222.

lpechacek:
> Currently the color window does not allow to select more than one color.

As a side note, we've got enhancement request #2175 which is about moving multiple colors at once. I started the work but it's not yet finished. If nothing else, I've got vector versions of the green arrows for "move color up/down" and "move color to the top/bottom". I can share the individual bits if needed.
Screenshot From 2024-10-27 20-23-59

Issue #291, opened 2014-03-13
OCD export broken for long path objects


dg0yt:
**Private ticket on Sourceforge [[tickets:#291](https://sourceforge.net/p/oorienteering/tickets/291)]**

Path objects which have more than 32768 points exceed the limit of the OCD format version 8. Unfortunately, this case is not caught in Mapper at the moment.

PR #2290, opened 2024-10-31
OcdFileExport: Fix OCD v8 object size calculation

lpechacek:
Although OCD v8 is an old one, Mapper is the only software left to support writing these files. In my opinion, bugs are worth fixing in this part of the code.

Closes GH-291.


dg0yt:
macOS CI: Flaky CPack step, unrelated. Re-run must wait while I try to integrate the general CI updates.

dg0yt (approved):
LGTM. Thanks Libor.

dl3sdo:
Will close #291

PR #2295, opened 2024-11-09
MapEditorController: Delete object in destructor

dl3sdo:
The pointer variables `gps_display`, `gps_track_recorder`, `compass_display` and `gps_marker_display` are all zero-initialized in the constructor. In `MapEditorController::detach()` all of their associated objects are deleted and the variables are set to nullptr. In the destructor however the `gps_marker_display` object was not deleted whereas the other three objects were.


dg0yt:
In normal operation, `detach` will be called before the destructor, and the object will be already deleted (and the pointer set to `nullptr`. But I don't see harm from symmetry with the other gadgets.

PR #2296, opened 2024-11-19
ObjectSelector: Refactoring

dl3sdo:
Fix TODO for sortObjects() by embedding into the new function sortObjectSelection().

### DIFF
--- a/src/tools/object_selector.cpp
+++ b/src/tools/object_selector.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2013-2017 Kai Pastor
+ *    Copyright 2013-2017, 2024 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
@@ -21,14 +21,8 @@
 #include "object_selector.h"
 
 #include <algorithm>
-#include <cstddef>
-#include <map>
-#include <memory>
-
-#include <QRectF>
 
 #include "core/map.h"
-#include "core/map_part.h"
 #include "core/objects/object.h"
 
 
@@ -41,12 +35,11 @@ ObjectSelector::ObjectSelector(Map* map)
 }
 
 
-
 bool ObjectSelector::selectAt(const MapCoordF& position, qreal tolerance, bool toggle)
 {
 	bool selection_changed;
 	
-	bool single_object_selected = map->getNumSelectedObjects() == 1;
+	const bool single_object_selected = map->getNumSelectedObjects() == 1;
 	Object* single_selected_object = nullptr;
 	if (single_object_selected)
 		single_selected_object = *map->selectedObjectsBegin();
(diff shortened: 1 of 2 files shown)


src/tools/object_selector.cpp
```diff
@@ -161,19 +154,22 @@ bool ObjectSelector::selectBox(const MapCoordF& corner1, const MapCoordF& corner
...
+		[](const std::pair<int, Object*>& a, const std::pair<int, Object*>& b) {
+			if (a.first != b.first)
+				return a.first < b.first;
+			
+			auto a_area = a.second->getExtent().width() * a.second->getExtent().height();
+			auto b_area = b.second->getExtent().width() * b.second->getExtent().height();
+			return a_area < b_area;
+		}
+	);
 }
```
dg0yt, line 169: We could have done that from the beginning, but we didn't. Embrace `std::sort` and the custom operator, don't add `sortObjectSelection`.

src/tools/object_selector.cpp
```diff
@@ -161,19 +154,22 @@ bool ObjectSelector::selectBox(const MapCoordF& corner1, const MapCoordF& corner
...
 
 
-bool ObjectSelector::selectionInfosEqual(const SelectionInfoVector& a, const SelectionInfoVector& b)
+bool ObjectSelector::selectionInfosEqual(const SelectionInfoVector& a, const SelectionInfoVector& b) const
```
dg0yt, line 172: This could probably be `static`, not just `const`.

dl3sdo: Or even move it into an anonymous namespace?

src/tools/object_selector.cpp
```diff
@@ -82,7 +88,7 @@ bool ObjectSelector::selectAt(const MapCoordF& position, qreal tolerance, bool t
 		{
 			// Results different - select object with highest priority, if it is not the same as before
 			last_results = objects;
-			std::sort(objects.begin(), objects.end(), sortObjects);
+			sortObjects(objects);
```
dg0yt, line 91: I would like to make it explicit: I want `std::sort` in these lines. 
But I acknowledge that the original `sortObject` identifier causes confusion. Maybe it should be `compareTypeAndExtent`.

dl3sdo: Ok, keeping `std::sort` was my preference as well but I misunderstood your "Embrace std::sort and the custom operator".

dg0yt:
I don't want to remove it from the public surface. The TODO comment is just about where it should have its natural home. It doesn't ask for hiding it.

dl3sdo:
Since this function is not used by any other class I think that it should remain in `ObjectSelector` to keep local stuff local.
I don't mind whether it remains a static function or changes to a lambda expression, so the first commit will provide the lambda expression, whereas the second reverts it to a static function.
Please feel free to just close this PR to keep the current state.

PR #2308, opened 2024-12-13
TextSymbolSettings: Fix initial state of standard buttons

dl3sdo:
When entering the text symbol settings dialog the 'OK' and 'Reset' buttons where initially enabled.
Don't misuse the `framingCheckClicked()` function to enable/disable the `framing_widget` tab in `updateGeneralContents()` as `framingCheckClicked()` will signal modified properties.


dg0yt:
IWYU is somewhat sensitive to versions. I tried to reflect the actual usage.

IMO the problematic pattern was restoring `react_to_change` to `true` instead of to the original value, given nested scopes of `react_to_change` being temporarily `false`. But this PR dissolves a nested scope which was causing a problem. Fine, thanks.

dg0yt (approved):

Issue #2083, opened 2022-08-22
Measure tool does not warn about minimum sizes on objects with a combined symbol


lpechacek:
### Steps to reproduce
1. Create undersized (<0.5 mm²) paved area with a bounding line (501 in Mapper's default symbol set).
2. Active Measure lengths and areas tool.
3. Switch the area to symbol Paved area (501.1).

### Actual behaviour
The tool does not warn about the small size in step 2. After step 3, it issues the minimum area size warning.

### Expected behaviour
The measure tool should warn when the object size is smaller than *any* of the assigned symbol areas. That is, in the case of combined symbols, it should examine all area settings in the embedded symbols.

### Configuration
Mapper Version: master-v20201227.4-200-g0fda53dfd200
Operating System: openSUSE Tumbleweed

PR #2309, opened 2024-12-26
Consider minimum area and length in combined symbols

dl3sdo:
Combined symbols may contain areas with a minimum size, either in area symbols or private sub-symbols.
Combined symbols may contain lines with a minimum size, either in line symbols or private sub-symbols.
This is an alternative implementation of GH-2084.

Closes GH-2083.


dl3sdo:
PR looks good to me. Thank you for the improvements Kai.

dg0yt (approved):
I will squash-merge this when CI finished.

Issue #2316, opened 2025-01-15
Map is always visible after loading


dl3sdo:
### Steps to reproduce
1. Disable map visibility
2. Save map
3. Load map

### Actual behaviour
The map is always visible after loading.

Looking at the source code explains why this happens:
```
if (map_element.hasAttribute(literal::visible))
   map_visibility.visible = map_element.attribute<bool>(literal::visible);
else
   map_visibility.visible = true;
```
`XmlElementWriter::writeAttribute(const QLatin1String& qualifiedName, bool value)`
won't write a visible attribute being `false`, thus `map_element.hasAttribute` will only be `true` if the visible attribute is `true`.
If the visible attribute shall be correctly read then a plain `map_visibility.visible = map_element.attribute<bool>(literal::visible);` does the job. If it shall be ignored instead then a plain `map_visibility.visible = true;` does the job.

It's unclear to me what the desired behaviour shall be:

- shall the visibility be loaded as saved?
- or shall a map always appear visible after loading?

lpechacek:
> shall the visibility be loaded as saved?

This option IMHO best fits the principle of the least surprise. It aligns the map layer behavior with the other ones.

PR #2317, opened 2025-01-16
MapView: Restore map visibility after loading

dl3sdo:
Although the map visibility is stored in the map file the map became always visible after loading.

Amends commit 701902d (MapView/MapWidget: Review, fast pinching support).
Closes GH-2316 (Map is always visible after loading).

### DIFF
--- a/src/core/map_view.cpp
+++ b/src/core/map_view.cpp
@@ -34,6 +34,7 @@
 
 #include "core/map.h"
 #include "core/map_coord.h"
+#include "fileformats/xml_file_format.h"
 #include "gui/util_gui.h"
 #include "templates/template.h" // IWYU pragma: keep
 #include "util/util.h"
@@ -43,7 +44,7 @@
 namespace literal
 {
 	static const QLatin1String zoom("zoom");
-	static const QLatin1String rotation("rotation");
+	//static const QLatin1String rotation("rotation");
 	static const QLatin1String position_x("position_x");
 	static const QLatin1String position_y("position_y");
 	static const QLatin1String grid("grid");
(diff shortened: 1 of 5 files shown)


pkturner:
With commit 701902d, Mapper changed from writing `visible="false"` for the unusual case, to `visible="true"` for the usual case. The present PR changes Mapper's behavior when loading a map with no `visible` attribute. This can occur either with recent maps that have an invisible map view, or with maps created prior to Mapper 0.5.97 that have a visible map.

pkturner:
It is disconcerting to load a map, with the result being a blank view. Often the "Show whole map" action provides the remedy, for example if the view somehow became centered far from the map's objects. 

The change in this PR will occasionally lead to a blank view. Perhaps as part of the same PR, the "Show whole map" action should force the map layer to be visible.

dl3sdo:
@pkturner: thank you for the analysis. I try to summarize it:
 ```
                                       map was visible      map was hidden
map file <0.597 attribute                                      "false"
mapper 0.95 loads <0.597 file            => visible           => hidden
mapper 0.95 patch loads <0.597 file      => hidden            => hidden

map file 0.95 attribute                     "true"                     
mapper 0.95 loads 0.95 file              => visible           => visible
patch 0.95 loads 0.95 file               => visible           => hidden

```

dg0yt:
So AFAICS the behavior with this PR is right for "new" maps, and wrong "hidden" for some "old" maps.
Do we need to take some action for those old maps?

At the time of https://github.com/OpenOrienteering/mapper/commit/701902d5a066bb68e00d8a8e0533c4cd10bac228, i.e. 2014, the global version attribute was at `5`.  
It was bumped to `6` in 974fb826, i.e. 2016.  
A possible mitigation might be to always enable visibility for version <= `5`.  
This would still mean there are some old maps with wrong visibility, but now a wrong "visible".

dl3sdo:
@dg0yt: LGTM, thank you.

PR #2319, opened 2025-01-17
Map: Minor Refactoring

dl3sdo:


dl3sdo:
After I modified the function Map::countObjectsInRect I noticed that this function is apparently not invoked, at least my search for 'countObjectsInRect' was not successful.

Issue #2320, opened 2025-01-17
Deleting objects corrupts undo step


dl3sdo:
For version 0.95 as well as for later development builds (master branch) there are scenarios
with moving parts of the map, adding an map part, deleting some objects and merging all map parts which
result in a map state where undo operations won't recover the initial state, see:

https://github.com/user-attachments/assets/c6098d34-d741-4424-a2ac-b50e66be8112

https://github.com/user-attachments/assets/0f0a8b9d-634c-4bc8-8dc2-b2f44c7db4a8

dl3sdo:
Minimum reproducible scenario (again using 'forest sample'):

1. move some objects
2. add a new map part
3. switch back to original part
4. delete some objects
5. perform as many undos as possible => some objects that were moved in the first step remain at their moved position

dl3sdo:
I'm stunned, but there is an even more simple scenario:

1. load  'forest sample'
2. move some objects
3. delete some objects
4. perform two undos => some objects that were moved in the 2nd step remain at their moved position

dl3sdo:
I suspect the root cause in `void Map::deleteSelectedObjects()`
Releasing the objects in the `for` loop while still adding other objects to the undo step will lead to duplicated index values being added to the `modified_objects` list in `object_undo`.
I assume that this scenario is the reason for this function (being used by `CutoutOperation::finish()`):
```
/**
 * Removes all contained objects from the map.
 * 
 * This can be useful after constructing the undo step,
 * as it is often impractical to remove the objects directly
 * when adding them as the indexing would be changed this way.
 */
void removeContainedObjects(bool emit_selection_changed);
```
This change seems to solve the issue:
```
		for (; obj != end; ++obj)
		{
			int index = part->findObjectIndex(*obj);
			if (index >= 0)
			{
				undo_step->addObject(index, *obj);
				//part->releaseObject(index);
			}
			else
			{
				qDebug() << this << "::deleteSelectedObjects(): Object" << *obj << "not found in current map part.";
			}
		}
		
		undo_step->removeContainedObjects(false);
		//setObjectsDirty();
		clearObjectSelection(true);
		push(undo_step);
```

PR #2322, opened 2025-01-21
Map: Fix deleting object selection

dl3sdo:
Deleting a selection of objects lead to a corrupted undo step which could lead to a failure when undoing the undo step before. The reason was that the selected objects were added in a loop to the undo step including their object index. After adding an object it was removed which caused the indexes of the remaining (and sometimes to be removed as well) objects to be renumbered.
Use the AddObjectsUndoStep::removeContainedObjects() function to remove the objects once they all have been added.

Closes GH-2320 (Deleting objects corrupts undo step).


dg0yt (commented):
LGTM. Tnx Matthias.

Issue #2324, opened 2025-01-24
Undefined behavior in `SplitPathCoord::at(length_type, SplitPathCoord)`


dg0yt:
Crashing in debug builds:

Use ISOM 2017-2 symbol 517 Ruined fence,  
change Mid symbols settings to Center of gaps, 
3 mid symbols per spot and increase mid symbol distance.  
The symbol editor will crash Mapper while increasing to 0.xx mm.  
`ASSERT: "p >= 0.0f" in file path_coord.cpp, line 471`

_Originally posted by @dl3sdo in https://github.com/OpenOrienteering/mapper/issues/2099#issuecomment-2611068947_

dg0yt:
- `assert` is inactive in release builds.
- `SplitPathCoord::at(length_type len, SplitPathCoord first)` is defined to search from first. It is not defined what should happen when len is negative, before first, or after the end.

dl3sdo:
I thought about whether it makes sense to either inhibit or indicate in the line symbol editor a combination of number of mid symbols and their distance which exceeds the gap, however:

- most users just use provided symbols sets
- users editing a symbol set should know what they do and they will immediately see the result
- Mapper will not crash

Therefore I'm fine with your proposal, Kai.

dg0yt:
#2325 (merged) is one change for this issue. 
https://github.com/OpenOrienteering/mapper/pull/2331 is another relevant PR but I don't want to merge without more testing. (It refers to #2325 but I think it was meant to refer to #2324.) The assertion is only enabled in debug builds.

PR #2325, opened 2025-01-24
LineSymbol: Revise CenterOfGap symbols

dg0yt:
For GH-2324.
No change necessary or implemented for the intented use case, which is symbols which fit into the gap between the dashes.
For symbols outside the gap, this PR avoids undefined rendering behavior signaled by an assert in SplitPathCoord:at by controlling rendering based symbols being outside (omitted) or inside (rendered) of the [start,end] range passed into `createDashGroups`. This interval is a moving window over start, explicit dash points, and end of a path.
The new behavior also creates consistent mid symbol positions regardless of path direction.


test/path_object_t.cpp
```diff
@@ -1206,6 +1207,94 @@ void PathObjectTest::recalculatePartsTest()
...
+		{
+			// undefined: before split.clen
+			QEXPECT_FAIL("", "unexpected", Continue);  // actual: 1.5
+			QCOMPARE(SplitPathCoord::at(-0.2f, split).clen, 0.0f);
+			QEXPECT_FAIL("", "unexpected", Continue);  // actual: 1.5
+			QCOMPARE(SplitPathCoord::at(0.2f, split).clen, 0.0f);
+			QEXPECT_FAIL("", "unexpected", Continue);  // actual: 1.5
+			QCOMPARE(SplitPathCoord::at(1.0f, split).clen, 1.0f);
+			QEXPECT_FAIL("", "unexpected", Continue);  // actual: 1.5
+			QCOMPARE(SplitPathCoord::at(1.2f, split).clen, 1.2f);
```
dg0yt, line 1278: Note that for this curve, the result is different vs. the straight case.

dg0yt:
It would probably be best to degrade gracefully for cases before the search start.

But cases outside the path are tricky: You could reasonably argue to force them into the valid range. OTOH this may break loops which are naively limited by reaching a desired length.

src/core/path_coord.cpp
```diff
@@ -467,8 +467,9 @@ SplitPathCoord SplitPathCoord::at(
 			
 			if (prev_coord.index == first.index)
 			{
+				Q_ASSERT(split.param >= first.param);
+				Q_ASSERT(1.0 >= first.param);
 				auto p = (split.param - first.param) / (1.0f - first.param);
```
dg0yt, line 472: This branch is suspicious: 
`first` is not `prev_coord`, even when the point to the same `index`.

dg0yt:
This PR now includes a test for actual rendering.
The test of `SplitPathCoord::at` will be in a different PR, connected with different behavioral changes.

dg0yt:
Reduced this PR to changing only linesymbol, i.e. the calling site triggering the particular assert. Changes to, and tests of, `SplitPathCoord::at` are more sensitive and go to a different PR.

src/core/symbols/line_symbol.cpp
```diff
@@ -1231,15 +1231,25 @@ SplitPathCoord LineSymbol::createDashGroups(
...
-					auto split = dash_start;
+					// 'dash_start' is now the begin of the gap.
+					// 'CenterOfGap' is meant for symbols placed in the gap only.
+					// But actual required length can be longer. Output symbols
+					// from 'start' to 'end' and omit everything outside this range.
+					auto split = (position >= dash_start.clen) ? dash_start : start;
 					for (int i = 0; i < mid_symbols_per_spot; ++i, position += mid_symbol_distance_f)
 					{
+						if (position < start.clen)
+							continue;
```
dg0yt, line 1247: Key changes for avoiding the `assert`.

src/core/symbols/line_symbol.cpp
```diff
@@ -1231,15 +1231,25 @@ SplitPathCoord LineSymbol::createDashGroups(
...
+						
 						split = SplitPathCoord::at(position, split);
+						if (position > end.clen)
+							break;
```
dg0yt, line 1251: For symmetry. Noticable when changing the direction of the path.

dl3sdo (approved):
Successfully tested with dashes either outside the gap, outside the dash or even outside the path.
Switching the dash direction no longer changes the placement of dashes.
Code changes look good to me.

PR #2327, opened 2025-01-28
Refactoring image template default parameters

dl3sdo:
Resolves a known TODO.

### DIFF
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -486,7 +486,7 @@ void Map::init()
 	
 	printer_config.reset();
 	
-	image_template_defaults = {true, 0.0, 0.0, 0.0};
+	image_template_defaults = {};
 	
 	colors_dirty = false;
 	symbols_dirty = false;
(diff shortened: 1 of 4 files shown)


dl3sdo:
Kai, thank you for improving the PR. I was also not satisfied with the long variable names.

dg0yt:
Ideally, the details of "image template defaults" shouldn't be in the `Map` sources, but I would really prefer to do (and review) not so many nice-to-have changes to core code at the moment, and bring back to to what deserves more priority.

dl3sdo:
> Ideally, the details of "image template defaults" shouldn't be in the `Map` sources, but I would really prefer to do (and review) not so many nice-to-have changes to core code at the moment, and bring back to to what deserves more priority.

I agree. Would you be fine with adding a TODO comment for a later change and merge the current state? It's still better compared to what was previously in.

dg0yt (approved):

dg0yt:
I don't think we need another TODO comment.

PR #2331, opened 2025-02-02
SplitPathCoord: Ensure monotony and bounds for `at`

dg0yt:
For #2325. Test first.
Unlike #2325, this touches the behavior of a core function. While the change is reasonable IMHO, this may need more testing. Other parts of Mapper might implicitly rely on the broken cases.

### DIFF
--- a/src/core/path_coord.cpp
+++ b/src/core/path_coord.cpp
@@ -388,7 +388,10 @@ SplitPathCoord SplitPathCoord::at(
 	const auto& flags  = path_coords.flags();
 	
 	SplitPathCoord split = first;
-	split.path_coord_index = path_coords.upperBound(length, first.path_coord_index, first.path_coords->size()-1);
+	if (length > first.clen)
+	{
+		split.path_coord_index = path_coords.upperBound(length, first.path_coord_index, first.path_coords->size()-1);
+	}
 	if (split.path_coord_index > first.path_coord_index)
 	{
 		// New path coordinate, really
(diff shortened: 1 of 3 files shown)


src/core/path_coord.cpp
```diff
@@ -388,7 +388,10 @@ SplitPathCoord SplitPathCoord::at(
 	const auto& flags  = path_coords.flags();
 	
 	SplitPathCoord split = first;
-	split.path_coord_index = path_coords.upperBound(length, first.path_coord_index, first.path_coords->size()-1);
+	if (length > first.clen)
+	{
+		split.path_coord_index = path_coords.upperBound(length, first.path_coord_index, first.path_coords->size()-1);
+	}
```
dg0yt, line 394: Fix for monotony.

src/core/path_coord.cpp
```diff
@@ -403,10 +406,12 @@ SplitPathCoord SplitPathCoord::at(
...
 		auto factor = 1.0f;
 		if (qFuzzyCompare(1.0f + length, 1.0f + current_coord.clen) ||
-		    qFuzzyCompare(1.0f + curve_length, 1.0f))
+		    qFuzzyCompare(1.0f + curve_length, 1.0f) ||
+		    length > current_coord.clen)
 		{
 			// Close match at current path coordinate,
-			// or near-zero curve length.
+			// or near-zero curve length,
+			// or length exceeding path length.
```
dg0yt, line 414: Fix for upper bound.

src/core/path_coord.cpp
```diff
@@ -506,9 +511,10 @@ SplitPathCoord SplitPathCoord::at(
 		{
 			--split.path_coord_index;
 		}
+		
+		split.index = path_coords[split.path_coord_index].index;
 	}
 	
-	split.index = path_coords[split.path_coord_index].index;
```
dg0yt, line 511: Drive-by change. When not entering the `if` branch, the right `split.index` is implied by initialization of `split`.

PR #2335, opened 2025-02-16
Licensing: Consider derivative distributions

akikoskinen:
The `$ID_LIKE` variable in `/etc/os-release` file may list space separated names that the current distribution is similar to. For example in some random distribution that derives from Ubuntu the values could be:

    ID="random-distro"
    ID_LIKE="ubuntu debian"

When trying to guess the current system, use all those values in order and use the first one that is known.

Here's documentation for the `ID_LIKE` variable: https://www.freedesktop.org/software/systemd/man/latest/os-release.html#ID_LIKE=

### DIFF
--- a/doc/licensing/CMakeLists.txt
+++ b/doc/licensing/CMakeLists.txt
@@ -36,23 +36,19 @@ endforeach()
 if(ANDROID OR APPLE OR WIN32)
 	set(default_provider OFF)
 elseif(UNIX)
+	set(default_provider NOTFOUND)
 	execute_process(
 	  COMMAND sh -c ". /etc/os-release; echo \"$ID $ID_LIKE\""
-	  OUTPUT_VARIABLE system_candidates
+	  OUTPUT_VARIABLE system_ids
 	  OUTPUT_STRIP_TRAILING_WHITESPACE 
 	)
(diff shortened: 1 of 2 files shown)


dg0yt:
That's nice. :heart

dg0yt:
I will push some nitpicks which I cannot add as suggestions.

dg0yt:
FTR Ubuntu has:
~~~
ID=ubuntu
ID_LIKE=debian
~~~
So with this change, we no longer need the ubuntu-licensing.cmake script which forwards to debian. We just cannot remove it immediately. We can migrate the setting for a while.
~~~
[cmake] -- Licensing information provided by: ubuntu
[cmake] CMake Warning at doc/licensing/ubuntu-licensing.cmake:1 (message):
[cmake]   LICENSING_PROVIDER=ubuntu is deprecated.  Switching to debian.
[cmake] Call Stack (most recent call first):
[cmake]   doc/licensing/CMakeLists.txt:160 (include)
~~~

dg0yt:
If there are no objections to my modifications, I will merge this.

akikoskinen:
LGTM

dg0yt:
Thank you very much for you contribution!

Issue #2334, opened 2025-02-05
Compile error with Clang 19


sikmir:
```sh
       > In file included from /tmp/nix-build-OpenOrienteering-Mapper-0.9.5.drv-0/source/3rd-party/cove/libvectorizer/Vectorizer.cpp:39:
       > /tmp/nix-build-OpenOrienteering-Mapper-0.9.5.drv-0/source/3rd-party/cove/libvectorizer/Concurrency.h:120:80: error: overload resolution selected deleted operator '='
       >   120 |         Job& operator=(Job&& job) noexcept { future = std::move(job.future); progress = job.progress; }
       >       |                                                                              ~~~~~~~~ ^ ~~~~~~~~~~~~
       > /tmp/nix-build-OpenOrienteering-Mapper-0.9.5.drv-0/source/3rd-party/cove/libvectorizer/Concurrency.h:63:12: note: candidate function has been explicitly deleted
       >    63 |         Progress& operator=(const Progress&) = delete;
       >       |                   ^
       > /tmp/nix-build-OpenOrienteering-Mapper-0.9.5.drv-0/source/3rd-party/cove/libvectorizer/Concurrency.h:64:12: note: candidate function not viable: expects an rvalue for 1st argument
       >    64 |         Progress& operator=(Progress&& p) = delete;
       >       |                   ^         ~~~~~~~~~~~~
       > 1 error generated.
```

dg0yt:
Approximation in compiler explorer: https://godbolt.org/z/4Ebf3vxMM

dg0yt:
Can you test the changes from https://github.com/OpenOrienteering/mapper/pull/2337?

PR #2337, opened 2025-02-21
Progress,Job: Fix construction and assignment

dg0yt:
Move construction of Progress: No problem, sharing from const ref.  
Copy construction of Job: No problem.  
Assignment of Progress or Job: Not possible.  
Fixes GH-2334.


sikmir (approved):
It works for me. Thanks!

PR #2338, opened 2025-02-21
TagSelectWidget: Change column width

dl3sdo:
The 'Find objects' dialog contains a table which is shown in the 'Query editor' view.
The 'Relation' and 'Comparision' columns were set up with a fixed width based on the width of the translated labels 'Relation' and 'Comparision'. The result looks good for English but not for other languages, e.g., German or Czech.
Set up both columns to automatically resize based on its contents.

### DIFF
--- a/src/gui/widgets/tag_select_widget.cpp
+++ b/src/gui/widgets/tag_select_widget.cpp
@@ -52,6 +52,16 @@ namespace OpenOrienteering {
 
 namespace {
 
+QTableWidgetItem* createPlaceholderItemFor(QWidget* widget)
+{
+	auto* item = new QTableWidgetItem();
+	item->setFlags(Qt::NoItemFlags);
+	item->setBackground(QBrush(QGuiApplication::palette().window()));
+	if (widget)
+		item->setSizeHint(widget->sizeHint());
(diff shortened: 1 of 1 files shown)


dl3sdo:
Example (Windows):
Previous_englishPrevious_german
Previous_czech
Czech: streching the window does not help either.

dl3sdo:
The first commit fixes the issue that the beginning of Key and Values fields may be hidden permanently.
However, there is a strange behaviour for the 'Relation' combo box:

https://github.com/user-attachments/assets/4a28f5b3-2a2f-458a-9054-7cc75ae7276b

dg0yt:
What we have in master is that newly added lines look better than previously added lines.

You can test in plain English if temporarily shorten the source strings.

dl3sdo:
I notice that this issue does not occur on my Linux development build (which is based on QT 5.15).
I wonder whether this is a problem of Windows or of QT 5.12 (which is used by our CI).

dg0yt (approved):

dl3sdo:
@dg0yt: thank you for refactoring the code. I tested it successfully under Windows.

Issue #2341, opened 2025-03-04
Feature: Search for dynamic object properties


dl3sdo:
Search for dynamic object properties (also mentioned in #2211), e.g.,:

- object violates minimum area or line length constraints
- object area size / object line length
- object creation and modification dates

This is a WIP, current state:

- calculation of object area size / object line length and constraint violation checks moved from measure widget to object base class
- 'Delete & Find next' functionality added to 'Find objects' dialog
- search for objects violating minimum area or line length constraints

Please have a look on the demo where I 'misuse' the minimum line length to search and delete small contour line artifacts (since not having implemented a real comparison of object area size / object line length yet)

https://github.com/user-attachments/assets/1a2d9ab0-c5b9-47c2-8fd4-312945696f0b

I personally like the 'Delete & Find next' functionality to easily check an object before deleting it when stepping through the found objects.
What do you think of this functionality, do you want to have it in Mapper?

krticka:
This function is great enhancement! Few notes:

- because screen is not centered to new object with every find next click, it can be difficult to identify feature on the screen. I propose to add flashing of feature. For example frame around feature can have distinct glow and flash three times quickly
- I expect that for areas you will use AreaTooSmall, but other attributes to query would be beneficial. For example to query $geometry type and also to have option to handle/select objects without symbol which can come from OCAD and it is not easy to fix it efficiently in Mapper.
- friendliness of use - I would love to see intelligent query menu 1) you will define WHERE, you can enter symbol number but dropdown list with symbols including names will be also available, 2) you will define OPERATOR (is equal to, is greater than,...), 3) you will define VALUE(s) and another clause with AND/OR can be used
- future - I can imagine future application in enhanced legibility check which checks distances between objects of the same color. For example you can query $color and buffer distance to identify vertices which are too close.

dl3sdo:
Hello @krticka,
thank you for your feedback and your proposals:

- I agree, I find it difficult as well. I have not found a Mapper function to center the screen to a selected object. Improving the usability is on my list.
- You are right ;-) _AreaTooSmall_ is exactly the keyword. I added a new keyword _UndefinedSymbol_ to seach for objects without symbol. I did not extend the SYMBOL keyword to something like SYMBOL UNDEFINED due to the special implementation by Mapper (the specified symbol number is changed to an internal symbol address, something that could be handled differently, but that decision is up to Kai).
Can you explain what you mean with $geometry type  ?
- Indeed, there is room for improvements.
[...]

PR #2343, opened 2025-03-07
Refactoring minimum area and length checks

dl3sdo:
Move the calculation and constraint checks of the area and length of
an object from MeasureWidget to Object class.
Add a test.


dl3sdo:
Split the tasks to implement #2341 in smaller and distinct pieces, facilitating reviewing and testing.

dl3sdo:
Although the test could be moved to path_object_t.cpp (PathObjectTest) we could decide to keep object_t.cpp and add other test cases later (e.g., when adding creation and modification dates to objects).

dg0yt:
> Although the test could be moved to path_object_t.cpp (PathObjectTest) we could decide to keep object_t.cpp and add other test cases later (e.g., when adding creation and modification dates to objects).

Yeah, I understood why it is separate, and we can keep it for now.

dl3sdo:
Kai, thank you for the updates. It's an improvement of the documentation and it's really better to use 'calculate' for calculations and 'get' for getter functions.
The usage of `std::accumulate` will make the code smarter but will most likely increase runtime since most objects don't contain holes. However, since `calculatePaperArea()` will only be invoked by the 'Measure lengths and areas' tool or in future by the 'Find' mechanism, this does not really matter.
I agree to the proposed changes, please proceed.

PR #2351, opened 2025-03-22
TagsWidget: Correct link to help file

dl3sdo:
There is not (yet) a specific help file just for the object tag editor, but the editor
is covered by 'object_tags.html'.


dg0yt:
The

> There is not (yet) a specific help file just for the object tag editor, but the editor is covered by 'object_tags.html'.

Hm, no?
https://www.openorienteering.org/mapper-manual/pages/object_tags.html

dg0yt (approved):
Got it.

PR #2353, opened 2025-03-24
TagsWidget: Disable editing if multiple objects are selected

dl3sdo:
If multiple objects were selected the tag editor would just operate on the so-called 'first selected object'.
If a group of objects is selected the user does not know which of them is the first selected object.

### DIFF
--- a/src/gui/widgets/tags_widget.cpp
+++ b/src/gui/widgets/tags_widget.cpp
@@ -126,13 +126,14 @@ void TagsWidget::objectTagsChanged()
 	
 	react_to_changes = false;
 	
-	const auto* object = map->getFirstSelectedObject();
-	if (map->getNumSelectedObjects() == 1 && object)
+	if (map->getNumSelectedObjects() == 1)
 	{
-		int row = 0;
+		const auto* object = map->getFirstSelectedObject();
 		auto const& tags = object->tags();
 		tags_table->clearContents();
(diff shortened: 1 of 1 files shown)


dl3sdo:
Allowing tag editing when multiple objects are selected also could give the user the false impression that bulk editing of tags is supported by Mapper.

src/gui/widgets/tags_widget.cpp
```diff
@@ -126,10 +126,10 @@ void TagsWidget::objectTagsChanged()
...
-	int row = 0;
-	const Object* object = map->getFirstSelectedObject();
-	if (object)
+	const auto* object = map->getFirstSelectedObject();
+	if (map->getNumSelectedObjects() == 1 && object)
 	{
+		int row = 0;
 		auto const& tags = object->tags();
 		tags_table->clearContents();
 		tags_table->setRowCount(tags.size() + 1);
```
dg0yt, line 135:
```suggestion
	if (map->getNumSelectedObjects() == 1)
	{
		const auto* object = map->getFirstSelectedObject();
		auto const& tags = object->tags();
		tags_table->clearContents();
		tags_table->setRowCount(tags.size() + 1);
		
		int row = 0;
```

dg0yt (approved):
When we have selected objects, there must exist a first selected object.

PR #2355, opened 2025-03-29
Feature: Remove object tags

### LINKED ISSUE #2354: Feature: Remove objects tags
Add possibility to selectively remove object tags in order to reduce file size and memory consumption.
This feature was proposed by @krticka in #2341.

Implementation proposal:

https://github.com/user-attachments/assets/fbd28fcb-fc78-4d76-8a8d-4cc68a5379c6

The proposed implementation follows Mapper's typical behaviour:

- operate on selected objects (thus limit removing to current map part)
- allow to undo the removal (since adding an undo step does not reduce file size and memory consumption, the option is not active by default)

If all object tags shall be removed the user needs to

- switch to every map part
- select all objects of that part
- and remove all object tags by selecting a non-existent key (e.g., !) in combination with 'not equal to'

dl3sdo:
Add dialog to remove object tags from selected objects that match a configurable pattern.
Allow to create an undo step.

Closes GH-2354 (Remove objects tags).

### DIFF
--- a/code-check-wrapper.sh
+++ b/code-check-wrapper.sh
@@ -104,6 +104,7 @@ for I in \
   symbol_rule_set.cpp \
   symbol_t.cpp \
   symbol_tooltip.cpp \
+  tag_remove_dialog.cpp \
   tag_select_widget.cpp \
   /template.cpp \
   template_image.cpp \
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -160,6 +160,7 @@ set(Mapper_Common_SRCS
   gui/select_crs_dialog.cpp
   gui/settings_dialog.cpp
   gui/simple_course_dialog.cpp
+  gui/tag_remove_dialog.cpp
   gui/task_dialog.cpp
   gui/text_browser_dialog.cpp
   gui/touch_cursor.cpp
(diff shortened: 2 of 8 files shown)


dl3sdo:
@dg0yt: I'm not sure whether using a 'trash' button in the tag editor widget is the right way for invoking the tag removal dialog. OTOH I did not find a suitable location in the menu and I don't want to overload Mapper's menu.

dg0yt (commented):
I think *you* want at least to move `TagRemoveDialog` to a different filename and to add it to the code check wrapper.

I'm not yet happy with the UI, but I can't propose actual improvements at the moment.

dl3sdo:
@dg0yt: Since I placed the action in the Map menu now should I move the .cpp and .h files to the gui/map folder (where e.g. map information and map notes are...but not the georeferencing dialog)?

dl3sdo:
There is currently a decoupling of the tags and objects being found (and shown) by the Find button and the actual deletion by the Remove button: If the user either changes the comparison operator or the key value and does not click Find again, the object tags that would be removed are not the ones actually being shown but those based on the current search parameters.

Proposal: Only remove those object tags that are the result of the search initiated by Find (and disable the Remove button if either the comparison operator or the key value field have changed).

Issue #103, opened 2012-09-24
Toolbar in color


dg0yt:
**oskarlin reported on Sourceforge [[tickets:#103](https://sourceforge.net/p/oorienteering/tickets/103)]:**

It would be a lot more easy to see which tools that are available and which isn't if the tools available were more colorful for clickable tools and grayed out for those not available. Right now they're only a bit lighter when not clickable which could be confusing.

PR #2356, opened 2025-04-01
Add a color tint to otherwise gray toolbar icons, makes easier to separate from disabled icons

elektronisk:
The icon set is not perfect, but this is an efficient fix for issue #103. Chose blue because it is already used by many of the icons.


dl3sdo:
@elektronisk: I noticed that the color depth of the icons was reduced from 32 bits per pixel to just 8 bits per pixel. For me outlines appear less smooth than before,

PR #2357, opened 2025-04-01
Add color to the otherwise grayscale menu icons to distinguish from disabled icons

### LINKED ISSUE #103: Toolbar in color

elektronisk:
This fixes #103 . I chose blue because it is already used by many of the icons.
This time retaining the 32 bit per pixel color depth :-)


dg0yt (approved):

dl3sdo (approved):

dg0yt:
Thank you very much for this improvement @elektronisk.

dg0yt:
It is really awesome. I enjoy the new look everytime a open new Mapepr window. :heart:

PR #2360, opened 2025-04-07
Add new box zoom tool, icon and cursor. Hook it into existing GUI.

### LINKED ISSUE #229: Box Zooming
**02JanDal reported on Sourceforge [[tickets:#229](https://sourceforge.net/p/oorienteering/tickets/229)]:**

A tool to select an area (rectangular), and then the view would be zoomed (and panned) so that the entire "selected" area would be visible and fill the screen as much as possible

elektronisk:
Fixes #229 .

This works fine for me on windows, but I'm unable to test it on tablet.

Note: the assertion in tool_base.cpp is commented because the assertion could be triggered by starting a drag operation (holding left mouse button), then clicking the right mouse button to bring up the pie context menu, and then letting go of the left mouse button, and starting a new drag. I'm not sure what the best way is to deal with the situation. Should I be disabling the pie context menu?

image

image

https://github.com/user-attachments/assets/e7101f68-085f-41f8-8c2f-0bac40eb8aba


dg0yt:
Thank you for the contribution.

> I'm unable to test it on tablet.

The new action won't be in the mobile UI until we add it. You can switch to the mobile UI on the PC by enabling "Touch mode" on the welcome screen.

> Note: the assertion in tool_base.cpp is commented because the assertion could be triggered by starting a drag operation (holding left mouse button), then clicking the right mouse button to bring up the pie context menu, and then letting go of the left mouse button, and starting a new drag. I'm not sure what the best way is to deal with the situation. Should I be disabling the pie context menu?

You might need to look at when to actual cancel the drag operation. Maybe as soon as another button is pressed.
(Shouldn't there be a similar challenge for drag-to-select?)

dg0yt:
> That worked. However I have a hunch that screenspace on a tablet might be constrained, and I don't have a good feel if there is available space for this tool or not. What do you think? Also, should a shortcut key be assigned?

Maybe it is not necessary to add it to the mobile UI (now) because it has an alternative: Use two fingers.

dl3sdo:
@elektronisk: thank you for your PR, I like the feature.
What I noticed  is that the zoom frame is not always visible, it depends whether objects were selected before:

https://github.com/user-attachments/assets/8a5ceea1-e7e7-4ae5-9cf2-59f270f186ea

dl3sdo:
@elektronisk: Tests are now successful, thank you for the fix.

I notice that the zoom tool remains active until the user selects another tool.
Personally I would prefer some easy way to finish zooming (and to change to the previous drawing tool?), e.g. by a left double-click.
What do you think?

dg0yt:
Dear Andreas,
thank you very much for your submission. I modified (and simplified) the code a little bit:
- It is now a tool with "checked state". Maybe it is not entirely what we want, but this is how it worked already, and I don't see a good way to make it a one-shot tool at the moment.
- I added canceling the operation with the Escape key. Similar to other tools.
- I make it require a minimal size of 1 mm x 1 mm. This should be enough IMHO and avoids extreme zoom.

I hope this still works as desired.

dg0yt:
... and I change the label to "Zoom to box". AFAICS this is much more common than "Box zoom".

src/tools/box_zoom_tool.h
```diff
@@ -0,0 +1,67 @@
...
+class QRectF;
+
+namespace OpenOrienteering {
+
+class MapEditorController;
+class MapWidget;
+
+
+/**
+ * Tool to pan the map.
```
dg0yt, line 41:
```suggestion
 * Tool to zoom to a user defined region.
```

src/tools/box_zoom_tool.h
```diff
@@ -0,0 +1,67 @@
...
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with OpenOrienteering.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#ifndef OPENORIENTEERING_TOOL_BOX_ZOOM_H
+#define OPENORIENTEERING_TOOL_BOX_ZOOM_H
```
dl3sdo, line 22:
```suggestion
#ifndef OPENORIENTEERING_BOX_ZOOM_TOOL_H
#define OPENORIENTEERING_BOX_ZOOM_TOOL_H
```

dg0yt:
@dl3sdo @elektronisk Please test and let me know if it can be merged.

dl3sdo:
There are some minor formal issues like `tools/box_zoom_tool.cpp` being not at the right position in `CMakeLists.txt` or `#include "tools/box_zoom_tool.h"` being not at the right position in `map_editor.cpp`. Github seems not to offer convenient review comments on moving code lines.
The tool works as intended, also in combination with panning or zoom the map.

dl3sdo (approved):

dg0yt:
Thank you very much!

Issue #2133, opened 2023-01-26
Point object angle/rotation is lost during DXF import


MaBreaker:
### Steps to reproduce
1. Generate DXF file with rotated POINT objects (slope lines, springs, cave, bouldery field etc.) 
2. Import DXF into new map
3. Point objects are not rotated as should, but pointing north

### Actual behaviour
Point object rotation angle is defined in DXF file (like below POINT code 50 angle -167.00), but are lost in Import

```
POINT
8
101.1_slope_line
10
12345.00
20
67890.50
50
-167.00
0
```

### Expected behaviour
Point object rotation angle should be set accordingly (when defined) during file import.

### Proposal
I believe DXF file is imported through OGR and POINTs with importPointGeometry function. If so and if original DXF angle is stored into OGR symbol details (which I do not know) the same way as with Text objects, then it could work something like..

```
Object* OgrFileImport::importPointGeometry(OGRFeatureH feature, OGRGeometryH geometry)
{
  ...
  if (symbol->getType() == Symbol::Point)
  {
    auto object = new PointObject(symbol);
    object->setPosition(toMapCoord(OGR_G_GetX(geometry, 0), OGR_G_GetY(geometry, 0)));

    // get rotation angle from symbol details
    const auto& description = symbol->getDescription();
    auto split = description.indexOf(QLatin1Char(' '));
    auto angle = QStringRef(&description, 3, split-3).toDouble(&ok);
    if (ok)
    {
      object->setRotation(qDegreesToRadians(angle));
    }

    return object;
  }
  ...
```

### Configuration
Mapper Version: 0.9.5
Operating System: All

dl3sdo:
@MaBreaker: could you attach a small DXF test file to this issue?

dl3sdo:
@MaBreaker: your proposed solution does not work that way.
For text objects the angle (and anchor) is 'transported' within the 'description' property of the text symbol (see getSymbolForLabel()).
For point objects no angle is retrieved and no description is set.

I doubt that this way of transporting object specific data should be applied to other objects types.

MaBreaker:
> @MaBreaker: your proposed solution does not work that way. For text objects the angle (and anchor) is 'transported' within the 'description' property of the text symbol (see getSymbolForLabel()). For point objects no angle is retrieved and no description is set.
> 
> I doubt that this way of transporting object specific data should be applied to other objects types.

[...]

PR #2362, opened 2025-04-11
OgrFileImport: Import rotation angle of point objects (DXF)

dl3sdo:
The DXF driver of GDAL/OGR does not (yet) import the rotation angle of point objects (code 50).
However, the driver stores all codes which are not handled by the driver itself in a generic field 'RawCodeValues' if the driver is configured with DXF_INCLUDE_RAW_CODE_VALUES being set to TRUE.
Parse the 'RawCodeValues' field for the occurrence of a code 50 property and set the rotation angle accordingly.

Closes GH-2133 (Point object angle/rotation is lost during DXF import).

### DIFF
--- a/src/gdal/ogr_file_format.cpp
+++ b/src/gdal/ogr_file_format.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2016-2020 Kai Pastor
+ *    Copyright 2016-2020, 2025 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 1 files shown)


dl3sdo:
I propose to merge #2136 first and then evaluate this PR.
Todos:

1. Add a test
2. Consider adding DXF_INCLUDE_RAW_CODE_VALUES = TRUE to the GDAL/OGR section of Mapper's Settings dialog:
ExtendedSettings

PR #2364, opened 2025-04-15
TemplateAdjustWidget: Disable tool

dl3sdo:
Disable the corresponding tool if no pass points are available to be moved or deleted and in case all pass points are removed.


dl3sdo:
Behaviour before fix:

https://github.com/user-attachments/assets/d51ec3a1-942c-4384-91ed-8d9355fad3a1

dg0yt (approved):
LGTM, but check my changes.

dl3sdo:
Thank you for the update Kai. Disabling new_act was an option and I'm fine with keeping the original behaviour.
I tested it and everything looks good to me.

Issue #2368, opened 2025-04-21
Rotated text cannot be selected


jmacura:
### Steps to reproduce:

1. Write some text with a symbol, which can be rotated (not fixed to north).
2. Rotate the text
3. Try to select the text by clicking on it

It can only be selected when clicked right at its definition point. Clicking anywhere on the text does not work like in the case of not-rotated text.
Additionally, it seems that the text can be selected when clicking on an axis perpendicular to its baseline (invisible). See the attached screen record.

https://github.com/user-attachments/assets/025d09f5-abec-4277-bc0d-e9462513aef3

### Tested Mapper versions:

- 0.8.4 ... OK
- 0.9.5 ... affected
- master v20250401.7 ... affected

dg0yt:
I also noticed this some days ago. At least it is not selectable in the expected region.

dl3sdo:
@jmacura: the text is selectable in the "opposite" y-direction:

https://github.com/user-attachments/assets/e26bae8d-57d7-4ceb-931f-f69f09516975

@dg0yt: It works when inverting the rotation, i.e.,
changing
`transform.rotate(-qRadiansToDegrees(getRotation()));`
to
`transform.rotate(qRadiansToDegrees(getRotation()));`
in 
`QTransform TextObject::calcMapToTextTransform() const`

Since `calcMapToTextTransform()` is also called by `OcdFileExport::exportTextCoordinatesSingle()` this change needs to be examined in more detail.

dg0yt:
Probably broken by https://github.com/OpenOrienteering/mapper/commit/53900087bcddf6d291c696a582556d2d77396b65.

PR #2369, opened 2025-04-23
TextObject: Apply inverse rotation

dl3sdo:
If the user clicks inside a rotated text object, rotate the position of the mouse click inverse to the rotation of the text in order to compare the position against the (unrotated) text object dimensions.

Closes GH-2368 (Rotated text cannot be selected).

### DIFF
--- a/code-check-wrapper.sh
+++ b/code-check-wrapper.sh
@@ -114,6 +114,7 @@ for I in \
   template_t.cpp \
   template_tool \
   template_track.cpp \
+  text_object.cpp \
   text_object_editor_helper.cpp \
   text_browser_dialog \
   toast.cpp \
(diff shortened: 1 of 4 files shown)


dg0yt (commented):
We don't need the extra parameter.  

There is only one proper implementation which does the reverse of the other function.

There are only two calls to this function, the one which was modified to use the new behavior, and another one in `OcdFileExport` which wasn't modified. This second case is only transforming the anchor point which isn't affected by rotation, and always mapped to 0,0. That call could be eliminated... Learned while writing the test case ;-)

I would appreciate some validation of the OCD side nevertheless.

dl3sdo:
That's interesting, but indeed: `auto anchor_text = map_to_text.map(anchor);` results in (0.0 ; 0.0)

Opening the exported .ocd in OCAD looks fine!

PR #2371, opened 2025-04-27
MapFindFeature: Exclude hidden and protected objects

dl3sdo:
The 'Find next' search failed if the object query returned an object with either a hidden or protected symbol since these kind of objects can't be added to an object selection.
Exclude all hidden and protected objects from the 'Find next' and 'Find all' search.

### DIFF
--- a/src/gui/map/map_find_feature.cpp
+++ b/src/gui/map/map_find_feature.cpp
@@ -20,8 +20,6 @@
 
 #include "map_find_feature.h"
 
-#include <functional>
-
 #include <QAction>
 #include <QAbstractButton>
 #include <QDialog>
@@ -45,8 +43,6 @@
 
 
 namespace OpenOrienteering {
-
-class Object;
 
 namespace {
 
@@ -217,8 +213,6 @@ void MapFindFeature::findNext()
 		}
 	};
 	map->getCurrentPart()->applyOnMatchingObjects(search, query);
-	
-	map->clearObjectSelection(false);
 	if (!next_match)
 		next_match = first_match;
 	if (next_match)
--- a/src/gui/map/map_find_feature.cpp
+++ b/src/gui/map/map_find_feature.cpp
@@ -22,8 +22,8 @@
 
 #include <functional>
 
-#include <QAction>
 #include <QAbstractButton>
+#include <QAction>
 #include <QDialog>
 #include <QDialogButtonBox>
 #include <QGridLayout>
@@ -129,7 +129,7 @@ void MapFindFeature::showDialog()
 		
 		auto button_box = new QDialogButtonBox(QDialogButtonBox::Close | QDialogButtonBox::Help);
 		connect(button_box, &QDialogButtonBox::rejected, &*find_dialog, &QDialog::hide);
-		connect(button_box->button(QDialogButtonBox::Help), &QPushButton::clicked, this, &MapFindFeature::showHelp);
+		connect(button_box, &QDialogButtonBox::helpRequested, this, &MapFindFeature::showHelp);
 		
 		editor_stack = new QStackedLayout();
 		editor_stack->addWidget(text_edit);
@@ -182,6 +182,7 @@ ObjectQuery MapFindFeature::makeQuery() const
 }
 
 
+// slot
 void MapFindFeature::findNext()
 {
 	auto query = makeQuery();
@@ -227,6 +228,7 @@ void MapFindFeature::findNext()
 }
 
 
+// slot
 void MapFindFeature::findAll()
 {
 	auto map = controller.getMap();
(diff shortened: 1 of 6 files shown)


dl3sdo:
For the 'Find all' mechanism the hidden and protected objects were just ignored when adding to the object selection, whereas for the 'Find next' mechanism the outcome depended on the internal order of the objects:

https://github.com/user-attachments/assets/213ec5d5-ff1b-436d-885d-fd67f8d5f39e

dg0yt:
Yeah, I got the problem. I'm a little bit slow with reviewing the changes :see_no_evil:

src/gui/map/map_find_feature.cpp
```diff
@@ -231,10 +235,13 @@ void MapFindFeature::findAll()
...
 	}
-	
+	auto search = [&query](const Object* object) {
+		const auto* object_symbol = object->getSymbol();
+		return !object_symbol->isProtected() && !object_symbol->isHidden() && query(object);
+	};
 	map->getCurrentPart()->applyOnMatchingObjects([map](Object* object) {
 		map->addObjectToSelection(object, false);
-	}, std::cref(query));
+	}, search);
```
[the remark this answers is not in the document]
dl3sdo: It's a runtime optimization: don't perform the query for objects which would not be added to the object selection anyway.

dg0yt:
Made `findNext` and `findAll` more similar, both using `applyOnMatchingObjects(..., query)` now. Avoid the second walk for `findNext`.

It is a little bit irritating that objects with *protected* symbols are skipped. But this can't be changed easily now. The PR is an improvements nevertheless.

src/gui/map/map_find_feature.cpp
```diff
@@ -172,46 +182,41 @@ ObjectQuery MapFindFeature::makeQuery() const
...
-			if (first_object)
-			{
-				if (object == first_object)
-					first_object = nullptr;
-			}
-			else if (query(object))
-			{
-				next_object = object;
-			}
+			if (after_pivot && !next_match)
```
dg0yt, line 209:
```suggestion
			if (after_pivot)
```
Already shortcutting on `next_match` in line 200.

dl3sdo:
@dg0yt: adding the `std::cref` around `query` again fixes the compiler issue for me (it's compiling however). CI stops for some targets.

src/gui/map/map_find_feature.cpp
```diff
@@ -198,23 +198,23 @@ void MapFindFeature::findNext()
...
+		if (isSelectable(object) && query(object))
 		{
 			if (after_pivot)
 				next_match = object;
 			else if (!first_match)
 				first_match = object;
 		}
 	};
-	map->getCurrentPart()->applyOnMatchingObjects(search, std::cref(query));
+	map->getCurrentPart()->applyOnAllObjects(search);
```
dg0yt, line 217: Reverting part of my change here: The current object might no longer match the query, but we still want to see it when determining the next match.

dg0yt:
I will probably revert most of [b0dcb5a](https://github.com/OpenOrienteering/mapper/pull/2371/commits/b0dcb5a2e30df769c919816d019376a708e341c4). `const` and `slot` were omitted intentionally.

dl3sdo:
I decided to not create another separate file but to add the test case to `ObjectQueryTest`.
It took me some time to find out that I could not make `ObjectQueryTest` a friend of `MapFindFeature` since `ObjectQueryTest` is in a different namespace. And changing the namespace will cause trouble with `Q_INIT_RESOURCE(resources); doStaticInitializations();`
I am surprised that the search operates in reverse order (this was already the situation before this PR).

dl3sdo:
Thank you Kai for refactoring my test code! BTW: tools_t.cpp was the file I also had in mind to put the test in and from which I tried to take over some code.

dg0yt:
The last commit is the result of being forced into refactoring...

We can make the testable functions public because they only need the controller and query, not the object state. (They might be member functions of controller, but you know ... I don't want to add more capabilities to that class.)

dl3sdo:
@dg0yt: I checked the code once more and everything looks good to me. I really like the clear structure and especially the `search` functor which does now everything in just one function! Thanks again.

dg0yt:
A few more twists.
It is really good now ;-)

dl3sdo:
I tested the final version and reviewed the code once again, everything now really looks good and works as intended. Thank you Kai for spending the effort.
Rebasing #2374 upon that PR will be really fun ;-)
Not related to that PR: I wonder if anyone really uses the 'Find next' entry in the 'Edit' menu instead of doing that search in the 'Find...' dialog.

dg0yt:
> Not related to this PR: I wonder if anyone really uses the 'Find next' entry in the 'Edit' menu instead of doing the search in the 'Find...' dialog.

This entry is for searching with the dialog hiding parts of the map. There used to be also shortcuts IIRC.

dg0yt:
Thanks Matthias.

PR #2372, opened 2025-05-01
Highlight destructive options in the Save file dialog

lpechacek:
This change highlights the "discard changes" option in the file save dialog. Although it's a tiny change, anytime I see the red highlight on the button, I enjoy its usefulness. I'm sharing the change with OpenOrienteering in the hope that the joy will spread.

Screenshot From 2025-05-01 10-29-34

### DIFF
--- a/src/gui/main_window.cpp
+++ b/src/gui/main_window.cpp
@@ -706,10 +706,16 @@ bool MainWindow::showSaveOnCloseDialog()
 		}
 		else
 		{
-			ret = QMessageBox::warning(this, appName(),
-			                           tr("The file has been modified.\n"
-			                              "Do you want to save your changes?"),
-			                           QMessageBox::Save | QMessageBox::Discard | QMessageBox::Cancel);
+			QMessageBox msgBox(QMessageBox::Warning, appName(),
+			                   tr("The file has been modified.\nDo you want to save your changes?"),
+			                   QMessageBox::Save | QMessageBox::Discard | QMessageBox::Cancel, this);
+			// Restyle destructive buttons
+			for (auto const& btn : msgBox.buttons())
+				if (msgBox.buttonRole(btn) == QMessageBox::ButtonRole::DestructiveRole)
+					btn->setStyleSheet(QLatin1Literal("QPushButton { color: red; }"));
+			msgBox.setDefaultButton(QMessageBox::Cancel);
+			auto exec = msgBox.exec();
+			ret = (exec == -1) ? QMessageBox::Cancel : msgBox.standardButton(msgBox.clickedButton());
 		}
 		
 		switch (ret)


dg0yt:
Increasing awareness at this point is fine. I'm not happy with the proposed form:
- From past experience, the stylesheet approach to customizing appearance has some issues (with platform specific styles IIRC).
- Red font color isn't a sufficient indicator from an accessibility POV.
- Both "Close without saving" and "Save" are potentially destructive: "Save" may accidentally overwrite the desired state. It might destroy much more work, given frequent manual saving and auto-save.

I wonder if adding icons to the buttons could achieve a similar effect of catching attention, and avoid stylesheets at the same time. 

But in general we try to follow the platform style. The dialog itself, with its icon, is already a fat STOP sign.

lpechacek:
Thanks for the comment, Kai. Yes, this is where we often disagree. I'm aware of the imperfections in the code. In my view the benefits outweigh the imperfections.

dg0yt:
The comments weren't only about code. And you could have waited for others to give feedback. 

How am I expected to give anything but positive feedback when the PR is closed on my first comment :disappointed:

lpechacek:
I really don't intend to mentor a seasoned FOSS maintainer but it helps to start the comment with the word "thanks".

Next, all your points are valid and I anticipated them. I know what are the deficiencies in this code change. So, if we don't agree on the usability value of the change (in my opinion the value is "substantial" while the first sentence in your comment suggests "marginal"), I consider further discussion moot. I don't have energy for a "who is more right" discussion and those who brought me to the idea for this change won't support me here as they are out in the forests making maps.

Let's not spoil our holiday. Wish you a good Labor Day!

dg0yt:
A copy of the the submission is stored at https://github.com/dg0yt/mapper/tree/lpechacek-highlight-destructive-options now.
Thank you.

dg0yt:
> I wonder if adding icons to the buttons could achieve a similar effect of catching attention

My current desktop has icons:
image

lpechacek:
For the record.

- The idea started with a comment from one of my fellow mappers that he occasionally hits the close button and the modified file dialog pops up. As this is unintended action, it's somewhat difficult to decide which button undoes the unwelcome effect. Two of three buttons mean a safe choice but one causes loss of work. Which one is it?
- I tend to believe that the difficulty of choice is caused by the fact that mappers in the field work with non-verbal information - colors, lines, shapes. After hours of work the brain gets into this non-verbal flow state and switching it back to analyzing words requires some effort. That led me to the idea of making the dangerous option visually distinct.
- I like the LibreOffice implementation the most - the button face is all red. It cannot be mistaken. But I was unable to copy the LibreOffice trick, although the code is publicly available, so red letters was the next best choice.
- Although I don't have the problem of triggering the close button during fieldwork, I often discard file modifications during testing. I very much like the simplicity of choosing between the destructive option the the others just by the color.

I agree that the icons may help. Still, I like the colored buttons more.

Illustration - LibreOffice Writer asking whether or not to save the document changes:
image

jmacura:
> 
>     * The idea started with a comment from one of my fellow mappers that he occasionally hits the close button and the modified file dialog pops up. As this is unintended action, it's somewhat difficult to decide which button undoes the unwelcome effect. Two of three buttons mean a safe choice but one causes loss of work. Which one is it?

I can say this happens to me from time to time as well. In that context, highlighting the "dangerous" option makes sense. On the other hand.... UI-wise... the highlighted button must be the most important one, right? It is the right thing to click the highlighted button, all right? (-:

dg0yt:
AFAICS the initial post lacks an import detail:  
The context is the touch mode, in particular Andoid app.

It is important because it raises extra questions:
- Did the proposed change have an effect on Android? It is a native dialog on Android, i.e not using the same code as shown in previous screenshots.
- Would it make sense to add better prevention of accidental touch of the close button? (e.g. menu, double touch, swipe to confirm.)

Another point which I discussed offline with @dl3sdo:
- Would it make sense to offer a kind of Undo option for the destructive action. (For example, my e-mail app lets me recover e-mails for some seconds after swiping them to the trash or archive.)

Mobile screenshot:
tempFileForShare_20250523-103825

lpechacek:
> AFAICS the initial post lacks an import detail: The context is the touch mode, in particular Andoid app.

Ummm, speaking for myself, it helps me the most in the desktop mode - during testing. But yes, those who initiated the idea would use this feature primarily on Android.

>     * Did the proposed change have an effect on Android? It is a native dialog on Android, i.e not using the same code as shown in previous screenshots.

Not tested.

>     * Would it make sense to add better prevention of accidental touch of the close button? (e.g. menu, double touch, swipe to confirm.)

Yes. When we are talking about the Android UI, hiding the close button under a menu makes a lot of sense. An average mapper uses the close button once a day if at all.

lpechacek:
> And that's exactly when I really want "Close without saving" most of the time, turning "Save" into the destructive option.

:) Same here. Please mind, that the change (making the destructive options stand out) is indented to **support the decision process**, not to deter the user from clicking it. So in testing I choose red. When I'm making a production map, I avoid red. That's all. I don't have to study the pixels, translate words and whatever else. That's what I like about the enhancement.

> (I tried to avoid presenting this special development experience because it is not the most relevant use case.)

Agreed. As I stated, my aim is to support the decision process rather than route users through a certain path.

> But yeah, Mapper is stable enough to not need any intermediate Ctrl-S ;-)

:+1:

PR #2378, opened 2025-05-09
AreaSymbol: Revise fill pattern rotation handling

dg0yt:
Let `AreaSymbol` adopt the 'rotatable' property of `Symbol`.
    
This property already exist for Symbol, but wasn't used in AreaSymbol:
area symbol fill patterns carry individual rotatability properties.
The new symbol property does not interfere with the pattern property:
Its role is to provide a hint to map rotation operations, as
implemented in the OC*D. (However, this effect is not yet implemented
here.)

Existing maps are not affected because the existing rendering and
editing behavior doesn't need to care for this property. Old versions
of Mapper would drop the "rotatable" property, but this doesn't have an
immediate effect until exporting to OCD or (in the future) rotating the
map. Given this low impact, this PR does not add a omap format version.

Known OCD format limitations:
- "Mixed" pattern rotatability configurations are not exported correctly to OCD.  
  A possible workaround would be duplicating the symbol for individidual configurations. However this could yield up to 3600 OCD symbols in the worst case.  
  Not a regression.
- Pattern origin cannot be exported to OCD.  
  Not a regression.

Resolves GH-1926.
Resolves GH-1677.


symbol sets/src/ISMTBOM_15000.xmap
```diff
@@ -6728,7 +6728,7 @@ Note: if the stream is wider than 0.25mm, adjust this symbol so it extends 0.5mm
             </symbol>
             <symbol type="4" id="166" code="601.1" name="North lines pattern" is_hidden="true">
                 <area_symbol inner_color="-1" min_area="0" patterns="1">
-                    <pattern type="1" angle="1.5708" rotatable="true" line_spacing="33330" line_offset="0" offset_along_line="0" color="1" line_width="140"/>
+                    <pattern type="1" angle="1.5708" line_spacing="33330" line_offset="0" offset_along_line="0" color="1" line_width="140"/>
```
dg0yt, line 6731: The ISMTBOM north lines pattern was rotatable. But they shall be parallel to "the sides of the map".

dg0yt:
Planned to "rebase and merge".

PR #2381, opened 2025-05-11
PointSymbolEditorWidget: Hide oriented-to-north for area symbol patterns

dg0yt:
Complements #2378. 

The point symbol editor widget is used for pattern fills in area symbols. It featured the "Always oriented to north (not rotatable)" checkbox, but this checkbox didn't have an effect in that use case. This PR hides the checkbox. This should reduce confusion, in particular with the same checkbox added to the area symbol itself in #2378.

NB: The checkbox has an effect for line symbol decorators. I wouldn't say it is intented, but I don't want to change this now.

### DIFF
--- a/src/gui/symbols/area_symbol_settings.cpp
+++ b/src/gui/symbols/area_symbol_settings.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2010, 2025 Kai Pastor
+ *    Copyright 2012-2020, 2025 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
--- a/src/gui/symbols/area_symbol_settings.cpp
+++ b/src/gui/symbols/area_symbol_settings.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2017 Kai Pastor
+ *    Copyright 2012-2010, 2025 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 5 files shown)


src/gui/symbols/point_symbol_editor_widget.h
```diff
@@ -62,13 +62,21 @@ class PointSymbolEditorWidget : public QWidget
...
 	/** Construct a new widget.
 	 * @param controller The controller of the preview map
 	 * @param symbol The point symbol to be edited
+	 * @param role The role of the symbol to be edited
 	 * @param offset_y The vertical offset of the point symbol preview/editor from the origin
-	 * @param permanent_preview A flag indicating whether the preview shall be visible even if the editor is not visible
+	 * @param parent Standard QWidget parentship
 	 */
-	PointSymbolEditorWidget(MapEditorController* controller, PointSymbol* symbol, qreal offset_y = 0, bool permanent_preview = false, QWidget* parent = 0);
+	PointSymbolEditorWidget(MapEditorController* controller, PointSymbol* symbol, SymbolRole role = PrimarySymbol, qreal offset_y = 0, QWidget* parent = nullptr);
```
dg0yt, line 79: This change doesn't only add more expressive values than false/true, but also changes the default value to match the PrimarySymbol use case.

src/gui/symbols/area_symbol_settings.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2017 Kai Pastor
+ *    Copyright 2012-2010, 2025 Kai Pastor
```
dl3sdo, line 3: *    Copyright 2012-2020, 2025 Kai Pastor

dl3sdo (commented):
Changes look good to me. Functional behaviour is as expected due to this change. Thank you Kai.

src/gui/symbols/area_symbol_settings.cpp
dg0yt:
```suggestion
 *    Copyright 2012-2020, 2025 Kai Pastor
```

PR #2389, opened 2025-05-29
Fix singular/plural forms for English

dl3sdo:
Since the current Weblate setup fails for singular/plural forms in English,
fix the translations in the .ts file directly.


dg0yt (approved):
LGTM, tnx.

PR #2402, opened 2025-06-23
MapEditorController: Paste at original location

dl3sdo:
The current paste tool inserts objects at the map's viewport center.
Add a new tool that pastes objects at their original (map coordinates) location.


lpechacek:
> If I want to backport those changes to the master map I can copy the object from the map extract to the master map but have to move it to the correct location.

FWIW, I'm using the import function for this purpose. I delete all objects in the modified map except those I want to transfer and save it as a new file. Then I open the master file and import the modified objects. The new "paste at the original location" function might be a welcome shortcut. Thanks!

nalajcie:
+1 for this feature. 

The difference from using "import" is that we're not doing any CRT symbols replacement (we will eventually add symbols/colors if needed - as with normal ^C^V between files).

PR #2409, opened 2025-07-04
TemplateListWidget: Fix cancelling map import

dl3sdo:
The import of a map that was loaded as a template before may invoke a dialog to select scaling options.
The Cancel button however did not cancel the import as expected (since `option.isEmpty()` is not possible to become true as editing is disabled (but it's ok to keep it)).


dl3sdo:
Example: Import of `forest sample.omap` which was previously scaled to 1:15000 into the original map (scale 1:10000):

https://github.com/user-attachments/assets/aee53480-6d08-4a64-be44-dcf42d70f671

PR #2421, opened 2025-10-03
OgrFileExport: Exclude objects with hidden or helper symbols

dl3sdo:
The geospatial vector export shall not export objects with hidden or helper symbols.
However, if the "Create a layer for each symbol" setting in the GDAL/OGR configuration was not set then objects with hidden or helper symbols would not be excluded.

### DIFF
--- a/src/gdal/ogr_file_format.cpp
+++ b/src/gdal/ogr_file_format.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2016-2020 Kai Pastor
+ *    Copyright 2016-2020, 2025 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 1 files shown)


dl3sdo:
ChatGPT (and me) propose a better commit message:
Previously, objects with hidden or helper symbols were exported if the “Create a layer for each symbol” option in the GDAL/OGR configuration was disabled.
This change ensures that such objects are always excluded, independent of the configuration.

dg0yt:
I prefer concise commit messages, and so far I prefer to not be bothered with reviewing LLM output.

dg0yt:
Hm, noticing a mismatch between function name and function result :thinking: 

The original design was:  
These functions really test only symbol type. (`ìs_point_symbol`).  
Helper symbols and hidden symbols are excluded by creating a curated list of symbols first (`auto symbols = symbolsForExport();`). This list is also useful for populating the style table.

But the "per-symbol layers" export doesn't use the curated list when populating the layers.

dg0yt:
Retriggering CI.

src/gdal/ogr_file_format.cpp
```diff
@@ -1972,137 +1973,103 @@ bool OgrFileExport::exportImplementation()
...
+		LayerPerType
+	} export_mode = ExportMode::LayerPerType;
 
 	if (quirks & SingleLayer)
 	{
-		auto layer = createLayer("Layer", wkbUnknown);
+		export_mode = ExportMode::SingleLayer;
+		layer = createLayer("Layer", wkbUnknown);
 		if (layer == nullptr)
 			throw FileFormatException(tr("Failed to create layer: %2").arg(QString::fromLatin1(CPLGetLastErrorMsg())));
```
dl3sdo: You reverted `"Failed to create layer: %1"` back to `"Failed to create layer: %2"`.
I don't understand the reason behind it (at least since this file contains multiple similar fomats, e.g.
`throw FileFormatException(tr("Failed to create feature in layer: %1").arg(QString::fromLatin1(CPLGetLastErrorMsg())));`

dl3sdo: Thank you for the clarification. That behaviour of QT was unknown to me. My bad, I should have read the online help first.

PR #2426, opened 2025-10-14
PointSymbolEditorWidget: Create point symbol from selected objects

### LINKED ISSUE #2425: Create point symbols from selected objects
Point symbols can themselves consist of other point, line and area subsymbols.
The point symbol editor provides limited support to create point, line and area subsymbols.

In principle, line and area subsymbols can have the same properties as standard line and area symbols. For example, an area subsymbol may have line and pattern fills.
However, the point symbol editor only allows to configure the color of a solid fill and does not support more advanced settings.

Mapper also allows to draw and modify shapes of any color in a much more convenient way than the point symbol editor (see also the feature request in #2383).

Proposal: Extend the point symbol editor to create point, line and area subsymbols from selected objects and add them to an existing point symbol.

Example: Create a North Arrow symbol from area objects

https://github.com/user-attachments/assets/da595b1b-68a7-4b5b-a2cf-34bad5004595

Another example: Merge one point symbol into another by using a point object

https://github.com/user-attachments/assets/5288eb4a-7412-445c-bae5-8090117b8a11

dl3sdo:
Extend point symbol editor to create point, line and area subsymbols from selected objects.

Closes GH-2425 (Create point symbols from selected objects).


src/gui/symbols/point_symbol_editor_widget.cpp
```diff
@@ -78,12 +80,13 @@
 
 namespace OpenOrienteering {
 
-PointSymbolEditorWidget::PointSymbolEditorWidget(MapEditorController* controller, PointSymbol* symbol, SymbolRole role, qreal offset_y, QWidget* parent)
+PointSymbolEditorWidget::PointSymbolEditorWidget(MapEditorController* controller, PointSymbol* symbol, Map* source_map, SymbolRole role, qreal offset_y, QWidget* parent)
```
dg0yt: Early feedback: 
I don't think we should add another parameter. The source map is implicitly known, via controller and symbol.
(Another suggested workflow was copy-and-paste. This would use a different Map, with all complications. And you wouldn't get that Map at constuction time, before "Paste".)

dl3sdo: I also didn't like to add a new parameter but had not found a better alternative.
The newly created controller being passed to PointSymbolEditorWidget only knows the preview map but does not know the source_map, neither does does the symbol. There is certainly the previously active controller who 'knows' the source map but I don't know how PointSymbolEditorWidget can access it.
MapEditorController describes itself as a 'HACK' for being used in the symbol editor for the preview. MapEditorController could be extended to store the source map in addition but that only extends the hack.

dl3sdo:
When adding an area object that consists of a line pattern, things get tricky.
Currently, point symbols don't pass the rotation to their children, which is documented in `point_symbol.cpp`:
`// TODO: if this point is rotated, it has to pass it on to its children to make it work that rotatable point objects can be children. But currently only basic, rotationally symmetric points can be children, so it does not matter for now.`

Adding that requires changing pure virtual `createRenderables()` and all it's overridden instances.
When doing that the result still is not a point symbol with a 'frozen' layout, since rendering patterns will be done in the same way as rendering them in area objects, i.e. the position of the pattern lines depends on their position on the map, not on their relative position to the area object itself:

https://github.com/user-attachments/assets/be371fb8-7b61-4f37-af90-8d2802576847

The same behaviour would apply to point patterns.

**It might therefore be sufficient to support only solid patterns.**

Issue #2427, opened 2025-10-17
Removing last but one map part causes out-of-bounds access


dl3sdo:
This issue was brought up by @lpechacek in [Remove map part triggers assertion fail in std::vector](https://github.com/LibreMapper/mapper/issues/7).

If the last but one map part is removed (either explicitly or by merging this map part), then 
`Map::setCurrentPartIndex()` will do an out-of-bounds read access to the parts vector.

Example:
There are three map parts A, B and C. If B shall be removed, then `Map::removePart` is called with index == 1 while current_part_index is also 1.
```
	if (current_part_index == index)
		// First switch to another part when removing the current part
		setCurrentPartIndex((index == parts.size() - 1) ? (parts.size() - 2) : (index + 1));
```
will switch to part C (current_part_index ==2).
Part B will be removed, so that the parts vector now has only two elements, index 0 and 1.
```
	if (current_part_index >= index)
		setCurrentPartIndex((index == parts.size()) ? (parts.size() - 1) : index);
```
is now calling `setCurrentPartIndex()` again (current_part_index == 2 while index == 1).

Inside `setCurrentPartIndex()` 
`MapPart* const old_part = parts[current_part_index];`
will try to read the element 2 from the parts vector which now only contains elements 0 and 1.

Depending on the compiler version (or compiler settings?) this results in an error in debug mode.

PR #2428, opened 2025-10-17
Map: Removing map parts can cause out-of-bounds access

dl3sdo:
If the last but one map part was removed, setCurrentPartIndex() did an out-of-bounds read access to the parts vector, due to switching to the last map part before removing the last but one part.

Closes GH-2427 (Removing last but one map part causes out-of-bounds access).

### DIFF
--- a/src/core/map.cpp
+++ b/src/core/map.cpp
@@ -2061,8 +2061,8 @@ void Map::removePart(std::size_t index)
 	Q_ASSERT(parts.size() > 1);
 	
 	if (current_part_index == index)
-		// First switch to another part when removing the current part
-		setCurrentPartIndex((index == parts.size() - 1) ? (parts.size() - 2) : (index + 1));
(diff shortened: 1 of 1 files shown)


src/core/map.cpp
```diff
@@ -2061,8 +2061,8 @@ void Map::removePart(std::size_t index)
 	Q_ASSERT(parts.size() > 1);
 	
 	if (current_part_index == index)
-		// First switch to another part when removing the current part
-		setCurrentPartIndex((index == parts.size() - 1) ? (parts.size() - 2) : (index + 1));
+		// First switch to another part when removing the current part, this will also clear any object selection
+		setCurrentPartIndex((index == parts.size() - 1) ? (index - 1) : (index + 1));
```
dl3sdo: Although `index - 1` is equivalent to `parts.size() - 2` I prefer it due to readability.

Issue #2435, opened 2025-11-04
Crash when interrupting Clip/Erase Area tool by graphical export


dl3sdo:
Steps to reproduce the crash in either debug or release mode:

1. Select any area object
2. Select either Clip or Erase Area tool
3. Open either the _Print_ dialog or one of the _Image/PDF/KMZ_ dialogs below _Export as..._
4. Close the dialog

=> Mapper crashes with message (in debug mode)
`ASSERT: "map()->getNumSelectedObjects() == 1" in file cutout_tool.cpp, line 79`

https://github.com/user-attachments/assets/044db3a2-5b23-4f43-a815-7ef72d81faea

PR #2436, opened 2025-11-04
MapEditorController: Block graphical export while editing

dl3sdo:
If the Clip/Erase Area tool was activated and then interrupted by a graphical export (e.g., printing, PDF export), Mapper crashed after closing the graphical export dialog.
Show a warning that the editing operation needs to be finished first and block the graphical export.

Closes GH-2435 (Crash when interrupting Clip/Erase Area tool by graphical export).


dg0yt:
Thanks.

PR #2442, opened 2025-11-17
Export images with transparent background

dl3sdo:
Add option to export .png and .tif(f) images with transparent background.

### DIFF
--- a/src/gui/print_widget.cpp
+++ b/src/gui/print_widget.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2021 Kai Pastor
+ *    Copyright 2012-2021, 2025 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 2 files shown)


src/gui/print_widget.cpp
```diff
@@ -301,6 +301,10 @@ PrintWidget::PrintWidget(Map* map, MainWindow* main_window, MapView* main_view,
 	layout->addRow(world_file_check);
 	world_file_check->hide();
 	
+	transparent_background_check = new QCheckBox(tr("Make background transparent"));
```
dg0yt: We cannot enforce transparent background, but we can enforce white background.

Looking for feedback: Should this be 

 - [x] Draw white background

For formats without transparency, drawing the white background would be done implicitly.
We could completely avoid the error message. 
(But we might see questions why the background is always white in JPEG format.)

dl3sdo: That change is indeed an option as making the background transparent is sometimes a unrealizable promise.
OTOH a user might wonder what happens if he deselects the checkmark: will the background be black (BTW: this happens if Mapper would allow transparency for .bmp) as he might not think about transparency?
So I personally have a preference for the current wording but I'm fine with the alternative.

dl3sdo:
Pushing the minimal required set of changes.

PR #2443, opened 2025-11-17
PrintWidget: Show warning if world file export fails

dl3sdo:


dg0yt:
> Easy testable by trying to overwrite an existing write protected world file.

Who keeps read-only world files next to writable image files :thinking:

dg0yt:
> I stumbled across the Todo in the code

Indeed, an inline comment.

PR #2449, opened 2025-12-02
CombinedSymbol: Fix nullptr dereferencing

dl3sdo:
Add check for empty subsymbols when iterating over combined symbols.


dl3sdo:
The `validate()` functions of symbols are only called from `symbol_set_t.cpp`.
The test actually crashes if one of the sample maps contains a combined symbol with an empty subsymbol (which they don't do).

dg0yt:
Closing/Reopening for CI.

PR #2460, opened 2026-01-05
Fix: Flatten OGR curve geometries (CurvePolygon) to polygons before import

jirivis:
Hello,

This PR fixes an issue where geometries of type `CurvePolygon` ("polygonZ" in the attached picture, sometimes used in gpkg data, e.g., Czech Cadastral maps) were ignored or imported as empty geometries by the OGR importer.
Not_Loaded

Changes:
- Added a check for `OGR_G_HasCurveGeometry`.
- If a curve geometry is detected, it is converted to a linear approximation using `OGR_G_GetLinearGeometry` before processing.

This ensures that these complex polygon types are correctly imported as standard polygons.

### DIFF
--- a/src/gdal/ogr_file_format.cpp
+++ b/src/gdal/ogr_file_format.cpp
@@ -1059,7 +1059,16 @@ void OgrFileImport::importLayer(MapPart* map_part, OGRLayerH layer)
 			++empty_geometries;
 			continue;
 		}
-		
+				
+		if (OGR_G_HasCurveGeometry(geometry, 1)) {
+			OGRGeometryH linear_geom = OGR_G_GetLinearGeometry(geometry, 0, nullptr);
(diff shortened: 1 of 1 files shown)


src/gdal/ogr_file_format.cpp
```diff
@@ -1059,7 +1059,16 @@ void OgrFileImport::importLayer(MapPart* map_part, OGRLayerH layer)
 			++empty_geometries;
 			continue;
 		}
-		
+				
+		if (OGR_G_HasCurveGeometry(geometry, 1)) {
+			OGRGeometryH linear_geom = OGR_G_GetLinearGeometry(geometry, 0, nullptr);
```
dg0yt: "The ownership of the returned geometry belongs to the caller."

PR #2465, opened 2026-01-21
Test location fix for non GPS tablets

jLagosCarrera:
Hello good night,

I'm testing out a fix to retrieve the location from my [Lenovo Tab P11 Gen 2](https://www.lenovo.com/gb/en/p/tablets/lenovo-tab-series/tab-p11-2nd-gen/len103l0012#tech_specs:~:text=WiFi%20models%3A%20WiFi%206E%20802.11%20a/b/g/n/ac/ax). The tablet doesn't have an internal GPS and it uses the Fused Location Provider instead of the GPS provider required by Open Orienteering Mapper.

The thing is, I have a Garmin Glo 2 external GPS that I'm running with the [Bluetooth GNSS app](https://play.google.com/store/apps/details?id=com.clearevo.bluetooth_gnss&hl=en) mocking the location and still I'm not able to run the app because that doesn't change the tablet doesn't have a GPS provider available.

I opened the PR so that the pipeline builds the APK for me as I'm currently sitting at a Windows computer and I felt like this was going to be the fastest way of testing it by reading [this documentation](https://github.com/OpenOrienteering/mapper/wiki/Continuous-Integration).

Regardless of the answer, big thanks for you work on this project and for spending the time reading my answer. I hope you can find this idea cool and provide support to even more people that want to get started with OOM like me 😃 

EDIT:
Long story short, it works. Now OOM doesn't say I do not have GPS and the location is perfect because I'm running the Garmin Glo 2.

Now my question is... Would it be possible to include this on the next release @dg0yt?

Of course, not in the current state. I have no idea about C++ development but I could try to make this configurable through a checkbox on the location settings saying something like

> "**Enable approximate location:** Uses WiFi, Bluetooth, or Mock Location apps to find your position on devices without integrated GPS"

so that the users only checks them knowing what is it exactly for. Still I don't know if only the current changes and this checkbox I'm mentioning are the only needed changes... 😞 

I truly believe this is a cool feature that will enable a lot of users to use OOM. Let me know what you think!

### DIFF
--- a/src/android/java/org/openorienteering/mapper/MapperActivity.java
+++ b/src/android/java/org/openorienteering/mapper/MapperActivity.java
@@ -138,7 +138,15 @@ public class MapperActivity extends org.qtproject.qt5.android.bindings.QtActivit
 	static void checkGPSEnabled()
 	{
 		LocationManager locationManager = (LocationManager) instance.getSystemService(LOCATION_SERVICE);
-		boolean enabled = locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER);
+		// Use isLocationEnabled() for API 28 or higher to check if location services are enabled at all,
+		// on lower API levels check if either GPS or Network provider is enabled to cover devices without GPS (mocked with external device).
+		boolean enabled;
+		if (Build.VERSION.SDK_INT >= 28) {
+			enabled = locationManager.isLocationEnabled();
+		} else {
+			enabled = locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER)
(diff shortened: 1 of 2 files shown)


jLagosCarrera:
Update: I've tried [GPS Connector](https://play.google.com/store/apps/details?id=de.pilablu.gpsconnector&hl=en), another app to connect my Garmin Glo 2 and this one mocks the GPS properly.

Still, do you think the checkbox to enable less precise positioning methods (or to give support to other apps that mock the GPS signal) is something cool? Otherwise I'll just close the PR.

PR #2498, opened 2026-04-17
PointSymbol: Fix check for symmetry

dl3sdo:
The test for symmetry in function isSymmetrical() was broken since the check for the position being at (0,0) was not negated.
Use std::all_of to improve readability.

### DIFF
--- a/src/core/symbols/point_symbol.cpp
+++ b/src/core/symbols/point_symbol.cpp
@@ -1,6 +1,6 @@
 /*
  *    Copyright 2012, 2013 Thomas Schöps
- *    Copyright 2012-2019 Kai Pastor
+ *    Copyright 2012-2019, 2026 Kai Pastor
  *
  *    This file is part of OpenOrienteering.
  *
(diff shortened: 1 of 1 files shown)


dl3sdo:
Basically, the minimum fix would be
```
return std::none_of(begin(elements), end(elements), [](auto& element) {
	return element.symbol->getType() != Symbol::Point
	       || !static_cast<const PointObject*>(element.object.get())->getCoord().isPositionEqualTo({0,0});
});
```
Instead of using `none_of` in combination with two negations that are ORed, I prefer `std::all_of` as this is closer to the description in the .h-file: "only point elements at (0,0) are allowed".

dl3sdo:
The effect of the missing negation would only show up when importing an .ocd file where an unsymmetrical point object was rotated (I'm not aware that point symbols like that are really used):
OCAD:
RotatedPointObjectOCAD

Mapper (without fix) after import:
RotatedPointObjectWithoutFix

PR #2500, opened 2026-05-06
EditPointTool: Middle-click to toggle dash points

Abbe98:
This makes it possible for users to use "middle-click" to toggle dash points without first going into dash point editing through Ctrl+Space.


lpechacek:
> going into dash point editing through Ctrl+Space.

FWIW, dash points can be toggled by holding Space and clicking on a point.

Abbe98:
> > going into dash point editing through Ctrl+Space.
> 
> FWIW, dash points can be toggled by holding Space and clicking on a point.

TIL, that's neat, this still free up a hand for mouse users.

PR #2502, opened 2026-05-17
doc: update building instructions for macos

joneugster:
- Add instruction for building on macOS
- Fix: minimum `cmake` version below 3.5 leads to error with `cmake` 4+
- Update stale link for `clipper` library

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -94,7 +94,7 @@ Now you may start the build process by running
 make
 ```
 
-## Compiling for macos (without OpenOrienteering superbuild)
+## Compiling for macOS (with brew instead of OpenOrienteering superbuild)
 
 `brew` installs `cmake` 4.x but the project requires an older version 3. A legacy
 version of `cmake` can be downloaded at https://cmake.org/download/#text=Legacy%20Release then located at
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,9 +6,9 @@ The general build process prerequisites are:
  - A supported platform: 
    - Linux. Ubuntu 18.04 is known to work.
      Linux is also used to cross-compile for Android.
-   - macOS: 10.13 is known to work.
+   - macOS: 10.13 is known to work, so is 26.3 (Tahoe).
    - Windows: MSYS2 (MinGW subsystem).
- - CMake >= 3.7.
+ - CMake >= 3.7, < 4.0.
    CMake is available from https://cmake.org/.
  - A supported C++ compiler toolchain. C++14 is mandatory.
 
@@ -17,7 +17,7 @@ components. Direct dependencies are:
  - Qt >=5.5
    https://www.qt.io/download-open-source/
  - Clipper library (aka libpolyclipping) >= 6.1.3a
-   http://www.angusj.com/delphi/clipper.php
+   http://www.angusj.com/delphi/
  - PROJ Cartographic Projections Library >= 4.9
    https://proj4.org/
  - GDAL Geospatial Data Abstraction Library >= 2
@@ -77,11 +77,11 @@ zlib1g-dev
 ```
 
 When not using Qt Creator, open a terminal, and create a build directory, e.g.
-as subdirectory build in the source directory, and change to that directory.
+as subdirectory build in the project's root directory, and change to that directory.
 From the build directory, configure and build like this:
 
 ```
-cmake PATH/TO/SOURCE_DIR
+cmake PATH/TO/MAPPER_ROOT_DIR
 ```
 
 When building on openSUSE, you may want to add -DMapper_BUILD_CLIPPER=1. This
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -77,11 +77,11 @@ zlib1g-dev
 ```
 
 When not using Qt Creator, open a terminal, and create a build directory, e.g.
-as subdirectory build in the source directory, and change to that directory.
+as subdirectory build in the project's root directory, and change to that directory.
 From the build directory, configure and build like this:
 
 ```
-cmake PATH/TO/SOURCE_DIR
+cmake PATH/TO/MAPPER_ROOT_DIR
```
joneugster: I understood this as 

```
mkdir ./src/build
cd ./src/build
cmake ..
```

but it looks like what's meant is

```
mkdir ./build
cd ./build
cmake ..
```

isn't it?

dg0yt: I don't see a significant difference between "source directory" and "project's root directory". 
- The first term is actually found in CMake (`CMAKE_SOURCE_DIR`, `CMAKE_CURRENT_SOURCE_DIR`. I admit there is also `PROJECT_SOURCE_DIR` and `<ProjectName>_SOURCE_DIR`.)
- OTOH "root" can refer to installed trees in CMake.

INSTALL.md
```diff
@@ -94,6 +94,39 @@ Now you may start the build process by running
 make
 ```
 
+## Compiling for macos (without OpenOrienteering superbuild)
```
dg0yt:
```suggestion
## Compiling for macOS (with brew instead of OpenOrienteering superbuild)
```
FTR I'm playing with vcpkg instead of OpenOrienteering superbuild, and in works for macOS, too.

INSTALL.md
```diff
@@ -94,6 +94,39 @@ Now you may start the build process by running
...
+
+`brew` installs `cmake` 4.x but the project requires an older version 3. A legacy
```
dg0yt: Is there an error with CMake 4.x?

INSTALL.md
```diff
@@ -94,6 +94,39 @@ Now you may start the build process by running
...
+For the build it might be necessary to pass `-DCMAKE_PREFIX_PATH="$(brew --prefix qt@5)` to `cmake`:
+
+```
+mkdir build
+cd build
+cmake3 .. -DCMAKE_PREFIX_PATH="$(brew --prefix qt@5)"
+make
+```
+
+The built application can then be opened with `open src/Mapper.app`
```
dg0yt: Not sure if this useful for the intended audience. (At least it is not specific for brew.)
Otherweise we might also need to mention `make install` and `make package`.

dg0yt (commented):
Thanks for your suggestions. I did a cursory review now.

FTR future updates might consider ninja instead of make, and driving build steps via `cmake --build <target>`.

INSTALL.md
joneugster: The error I'm seeing is

```
-- Could NOT find Polyclipping (missing: POLYCLIPPING_INCLUDE_DIR POLYCLIPPING_LIBRARY) (Required is at least version "6.1.3")
CMake Warning at CMakeLists.txt:182 (message):
  System polyclipping is missing.  Enabling embedded build.

  Set Mapper_BUILD_CLIPPER=OFF to disable embedded build.


CMake Error at 3rd-party/clipper/CMakeLists.txt:22 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

actually I think I do get it to work by adding `-DCMAKE_POLICY_VERSION_MINIMUM=3.7` as suggested

INSTALL.md
joneugster: I might not be part of the target audience myself :) but I personally would appreciate anything mentioning these things.

INSTALL.md
joneugster: ok, I've dropped that change.

INSTALL.md
joneugster: I've now updated the one instance of `cmake_minimum_required` which was causing the error. I'm oblivious about any implications this might or might not have.

Actually, there's a few build warnings about path normalisations which would dissappear by requiring `cmake_minimum_required(3.31.0)` but I deemed it too bold to propose that now in this PR. Besides, there would be other options to address these comments, besides bumping required deps.

INSTALL.md
joneugster: I've removed the sentence now since you seem not to be thrilled about it :)

INSTALL.md
dg0yt: I see. I use some input variable to close the gap, but this need to eventually land in CMakeLists.txt. Then we don't need it in instructions.

PR #2508, opened 2026-05-29
EditLineTool: Fix vertical bar in status text

dl3sdo:
Only remove the trailing vertical bar from the status text if it exists.
Otherwise the last two chars from "<b>%1</b>: Free movement. " or any translation are chopped.
Visible by using the "Edit lines" tool and pressing the Shift key.


dg0yt:
Thanks.

PR #2512, opened 2026-06-02
Display keyboard shortcuts in the map editor toolbar

mfbehrens:
I am really missing getting the hotkeys displayed in OOMapper.
This is probably a problem that most maintainers here don't have, but when you start using this software you don't know all the hotkeys by heart.
Looking them up all the time is annoying and in the beginning most users probably don't know that they exist.
Displaying them whenever hovering over the icons really helps learning them.

Unfortutnately I am too stupid to actually take a screenshot of it, but it displays:

- Edit objects (E)
- Edit lines (L)
- Set point objects (S)
- and so on

AI was used during programming. I tried my best and I hope it is not complete garbage.

Edit:
image

### DIFF
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -1430,8 +1430,12 @@ void MapEditorController::createMenuAndToolbars()
 			if (action && !action->isSeparator() && !action->shortcut().isEmpty())
 			{
 				auto label = action->text();
-				label.remove(QLatin1Char('&'));
-				action->setToolTip(label + QStringLiteral(" (") + action->shortcut().toString() + QStringLiteral(")"));
+				// Remove unescaped '&', cf. qt_strippedText in qtbase/src/widgets/kernel/qaction.cpp
+				for (int i = 0; i < label.size(); ++i) {
+					if (label.at(i) == QLatin1Char('&'))
+						label.remove(i, 1);
+				}
+				action->setToolTip(tr("%1 (%2)").arg(label, action->shortcut().toString()));
 			}
 		}
 	}
--- a/src/gui/map/map_editor.cpp
+++ b/src/gui/map/map_editor.cpp
@@ -1421,6 +1421,20 @@ void MapEditorController::createMenuAndToolbars()
 	context_menu->addAction(switch_dashes_act);
 	context_menu->addAction(connect_paths_act);
 	context_menu->addAction(copy_coords_act);
+	
+	// Show keyboard shortcuts in toolbar tooltips
+	for (auto* toolbar : {main_toolbar, toolbar_view, toolbar_mapparts, toolbar_drawing, toolbar_editing, toolbar_advanced_editing})
+	{
+		for (auto* action : toolbar->actions())
+		{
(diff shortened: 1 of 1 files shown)


dg0yt:
Thanks for your contribution.
I understand what you are trying to achieve is to display the keyboard shortcut *in the tooltip* of toolbar items. (The toolbar doesn't display text.)
I like the idea. But I see two problems:
- Changing the text of the action has side effects when other means are used to also show the shortcut. In particlar, in the menu.
[Bildschirmaufzeichnung vom 02.06.2026, 22:04:56.webm](https://github.com/user-attachments/assets/f3618047-ac4b-42b7-a29f-6433c3f34bbb)
  I would really like to avoid this pattern. I do not know if it is possible, but I suppose this would need to be handled on the fly when the tooltip is constructed.
- I increasingly lean towards rejecting code contributions based on use of LLMs.
  This is not a specific assessment of this PR or your work. This PR is small and a reasonable.
- Nitpick: piecewise string construction can create problems for internationalization, in particular for right-to-left (RTL) languages. (Not tested.)

mfbehrens:
Thank you for your feedback. I understand your three points.

However, I am still convinced that the feature would really help (new) users of the app.
Do you have a good idea how a feature like this could be implemented?

Otherwise, you can just close the issue.

mfbehrens:
I figured out this nicer way of implementing it, unfortunately the mnemonics destroy it.

`&&` is currently unhandled.

src/gui/map/map_editor.cpp
```diff
@@ -1421,6 +1421,20 @@ void MapEditorController::createMenuAndToolbars()
...
+	
+	// Show keyboard shortcuts in toolbar tooltips
+	for (auto* toolbar : {main_toolbar, toolbar_view, toolbar_mapparts, toolbar_drawing, toolbar_editing, toolbar_advanced_editing})
+	{
+		for (auto* action : toolbar->actions())
+		{
+			if (action && !action->isSeparator() && !action->shortcut().isEmpty())
+			{
+				auto label = action->text();
+				label.remove(QLatin1Char('&'));
```
dg0yt:
```suggestion
				// Remove unescaped '&', cf. qt_strippedText in qtbase/src/widgets/kernel/qaction.cpp
				for (int i = 0; i < label.size(); ++i) {
					if (label.at(i) == QLatin1Char('&'))
						label.remove(i, 1);
				}
```

src/gui/map/map_editor.cpp
```diff
@@ -1421,6 +1421,20 @@ void MapEditorController::createMenuAndToolbars()
...
+				action->setToolTip(label + QStringLiteral(" (") + action->shortcut().toString() + QStringLiteral(")"));
```
dg0yt: Right-to-left languages may want a different order.
```suggestion
				action->setToolTip(tr("%1 (%2)").arg(label, action->shortcut().toString()));
```

dg0yt (commented):
LGTM, but I have nitpicks.

dg0yt:
Danke!

PR #2513, opened 2026-06-03
Require Qt 5.12 API

dg0yt:


dl3sdo:
I assume that workarounds for QT < 5.12 could be removed as well, e.g. in `transform_t.cpp`:
```
#if QT_VERSION == 0x050B01
// Qt 5.11.1 has a regression in QPointF::operator==.
// References:
// https://bugreports.qt.io/browse/QTBUG-69368
// https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903237
// https://github.com/OpenOrienteering/mapper/issues/1116
#  define QTBUG_69368_QUIRK
#endif
```

dg0yt:
Yes, there can be more cleanup. But it can be done in different PRs.

FTR I just fixed the problem with the app not starting with Qt 5.15 for Android. There is still a glitch with the full screen mode being shifted by the size of the camera notch.

src/gui/main_window.cpp
```diff
@@ -765,7 +765,7 @@ void MainWindow::loadWindowSettings()
 {
 #if defined(Q_OS_ANDROID)
 	// Always show the window on the whole available area on Android
-	resize(QApplication::desktop()->availableGeometry().size());
+	resize(QApplication::desktop()->availableGeometry(this).size());
```
dg0yt: This shall be ported away from QDesktopWidget, like done for PieMenu.

dg0yt:
More importantly, it is 0x050c00, not 0x051200 :facepalm:

dg0yt:
Okay, CI here is happy, and it also build for Linux, Windows (MSVC), Android, MacOS (inc. arm64). So it can be merged with rebase soon.

dg0yt:
> I assume that workarounds for QT < 5.12 could be removed as well

I added a commit to remove such code.

src/templates/template_image.cpp
```diff
@@ -349,12 +349,7 @@ void TemplateImage::drawTemplate(QPainter* painter, const QRectF& /*clip_rect*/,
 	// images. This can be worked around by setting a real brush.
 	// Fixed in Qt 5.12.0.
 	/// \todo Fix image opacity in AdvancedPdfEngine
-#if QT_VERSION < 0x051200
```
dg0yt, line 352: Look, another wrong version constant!

dl3sdo: I must have seen it, but it never caught my eye, although it's obvious ;-)

dl3sdo: OMG, I did the same mistake in one of my PRs:
`#if QT_VERSION < 0x051500`

src/gui/widgets/pie_menu.cpp
```diff
@@ -162,18 +162,22 @@ void PieMenu::popup(const QPoint& pos)
...
+		else if (cursor_pos.x() < total_radius)
+			cursor_pos.setX(total_radius);
+		
+		if (cursor_pos.y() > screen_rect.bottom() - total_radius)
+			cursor_pos.setY(screen_rect.bottom() - total_radius);
+		else if (cursor_pos.y() < total_radius)
+			cursor_pos.setY(total_radius);
+	}
 	
 	setGeometry(pos.x() - total_radius, pos.y() - total_radius, 2 * total_radius, 2 * total_radius);
```
dg0yt, line 182: `cursor_pos` is still unused after this PR, but that's a different topic and PR.

PR #2514, opened 2026-06-10
PieMenu: Ensure popup menu stays within viewport

dl3sdo:
The popup pie menu would be clipped when overflowing the screen edges. To avoid clipping, use the existing dead code to shift position.
pie_menu_clipped

### DIFF
--- a/src/gui/widgets/pie_menu.cpp
+++ b/src/gui/widgets/pie_menu.cpp
@@ -175,7 +175,7 @@ void PieMenu::popup(const QPoint& pos)
 	else if (cursor_pos.y() < total_radius)
 		cursor_pos.setY(total_radius);
 	
-	setGeometry(pos.x() - total_radius, pos.y() - total_radius, 2 * total_radius, 2 * total_radius);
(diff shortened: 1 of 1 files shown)


dl3sdo:
I'll either add an `Q_UNUSED(pos);` or modify a copy of pos.
Actually it does not make much difference:
While pos is captured when the user pressed the mouse button, `QCursor::pos();` is captured after the event has been received. I assume no-one to right click while doing a rapid mouse move.

dg0yt:
Hm, how is `popup(pos)` intented to work?
I assume it should show the popup next to pos; i.e. use that pos everywhere; do not use `QCursor::pos()` at all.

dl3sdo:
`MapWidget::_mousePressEvent()` calls `context_menu->popup(event->globalPos());` (with `context_menu = new PieMenu(this);`).
See [`QMouseEvent::globalPos() const`](https://doc.qt.io/qt-6/qmouseevent-obsolete.html#globalPos)
Ok, it seems that you prefer the modification of a copy of pos.

dg0yt:
To be revised after #2513.
Fun fact: I can *skip the calculation* and still the popup will be positioned on the screen...

PR #2519, opened 2026-06-18
MapTest: Fix QRegularExpression syntax error

dl3sdo:
A misplaced closing bracket caused `QRegularExpression::MultilineOption` (0x0004) to pass as a string length argument.
This limited the regular expression string to 4 bytes upon creation, resulting in truncated expressions like "^101" or "102 ".
The bug went unnoticed because the truncated patterns still passed the existing tests without compilation errors.


dg0yt:
Thanks!

Issue #795, opened 2016-11-11
Export to GeoSpatial PDF


Andreasox:
### Steps to reproduce
Export - PDF

### Actual behaviour
Creates a valid PDF

### Desired behaviour
Add georeferencing info. GDAL supports this with: (note BBOX order UL - LR)
gdal_edit.py -a_srs EPSG:3006 -a_ullr 396525 6234620 399405 6230710  <PDF-file>

### Configuration
Mapper Version: 0.6.6
Operating System: OSX 10.12

mfbehrens:
This seems to be implemented but the pdf is broken

<img width="2641" height="1682" alt="Image" src="https://github.com/user-attachments/assets/14cecaa8-8305-4704-b7f9-f8a5a180cb37" />

dg0yt:
> This seems to be implemented but the pdf is broken

There are two ways to produce PDFs.  
The primary way takes the graphics/printing path and produces valid graphical results (subject to limitations of Qt and PDF).  
The second way is only a byproduct of geospatial vector data export and produces insufficient results due to the limitations of GDAL's feature style specification (https://gdal.org/en/stable/user/ogr_feature_style.html).  

Improving the GDAL capabilities would be overstretching the purpose of that library.
So the open task is to inject geospatial references into the output of the graphics/printing path.

PR #2523, opened 2026-07-05
Implement command line interface for OpenOrienteering Mapper

### LINKED ISSUE #2345: RFC: expose PDF export to command line
Would it be possible to expose some commands of the Mapper to a command line tool, so that for example one could tell a Github action to create a PDF (and store as build artifact) from a .xmap file?

mfbehrens:
Fixes #2345

I am glad to receive some feedback.
Its been some time that I worked on QT projects :sweat_smile:

## Tiny Documentation
Documentation yet to come but this is a overview:

```shell
# Invoke cli using
Mapper --cli

# Export (pdf, image)
Mapper --cli export -i mymap.omap -o mymap.pdf
Mapper --cli export -i mymap.omap -o mymap.png
Mapper --cli export -i mymap.omap -o mymap.jpg

# You can also do cursed things (pdf with .jpg extension)
Mapper --cli export -i mymap.omap -o mymap.jpg --output-format pdf

# Export full map extend
Mapper --cli export -i mymap.omap -o mymap.pdf --full-map
Mapper --cli export -i mymap.omap -o mymap.png --full-map
Mapper --cli export -i mymap.omap -o mymap.jpg --full-map

# Convert between file formats
Mapper --cli convert -i mymap.omap -o mymap.xmap
Mapper --cli convert -i mymap.omap -o mymap.ocd --output-format OCD8
Mapper --cli convert -i mymap.omap -o mymap.gpx
```

```shell
> Mapper --cli --help
Usage: Mapper --cli <subcommand> [options]

Available subcommands:
  export   Export map to printable and image formats
  convert  Convert between orienteering map formats
  help     Show this help

> Mapper --cli export --help
Usage: Mapper --cli export [options]
Export map to printable and GIS formats

Options:
  -i, --input <path>    Input map file.
  -o, --output <path>   Output file path.
  --output-format <id>  Output format ID (e.g. pdf, png, OGR-export-DXF).
  --full-map            Export the full map extent instead of the saved print
                        area.
  --dpi <dpi>           Output resolution in DPI (default: 300).
  -h, --help            Displays help on commandline options.
  --help-all            Displays help including Qt specific options.

> Mapper --cli convert --help
Usage: Mapper --cli convert [options]
Convert between orienteering map formats

Options:
  -i, --input <path>    Input map file.
  -o, --output <path>   Output file path.
  --output-format <id>  Output format ID (e.g. XML, OCD, OCD12).
  -h, --help            Displays help on commandline options.
  --help-all            Displays help including Qt specific options.
```

Early documentation: <https://github.com/OpenOrienteering/mapper-manual/pull/79>

## Other cli designs
I was thinking about using another propsed format
```shell
Mapper mymap.omap --cli export -o mymap.pdf
```
This approach is ugly to handle.
What happens if the user provides multiple inputs at once?
What if you have a command does not receive an input file?
Or what if it needs to receive two?
What kind of backwards compatibility does this give you? (none because there is no previous cli)
Also a very unusual cli design in general, when you compare it to any other clis (at least on gnu/linux)

In the end it does not really change anything except for you not beeing able to open a file called `--cli` (however, I am not really sure if this was possible before and you should probably give your file a better name)

## Known issues

- [ ] `.omap` and `.xmap` are currently not an output format but it is simply decided by their file extension what happens to the output
- [ ] I am not sure if anybody gets the difference between export and convert
  - probably something that has to be explained in the docs
  - Basicly export is printing and convert is save as
- [x] missing documentation for output-format
- [x] I am not happy how the command system is implemented
  - there is no help yet that lists the commands
- [ ] `Mapper --cli convert --help` displays wrong application name
- [x] `--cli` has to be placed as first argument -> qt args like -platform cannot be set

Please squash merge this

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -85,6 +85,7 @@ qt5_add_resources(Mapper_RESOURCES_RCC ${Mapper_RESOURCES} OPTIONS -no-compress)
 # (To be used by Mapper executable and by system tests.)
 
 set(Mapper_Common_SRCS
+  cli.cpp
   global.cpp
   mapper_resource.cpp
   settings.cpp
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -1,3 +1,22 @@
+/*
+ *    Copyright 2026 Michael Behrens
+ *
+ *    This file is part of OpenOrienteering.
+ *
+ *    OpenOrienteering is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    OpenOrienteering is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with OpenOrienteering.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 #include "cli.h"
 
 #include <cstdio>
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -354,6 +354,19 @@ int runListFileFormats(const QStringList& sub_args)
 }  // namespace
 
 
+struct SubCommand {
+	const char* name;
+	const char* description;
+	int (*handler)(const QStringList&);
+};
+
+const SubCommand sub_commands[] = {
+	{ "export",          "Export map to printable and image formats", runExport },
+	{ "convert",         "Convert between orienteering map formats",  runConvert },
+	{ "listFileFormats", "List all available file formats",           runListFileFormats },
+};
+
+
 int execCli(int argc, char** argv)
 {
 	if (argc < 3)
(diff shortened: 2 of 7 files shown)


mfbehrens:
I also created a [Docker/Podman Container for executing the mapper CLI](https://codeberg.org/mfbehrens/openorienteering-mapper-cli-container).

It is just the `Containerfile` and the `.containerignore` so the two files could also be moved into the mapper repo if this is wanted.

Edit: fixed link

dl3sdo:
@mfbehrens: the link to your repo at codeberg is not accessible, I assume due to missing access rights.

mfbehrens:
Thank you. I fixed it. No I was apparently not able to copy a link :laughing:

dg0yt (commented):
From a cursory review, it looks very good to me. 

With regard to https://github.com/OpenOrienteering/mapper/issues/795, I would like to ensure that the default for PDF takes the graphics/printing path. 

The initial post lists `OGR-export-DXF`, so there could be `<prefix>PDF` in order to choose the other path. (We should avoid the term `OGR` because it is somewhat obsolete in modern GDAL.)

I also wonder if CLI and GUI should really be one binary in Windows. We should at least test that it works sufficiently.

Abbe98:
Any chance one can take this opportunity to make the Mapper binary lowercase by default?

dg0yt:
> Any chance one can take this opportunity to make the Mapper binary lowercase by default?

Documentation aside, this is an orthogonal concern and shouldn't be handled in this PR.

mfbehrens:
> With regard to https://github.com/OpenOrienteering/mapper/issues/795, I would like to ensure that the default for PDF takes the graphics/printing path.
>
> The initial post lists OGR-export-DXF, so there could be <prefix>PDF in order to choose the other path. (We should avoid the term OGR because it is somewhat obsolete in modern GDAL.)

This is ensured by `export` and `convert` beeing two different commands.
Export properly prints out the map using `QPrinter` while convert tries to find and exporter using `FileFormats.findFormat` / `FileFormats.findFormatForFilename`.

I would consider any changes/bugs in these format_ids/FileTypeRegistry seperate from this pr.
Or did I not get your point?
As far as I understand you say that the FileTypes have to be modernized now that they are exposed closer to the user right?

I implemented a listFileFormats to better debug the formats and also for the users to have a source of documentation

> I also wonder if CLI and GUI should really be one binary in Windows. We should at least test that it works sufficiently.

I think we can definitively disable the cli on android.

I just managed to download the ci build artefact and install them on my Windows 11 vm.
On Windows the cli itself works.
Unfortunately, the output text in the console does not get shown on Windows.
Does anybody know what the issue can be here?
Can anybody with a working Windows buildchain support me on this?

The reason I put the cli into the main binary is that there are no shared objects on Linux and macOS.
Thus, I triggered the cli from the main binary and that's the case on Windows as well to keep it consistent accros OSes.
Since the cli is just a wrapper around the core it shouldn't add that much to the main binary anyway or am I wrong about this?

> Any chance one can take this opportunity to make the Mapper binary lowercase by default?

I think this would be a good thing (except for backwards compatibility).
However, when touching the name I would also suggest prepending an `openorienteering-`.
The binary sits right in my `/usr/bin/` folder and mapper is imho a too generic name.

However, if this is a controversial change (renaming has never been controversion lol), please discuss it in a seperate issue.

Abbe98:
> > Any chance one can take this opportunity to make the Mapper binary lowercase by default?
> 
> Documentation aside, this is an orthogonal concern and shouldn't be handled in this PR.

The reason I think it's relevant here is that making a switch like this is probably much harder once there is an established CLI.

dg0yt:
> This is ensured by `export` and `convert` beeing two different commands.

Sorry, I missed that :facepalm: 

> On Windows the cli itself works.
> Unfortunately, the output text in the console does not get shown on Windows.
> Does anybody know what the issue can be here?

This is what I expected. A windows executable is either a GUI app or a console program. The Mapper executable is a GUI app. This the `WIN32` keyword in CMake's `add_executable`.

PR #2525, opened 2026-07-10
AdvancedPdf: Fix Qt API deprecations

dl3sdo:


dl3sdo:
Closed after conversation with @dg0yt

PR #2529, opened 2026-07-12
Rename binary and all packaging related files to `openorienteering-mapper`

mfbehrens:
~There is a 0% chance that this works~
Actually at least windows worked

## Current state of testing

I just took the builds from the ci and tried to run them on the systems

|Platform|Tested|Works| Comment |
| ---- | ---- | ----- | ----|
|Windows|✔|✔| |
|Ubuntu|✔| ? | ~When I install the deb the dependencies are missing, only tried in a docker container will try on vm~ I was experimenting a bit. I think the build is broken but its not only the fault of this change. Binary is named `openorienteering-mapper` and `openorienteering-mapper.desktop` points to that file |
|Arch AUR|✔|✔| Although not provided by this repo, I took this as a further validation that the linux build works |
|MacOS|✔ thanks to @pkturner| ✔ I have not mac, so i am looking for somebody who can test it |
|Android|✔|✔| ~App will not install, further investigation needed~ I had to sign the akp, then I could install it |


mfbehrens:
I updated the initial comment to include actual testing data.

If anybody has a mac please try to install the [artifacts](https://dev.azure.com/OpenOrienteering/Mapper/_build/results?buildId=3498&view=artifacts&pathAsName=false&type=publishedArtifacts) from azure (this link will be outdated, goto Checks > View more details on Azure Pipelines > MacOS (Release) > 1 artefact produced)

pkturner:
I installed it on an M1 MacBook running MacOS Tahoe 26.5.2. Tested a few little things and it works.

It installs into /Applications/openorienteering-mapper.app/. It's odd to be all lowercase where 57 of the 59 apps have capitalized names. It doubt it will be a bother.

mfbehrens:
@pkturner thank you very much!

Is there a difference between cli (like ls, cp, mv) and GUI programs on MacOS?
Or is the metadata about ooMapper just incorrect somewhere?

dg0yt:
In macOS, advanced applications come as "bundle" which is just a directory with a special name (`Mapper.app`) and a specific structure. 
And Mapper is an advanced application indeed, with resources like PROJ data, symbol sets, examples.
This is leads to a key question, apart from the *filename* of the executable: What is the *display name* of the program? It should probably remain unchanged ("Mapper" or "OpenOrienteering Mapper").

mfbehrens:
Thank you for clarifying.

I found out that the MacOS bundle name was set to `${project_name}`. I now hardcoded it to `Mapper` (since `OpenOrienteering Mapper` was longer that 16 characters, in contrary to the comment). If you think other names would suite better, please tell me

mfbehrens:
@dg0yt is there anything that I can do? If you just did not have time to review it, then everything is fine. I just don't want to provide a half finished/half tested pr, and I am also happy to put more work into it.

PR #2534, opened 2026-08-25
Template: Search recursively for missing templates

dl3sdo:
If templates are no longer found at their previous locations,
Mapper searches alternative locations, including the map directory.
Extend this search to recursively include its subdirectories.


dl3sdo:
The main usecase for this PR is reorganizing templates in folders below the map, for example separating aerial photos from hillshade maps and .gpx tracks. This includes renaming of existing directories.

PR #2542, opened 2026-09-16
Implement XMLDirectoryFormat

mfbehrens:
Might be usefull for #1290

Implemented the following file format:

```
my-map.map
├── colors.xml
├── georeferencing.xml
├── metadata.xml
├── parts
│   ├── part_1.xml
│   └── part_2.xml
├── print.xml
├── state.xml
├── symbols.xml
└── templates
    ├── template1.jpg
    └── template2.png
```

There is also a zipped version (currently `.omapzip`).
This implementation is currently really slow, we have to see if this can be improved.
Would be usefull for sharing single files includeing all templates.
Better explaination and documentation still needed.
Maybe this is also not worth the dependencies and effort to make it fast, and the commit can just be omitted.

Gittability still needs improvement (maybe other pr):
- remove counts
- add unique ids (semantic merge)

xml_directory_format(.h|_p.h|.cpp) still need a lot of work


dl3sdo:
Hi @mfbehrens,

thank you for your proposal.

I see pros and cons of your approach and it's certainly beneficial when using an RC system.
A potential improvement could be to only write those files that need an update, e.g., in many cases only 'parts' and 'state' need to be saved, which certainly speeds up writing and may even help with file corruption. So there should be a dirty flag for each map part.

Regarding the separate storage of map parts: your current approach to use the part name as filename results in orphan files when renaming or removing map parts. BTW: makePartFilename() is incomplete as not all forbidden characters are replaced (e.g., <>:*?"| on Windows).

Regarding templates: you propose to copy the templates initially below the templates folder and later on any template that is added? Since it seems to be flat below 'templates' this would not allow a more fine-grained and user-defined structure (e.g. store aerial photos, gpx tracks etc. in dedicated directories).

Best Regards,
Matthias

mfbehrens:
Thank you for your feedback!

There is quite some work to do.

You are probably right about writing the zip file.
I was already doing research on only writing partial changes today.

> Regarding the separate storage of map parts: your current approach to use the part name as filename results in orphan files when renaming or removing map parts.

That is a good catch. I did not think about that. I will see how I can handle that.

> BTW: makePartFilename() is incomplete as not all forbidden characters are replaced (e.g., <>:*?"| on Windows).

I was thinking about forbidding certain characters directly from the name part, however I was not sure what you and Kai think about it, since it would potentially "break" existing maps.
My idea was to limit the characters to alphanumeric,. -_ and limit the characters to 30 or 40.
On the other hand, I can also do proper sanitisation when writing out the file. Maybe QT even has a function for that.

> Regarding templates: you propose to copy the templates initially below the templates folder and later on any template that is added? Since it seems to be flat below 'templates' this would not allow a more fine-grained and user-defined structure (e.g. store aerial photos, gpx tracks etc. in dedicated directories).

That is probably a good point. My main concern is if we really want to put some storage manager ui into oomapper.

My currently tend to have features like that only available to when you edit the xml by hand.
In my eyes the map directory is a power user feature and those users should be able to move the file manually and change the path in the index file.
I think this is not yet possible with the implementation, since the current implementation will move the template back to its original flat location.

mfbehrens:
> I see pros and cons of your approach and it's certainly beneficial when using an RC system.
A potential improvement could be to only write those files that need an update, e.g., in many cases only 'parts' and 'state' need to be saved, which certainly speeds up writing and may even help with file corruption. So there should be a dirty flag for each map part.

I did some more research on that issue. To write a ZIP file there are two possibilities: 1. create a new zip archive every time you write 2. do some append operation that overrides the previous one. The issue of creating a new zip archive is that it takes long, the issue with the append is that your file size grows over time. In the case of one of my maps the file size is 7x72MB templates + 5MB map parts + 1MB everything else = 510 MB. Since my OOMapper sometimes crashes on wayland (#2485) i save every ~2 mins. After three hours of mapping my file is 960 MB which is almost double the file size. Thus, the approach is usually to occasionally rewrite the full ZIP to get rid of the old data. Sure you can implement that, but it became a lot more complicated than I initially thought, and it is also unpredictable to the user.

My current tendency is to move this to a different PR since it becomes too complicated for if being some "ohh we have a folder format lets also zip it" feature. Maybe keep it as some map sharing container, but converting back to some XML based format also needs to put the templates somewhere, which also introduces complexity.

What are your suggestions?

Appendix: I am currently also considering using a SQLite DB to store all the data, since it is a defined container that can store all data in a single file and can also be edited. But I do not really have a clear picture what this means for read/write speed and restructuring.

